Issue #102, opened 2021-05-15
[Bug]: `KryptonInputBox` does not compile


Wagnerp:
The `KryptonInputBox` does not compile!

Solution:
Re-implement the `KryptonInputBox` on original code, while adding new features, i.e. cue text/font etc.

Smurf-IV:
Can this also sort out #77 ?

Wagnerp:
> Can this also sort out #77 ?

Yes! ;)

Issue #77, opened 2021-04-22
[Bug]: When using `KryptonInputBox` focus is not on the edit box when it is shown


Smurf-IV:
**Describe the bug**

**To Reproduce**
``` c#
var markupName = KryptonInputBox.Show(this, @"Store Current View and Markup's as:", @"Name", @"Name Of this markup", @"OK", startPosition: FormStartPosition.CenterParent);
```
Have to physically click in the edit box.

**Expected behavior**
Edit box to have focus to allow simple user work-flow

**Screenshots**
![image](https://user-images.githubusercontent.com/2418812/115672680-b3d2b900-a343-11eb-9e54-8bf5a3b4c4e8.png)

and it should be the following (Notice the cursor in the edit box)
![image](https://user-images.githubusercontent.com/2418812/115672804-d238b480-a343-11eb-87e0-1831a6da72ff.png)

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 10]
 - Version [e.g. 10.0.19042]

Smurf-IV:
Other issues:
- The Cancel button has the focus
- Entering the Edit area with a mouse click enables the "OK" button, but is should not until something has been entered.
- The Default OK text is `O&k` which looks wrong!
- There shoul dbe more space under the  edit area to balance the lok of the entry
- The Hint text should be lighter. (Should re-use the KryptonTextEditBox hint text !)

Wagnerp:
@Smurf-IV I really think that the `Hint` functionality needs improvement, maybe something like this? https://www.codeproject.com/Articles/27849/WaterMark-TextBox-For-Desktop-Applications-Using-C

Wagnerp:
Fixed in the August 2021 (Canary) update.

PR #108, opened 2021-05-16
Alpha input box

Wagnerp:
* #102 
* #77


Smurf-IV:
New layout should also have had an updated picture ??

Wagnerp:
> New layout should also have had an updated picture ??

I'm going to update the demos today.

PR #115, opened 2021-05-28
Alpha ARM64 (Backburner)

Wagnerp:
* ARM/ARM64 support - may need looking at (only works for .NET 6!)

### DIFF
--- a/Source/Krypton Components/Directory.Build.props
+++ b/Source/Krypton Components/Directory.Build.props
@@ -1,12 +1,16 @@
 <Project>
   <Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.props', '$(MSBuildThisFileDirectory)../../'))" />
+
+  <PropertyGroup>
+    <LangVersion>latest</LangVersion>
+    <RuntimeIdentifiers>win-x86;win-x64;win-arm;win-arm64</RuntimeIdentifiers>
+  </PropertyGroup>
   
   <PropertyGroup>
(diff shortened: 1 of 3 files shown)


Smurf-IV:
Will this make a mess of installers that redistribute these dll's ?
i.e. they will ned to have an x64 and an x86 directory.

Isn't `AnyCPU` supposed to solve the **need** for these explicit identifiers

Wagnerp:
> Will this make a mess of installers that redistribute these dll's ?
> i.e. they will ned to have an x64 and an x86 directory.
> 
> Isn't `AnyCPU` supposed to solve the **need** for these explicit identifiers

As I understand it, the `RuntimeIdentifiers` is new to .NET 5/6. I don't know if this is going to supersede `AnyCPU` eventually. Here are the doc notes: https://docs.microsoft.com/en-us/dotnet/core/rid-catalog

Wagnerp:
May wait for the snapdragon dev kit to test, as I don't think Win 10 can run on a Raspberry PI

Smurf-IV:
If the RID stuff is to be followed, then that also means a lot of dll naming needs to be done.
As this "Suite" is just for windows, I would ignore it for now.

Smurf-IV:
https://www.zdnet.com/article/op-ed-windows-isnt-ready-for-arm-developers/

Wagnerp:
Closing due to the state of ARM64 support in Windows 11 today. Maybe something to revisit in the future.

PR #136, opened 2021-06-02
* Fixes

Wagnerp:


Smurf-IV:
Same comments as the other PR for these.

Wagnerp:
See comment in PR #137

PR #137, opened 2021-06-02
* Fixes

Wagnerp:


Smurf-IV:
Why are they "Disabled" ?
What's supposed to be wrong with them?
i.e. in the read me .. Please reference the problem issue with the link

Wagnerp:
The `KryptonListView` has a `ManifestExceptionError`, unsure why, as all resources are included. See latest comments on the `issues` thread on Discord for more info.

Issue #138, opened 2021-06-03
[Bug]: `KryptonListView` throws a `System.Resources.MissingManifestResourceException`


Wagnerp:
**Describe the bug**
Drag-and-drop a `KryptonListView` onto a KForm to trigger the bug.

Smurf-IV:
@Wagnerp I'm confused...
`KryptonListView` does not exist in the current standard alpha branch ??

Smurf-IV:
And due to the preview stuff of moving the `Using`s, I cannot build in Visual Studio 16.10 !!

Wagnerp:
> @Wagnerp I'm confused...
> `KryptonListView` does not exist in the current standard alpha branch ??
> ![image](https://user-images.githubusercontent.com/2418812/120884147-107af200-c5d9-11eb-864e-fa9be9cdc5f3.png)

@Smurf-IV Ahh, forgot, I removed the files from the `alpha` branch during testing. Will re-upload these files.

Smurf-IV:
Just found them in Master.. As they are "Experimental"..Then they should really have never got into master, and should have stayed in Alpha / Canary !

Wagnerp:
> Just found them in Master.. As they are "Experimental"..Then they should really have never got into master, and should have stayed in Alpha / Canary !

Yes, that was the reason behind `2106.1` stable.

Smurf-IV:
Ok assigning to you, to :
- Remove from master (And canary ?)
- Move to alpha
- Make code compile in 16.10 ??

Then assign back to me when above 2 (or 3) are done ;-)

Wagnerp:
@Smurf-IV Have you tried compiling with `build.cmd`?

Wagnerp:
@Smurf-IV Changed assignments, removed files from both `master/signed` and `canary`.  Don't forget to compile by using `build.cmd`, as VS doesn't work ;)

Wagnerp:
![image](https://user-images.githubusercontent.com/949607/120885389-833b9b80-c5e0-11eb-906b-bd36c18cf01b.png)

![image](https://user-images.githubusercontent.com/949607/120885401-8f275d80-c5e0-11eb-96be-b372115be3b3.png)

Smurf-IV:
Still in progress.. I've not forgotten about it.
(I can see why it was not in the original list of controls - It's a bit of a mare !)

Smurf-IV:
[...]

PR #140, opened 2021-06-05
* KryptonListView

Wagnerp:
* For #138


Smurf-IV:
I do not like this implementation, Can I change it ?

Wagnerp:
> I do not like this implementation, Can I change it ?

Go ahead! ;-)

PR #156, opened 2021-06-10
Alpha generalfixes

Wagnerp:
* Some updates


Wagnerp:
Need a look at the `KryptonFormDesigner`, as the  links are not showing.

PR #158, opened 2021-06-11
* Updates

Wagnerp:
* New the `KryptonButton` will now change its text, based on the `DialogResult` property (this is based on the `KryptonManager` 'Strings' values)

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
@@ -38,9 +38,8 @@ namespace Krypton.Toolkit
         private readonly PaletteTripleOverride _overrideTracking;
         private readonly PaletteTripleOverride _overridePressed;
         private IKryptonCommand _command;
-        private bool _isDefault;
-        private bool _useMnemonic;
-        private bool _wasEnabled;
+        private bool _alwaysUseSetText, _isDefault, _useMnemonic, _wasEnabled;
+        private string _persistantText;
         #endregion
 
         #region Events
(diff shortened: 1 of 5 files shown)


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
```diff
@@ -611,6 +611,51 @@ protected override void ContextMenuClosed()
         {
             _buttonController.RemoveFixed();
         }
+
+        protected override void OnPaint(PaintEventArgs e)
+        {
+            if (DialogResult == DialogResult.None)
+            {
```
Smurf-IV, line 618: What ?
When does this happen,
Always..
What about user text already applied to the button ?

Wagnerp: Can put in a temporary variable to save the current text. The `None` will fire **only** if the dev had previously changed its `DialogResult` state, but then changed it back  to `None`.

Wagnerp: Or... to remove the `None` block, as KM does not have a `GlobalString` for `DialogResult.None`?

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
```diff
@@ -228,6 +227,11 @@ private void ResetButtonStyle()
             ButtonStyle = ButtonStyle.Standalone;
         }
 
+        [DefaultValue(false), Description("Always use the set text, even if the 'DialogResult' value has been changed. (Note: You can manage the 'DialogResult' string values via the KryptonManager on the parent KryptonForm.)")]
```
Smurf-IV, line 230: This should be true, to be compatible with existing iplementations..
i.e. a DefaultOK result for a button that the dev has set to be `Commit`, should still say Commit after updating to this dll.

Currently it will state something different, as therefore translations etc will also be messed up.

Wagnerp: Fixed

Wagnerp:
Merged into PR #161

PR #163, opened 2021-06-19
* Files

Wagnerp:


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
```diff
@@ -228,6 +227,18 @@ private void ResetButtonStyle()
             ButtonStyle = ButtonStyle.Standalone;
         }
 
+        [DefaultValue(true), Description("Always use the set text, even if the 'DialogResult' value has been changed. (Note: You can manage the 'DialogResult' string values via the KryptonManager on the parent KryptonForm.)")]
```
Smurf-IV, line 230: When this class is initialised.. What is setting this value to true ?
The `DefaultValue` is just for the designer, Not the initial field value.

Source/Krypton Components/Krypton.Toolkit/General/GlobalStaticValues.cs
```diff
@@ -17,6 +17,6 @@ public class GlobalStaticValues
     {
         public const int PRIMARY_CORNER_ROUNDING_VALUE = 5;
 
-        public const int SECONDARY_CORNER_ROUNDING_VALUE = 2.5;
+        public const int SECONDARY_CORNER_ROUNDING_VALUE = (int)2.5;
```
Smurf-IV, line 20: Why does this have a `.5` if it's always going to be an `int` ?

Smurf-IV:
2022 is not even released yet, so is not production ready.. Why do this.. This just prevents me even more from trying to fix code (It's bad enough with the `usings` disapearing before .net 6 is even here !!!!!)

PR #255, opened 2021-08-08
* Fixes #254

### LINKED ISSUE #254: [Bug]: 2108 Canary: VisualStudio Themes should not be listed yet, as they cause an exception when used.
image

Wagnerp:


Source/Krypton Components/Krypton.Toolkit/Palette Base/PaletteDefinitions.cs
```diff
@@ -2201,12 +2201,12 @@ public enum PaletteMode
         /// <summary>
         /// Specifies the visual studio dark palette theme.
         /// </summary>
-        VisualStudioDark,
+        //VisualStudioDark,
```
Smurf-IV, line 2204: Should these also be removed from the the ThemeManager as well ?

Wagnerp: Had a look, they don't appear in the ThemeManager

PR #276, opened 2021-08-17
* UAC option for `KryptonButton`

Wagnerp:
* Example to follow


Smurf-IV:
Restored to investigate #323

Issue #291, opened 2021-08-24
[Enhancement]: Build: can the echo time have a time zone, so that when used for PR's it can be checked against the user locale


Smurf-IV:
The following shows a timestamp
https://dotnet.microsoft.com/platform/support/policy/dotnet-framework
Which we have been using as a reference to show that the PR has performed a "Compete" build on out local systems (To prevent merge build breaks etc.)

It would be useful if it also showed the timezone, so that international reviews can check that the time displayed is relevant their current locale, and vice-a-versa, in that reviewers can check international submissions are relevant,

Issue #282, opened 2021-08-20
[Feature Request]: `KryptonScrollbars` need to have smart tags


Wagnerp:
**Is your feature request related to a problem? Please describe.**
As title states!

Wagnerp:
Available in build 2109

PR #292, opened 2021-08-24
Alpha resource sorting

Wagnerp:
* Implements #282 & #291


Wagnerp:
* Features merged into `alpha-enh#290`

PR #334, opened 2021-09-11
Alpha 178 designers

### LINKED ISSUE #178: [Bug]: Many controls are missing if the project is .NET 5/6
**Describe the bug**
If you install the NuGet package in a project using .NET 5/6, half of the controls are missing.

**Workaround:** Configure projects to use `net48` to get access to the full toolkit suite.

Needs to be investigated.

Smurf-IV:
- Add in the `System.Design` types for the different frameworks
- Add in `AssemblyRef.cs` as stolen from Winforms Github
- Replace `typeof`'s in `Designer`s and `Editor`s
- Restore `//[DefaultValue` as it wasn't that causing the designers to not to load

Attempts to fix #178

image

Still not sorted, but must be getting closer !


Wagnerp (commented):
Has the changelog been updated to reflect the changes made?

Smurf-IV:
> 
> 
> Has the changelog been updated to reflect the changes made?

No, as this is not fixed yet.
And it is a restore **back** for the Default values.

Wagnerp:
> > Has the changelog been updated to reflect the changes made?
> 
> No, as this is not fixed yet.
> And it is a restore **back** for the Default values.
> 
> But I need this to go into a Nuget to see if pulling it in via that has solved the Designer Issues.
> If it does not, then it might have to be a forced V4.0.0.0 as theorised in the Work Items

I think that 'Nightly' packages may be needed

Wagnerp (approved):

Issue #335, opened 2021-09-12
Create 'Nightly' nuget packages When Merges into alpha are done


Smurf-IV:
> > Has the changelog been updated to reflect the changes made?
> 
> No, as this is not fixed yet.
> And it is a restore **back** for the Default values.
> 
> But I need this to go into a Nuget to see if pulling it in via that has solved the Designer Issues.
> If it does not, then it might have to be a forced V4.0.0.0 as theorised in the Work Items

I think that 'Nightly' packages may be needed

_Originally posted by @Wagnerp in https://github.com/Krypton-Suite/Standard-Toolkit/pull/334#issuecomment-917569366_

Wagnerp:
Getting somewhere :)

![image](https://user-images.githubusercontent.com/949607/133049475-1de3e494-ad23-42d4-902c-b2820a4e6ec1.png)

Wagnerp:
Completed

PR #345, opened 2021-09-13
* Nightly Packages

Wagnerp:
- Implement #335


Wagnerp:
Going to pull 2109 & 2109.1 as there are too many issues with them

PR #346, opened 2021-09-13
* Changes

Wagnerp:
- Pulled build 2109(x)
- Corrected Nightly package authoring issues


Smurf-IV:
@Wagnerp Does this mean that the nightlies are done "Automatically" now and are pushed to nuget with the "Pre-Release flag" set?

Wagnerp:
> @Wagnerp Does this mean that the nightlies are done "Automatically" now and are pushed to nuget with the "Pre-Release flag" set?

@Smurf-IV Need to set the pre-release flag, but uploads can be done via the `publish` script

PR #350, opened 2021-09-14
* Changes

Wagnerp:
image

image

image


Smurf-IV:
Why is the DataeTime not in the package version info any more ?
How will the nuget explorer (i.e. in visual studio) be able to determine a "Newer" nightly build ?

Wagnerp:
VS was complaining that the build string didn't conform. Haven't tried tacking it onto the end of `-alpha` though, i.e `-alpha.20210914140000`.

Issue #381, opened 2021-09-23
[Feature Request]: Allow KColorDialog to set form title text


Smurf-IV:
Follow steps here and add to the Kryptonised ColorDialog

https://stackoverflow.com/questions/762009/how-do-i-change-the-title-of-colordialog

Note: Could be done for the Font as weel ?

Smurf-IV:
![example](https://user-images.githubusercontent.com/2418812/134774709-284d6be2-c6be-4134-b75c-e50348ccf571.png)

PR #390, opened 2021-09-25
- Implement changing the Common Dialog Title

Smurf-IV:
Implements: #381
example

image


Smurf-IV:
@Wagnerp Note: The build time displayed states that it is GMT, but it is actually the locale time being displayed (Which for me is BST)

Wagnerp (approved):

PR #420, opened 2021-10-05
Alpha removedotnet5

Wagnerp:
- Remove .NET 5 as per lifecycle (Feb 2022), planned build 6.2201

### DIFF
--- a/Documents/Help/Changelog.md
+++ b/Documents/Help/Changelog.md
@@ -5,6 +5,7 @@
 * Implemented [#395](https://github.com/Krypton-Suite/Standard-Toolkit/issues/395), Add ability to add form icon to common dialogs
 * Implemented [#404](https://github.com/Krypton-Suite/Standard-Toolkit/issues/404), `KryptonInputBox` to have a default button
 * Fixed [#403](https://github.com/Krypton-Suite/Standard-Toolkit/issues/403), Krypton.Toolkit.Nightly `Version="6.2109.272-alpha"` has removed code that was in 270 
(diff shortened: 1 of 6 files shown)


Wagnerp:
Setting this to draaft

Smurf-IV:
.Net 5 "May" be removed from 2022 solution but not the 2019.. 
There are still a lot of Nuget packages that do not support .net6 
AND a lot of Pluggins for VS that are not compatible with VS17 (i.e. Resharper !!)

Wagnerp:
> .Net 5 "May" be removed from 2022 solution but not the 2019.. There are still a lot of Nuget packages that do not support .net6 AND a lot of Pluggins for VS that are not compatible with VS17 (i.e. Resharper !!)

Might be better to wait until EOL (May 2022)

Wagnerp:
@Smurf-IV What if we reopen the idea of having `Lite` packages?

- `Lite` => net framework 4.8, .NET Core 3.1, .NET 6/7
- 'Other' => net framework 4.6.2 - 4.8, .NET Core 3.1, .NET 6/5/7

Smurf-IV:
> @Smurf-IV What if we reopen the idea of having `Lite` packages?
> 
>     * `Lite` => net framework 4.8, .NET Core 3.1, .NET 6/7
> 

The "Lite" package should not include 7 in that case !
Not until it is actually released.

Wagnerp:
> The "Lite" package should not include 7 in that case

The change would only affect the Stable/Signed releases, so 7 won't be included until November.

Wagnerp:
Closing, due to previous comments

Issue #402, opened 2021-09-29
[Bug]: KryptonInputBox cutting off text


Ash-IPRANOps:
KryptonInputBox cut off text as seen in image below.

![image](https://user-images.githubusercontent.com/43791859/135309696-5822d17b-859a-42f1-9546-7102c5e645ea.png)

I have the latest Nightly update as of yesterday. I am using Windows 10 Enterprise. Build 19042.1110

Wagnerp:
@Smurf-IV Is it possible to dynamically resize a KIB like a messagebox?

Smurf-IV:
> 
> 
> @Smurf-IV Is it possible to dynamically resize a KIB like a messagebox?

Looks like it'll have to, Word Wraps always complicate things...

Ash-IPRANOps:
Can you allow the user to set the size manually too? that way it ensures in case the auto size fail, the developer can have some control. Also, it would allow the developer to create a size that is maybe visually preferred.

Smurf-IV:
Next would be `multi-line` entry ??
@Ash-IPRANOps Can you add them as separate requests "Feature" please?

Ash-IPRANOps:
@Smurf-IV I created a new feature request. thanks.

Smurf-IV:
Thanks for testing.. 
I've missed something obvioulsy !!

PR #443, opened 2021-10-17
- Remove designer elements from the KIB and add to the manager

Smurf-IV:
- Add table layout for AutoSizing
- Test AutoSizing
#402

image


Wagnerp:
Only need to update `Changelog.md`

PR #467, opened 2021-11-11
* Edits & comment out unused code

Wagnerp:
* Comment out unused code to reduce compiler warnings

image


Source/Krypton Components/Krypton.Ribbon/Controls Ribbon/KryptonRibbonContext.cs
```diff
@@ -120,7 +120,7 @@ public Color ContextColor
             set
             {
                 // We never allow a null or transparent color
-                if ((value == null) || (value == Color.Transparent))
+                if (/*(value == null) ||*/ (value == Color.Transparent)) // 'color' is never 'null' CS8073
```
Smurf-IV, line 123: **Do not do this** for public setters, the compiler can never know how these are going to be used in the "Wild"
Just add the warning to the `.editorconfig`

Wagnerp:
A bit lost here... 

image

Smurf-IV:
> 
> 
> A bit lost here...
> 

That view is supposed to be helpful.. I have never found it to be more confusing..
Goto "Open with (Right click on file) and use the default text editor.. It's a lot easier for a developer to use then the unhelpful management view

Source/Krypton Components/Krypton.Ribbon/Controls Ribbon/KryptonRibbonContext.cs
Wagnerp: Added to `.editorconfig`

Wagnerp:
image

PR #474, opened 2021-11-15
Alpha themes

Wagnerp:
* Complete #469 

image


Smurf-IV:
Can there also be an image attached here (And / or) in the bug to show that it has been fixed ?

Wagnerp:
ToDo 469

Need to adjust the images

Wagnerp:
Animation

Smurf-IV:
> Need to adjust the images

Yes, it appears that the hover over image is slightly smaller, causing the "next to" image, to move.

Wagnerp:
Aiming to fix this tomorrow

Wagnerp:
@Smurf-IV So we have 2 routes, make the images 11 x 9 (which will cause loss of definition) **_or_** make all control box images larger. Your thoughts?

Smurf-IV:
Make them larger so that when used on larger resolutions they still look good.

Wagnerp:
> Make them larger so that when used on larger resolutions they still look good.

That's what I thought, I might not be able to do it for the Sparkle themes, but definitely the O2k7/10/13/365 themes

Wagnerp:
Setting new images to 25 x 25, which is the true size

Smurf-IV:
> Setting new images to 25 x 25, which is the true size

Can I see a new demo please ?

Wagnerp:
> > Setting new images to 25 x 25, which is the true size
> 
> Can I see a new demo please ?

Will do when I grab all the images

Wagnerp:
@Smurf-IV Is there a way to make the buttons slightly bigger?

Wagnerp:
Animation

Smurf-IV:
> @Smurf-IV Is there a way to make the buttons slightly bigger?

Seems like the drawing of the position of the buttons is now slightly too low, raise it as a separate bug once this has been merged.
And yes it is possible, but you'll find that all the other forms elements will then be in the wrong place (Icon, Title, Extra text, Extra icon) so best just to sort positioning, and then take care of the position when maximised (Bit like I did a while ago)

Smurf-IV:
Is this ready to go now ?

Wagnerp:
> Is this ready to go now ?

Yes it is :)

PR #505, opened 2021-12-04
- Handle the Culture, and also remove the trailing decimal point when …

### LINKED ISSUE #502: [Bug]: KNumericUpDowner, when told to display 1 decimal place, does not display a `0` when needed
Latest Nightly - .net48 x64
Value is 2.0
image

When click on it then displays the `0` !!

image

Smurf-IV:
…numbers trimmed.

- Now works correctly if the number is only 3000 and trimming takes place.
- Respects the `ThousandsSeparator` override
Fixes #502

There are problems with the 2019 build:
image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonNumericUpDown.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonNumericUpDown.cs
@@ -13,6 +13,7 @@
 // ReSharper disable MemberCanBePrivate.Local
 // ReSharper disable UnusedMember.Local
 
+// ReSharper disable MemberCanBeProtected.Global
 namespace Krypton.Toolkit
 {
     /// <summary>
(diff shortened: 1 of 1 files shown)


Smurf-IV:
Fine with 2022 command (But I think all the analyzers are turned off, so it's telling lies !)
image

Wagnerp:
> …numbers trimmed.
> 
> * Now works correctly if the number is only 3000 and trimming takes place.
> * Respects the `ThousandsSeparator` override
>   Fixes [[Bug]: KNumericUpDowner, when told to display 1 decimal place, does not display a `0` when needed #502](https://github.com/Krypton-Suite/Standard-Toolkit/issues/502)
> 
> There are problems with the 2019 build: image

@Smurf-IV I don't know if VS2019 fully supports C# 10

Smurf-IV:
`global using` used to work, before the solution where branched to support 2022.
Also noticed that this does not contain a change to the changelog file...

Wagnerp:
> `global using` used to work, before the solution where branched to support 2022. Also noticed that this does not contain a change to the changelog file...

@Smurf-IV Ahh... figured it out. VS2019 by default _only_ supports C# 9,  `global using` is a >= C# 10 feature, easy to fix!

Wagnerp (commented):
Reviewed

Issue #452, opened 2021-10-26
[Bug]: KryptonDockingManager - KryptonPage - MinimumSize not working


raygraham442:
I've created a KryptonPage that has an attached UserControl.  I want to configure the page so that it can't be resized smaller than 682 x 533.  I'm setting the MinimumSize to 682 x 533.  Though as I move and pages around in the WorkSpace I find that the KryptonPage set the MinimumSize on can be resized smaller than the MinumumSize of 682 x 533.  Is there something more that I need to configure for the MinimumSize to work?

Thanks in advance,

Smurf-IV:
This might have been fixed with the code changes in #459.
Please retest with a nightly from today 👍 (i.e. greater than 316)

Smurf-IV:
> This might have been fixed with the code changes in #459. @raygraham442 Please retest with a nightly from today 👍 (i.e. greater than 316)

No it has not been fixed.

It appears that there are several issues here:
- `KryptonDockspace` doe snot have it's `MinimumSize` set to the pages it's contains minimum size.
-  `KryptonPage` doe snot respect the `MinimumSize` of any controls it contains when it is not docked (If at all !)
- Sizing in docking does not appear to work as expected when the docking is moved around on different edges.

Smurf-IV:
Arggh.. I don't like this Docking code ....

Smurf-IV:
Massive cleanup has been triggered to try and "see the wood for the trees"
https://github.com/Krypton-Suite/Standard-Toolkit/pull/542

Smurf-IV:
The following video shows:
- Designer with Min Size set to the area shown
- Multi Control Demo App initial start showing the overlapping controls obeying the minimum Size
- Floating window showing minimum size to show
- Attempt to make the Docked control smaller
- Maximise app, then resize still works

![DEsignerControlHasMinSizeSet](https://user-images.githubusercontent.com/2418812/147356731-ed8fc012-9832-4e1d-bb63-0f27658d9c88.gif)

PR #544, opened 2021-12-24
- Add sizing of panels based on "MimimunSize" being set on controls

Smurf-IV:
- Make usre "Special strings" are marked with a `@`
- Make sure `ArgumentException` has the actual "Argument" that is causing the "Exception"
- Apply `linq` over `foreach` condition statements
- Fix some spelling mistakes
- Apply constructor should be `protected` for abstract classes
Fixes: #452

Build-2022:
image


Smurf-IV:
Build-2019:
image

Wagnerp (approved):

Issue #70, opened 2021-04-04
[Bug]: If the BlurWhenFocusLost is set and the app is underneath another, then it will still have the blur overlay topmost (i.e. on top of the other app which is currently fullscreen!)


Smurf-IV:
**Describe the bug**
As found in #56 

https://github.com/Krypton-Suite/Standard-Toolkit/issues/56#issuecomment-808719860

**To Reproduce**
win 10 (currently 21H2 but was 1901)
Start app with blurring enabled
Does not always happen, but when it does it looks pants! 

Below shows IDE being clicked onto to bring to top:

![BlurIsOnTop](https://user-images.githubusercontent.com/2418812/150639876-77313d5c-6b0f-48c8-bf59-6e88c00058f3.gif)

Smurf-IV:
@Wagnerp Can this be dropped, as it is not a recommended feature of Windows since win 10 came out (Due to deprecation of the API#s started in Vista)

Wagnerp:
> @Wagnerp Can this be dropped, as it is not a recommended feature of Windows since win 10 came out (Due to deprecation of the API#s started in Vista)

@Smurf-IV Should the `Obsolete` tag be used?

Smurf-IV:
> @Smurf-IV Should the `Obsolete` tag be used?

No, as that infers that some functionality may be working (And also brings up a warning during build, which can be ignored)

In it's current guise - It does not work as intended (Or did in win 7!)

Some of the parameters will need to be removed (Like the opacity, as I cannot get that to work in Win 10)
And it does not do what it was intended even in Win 7, because if you have a window (another app) partially over the Form, and you click on the window to bring to foreground, then that part of the window is used as part of the opaque image(i.e. move the foreground and it still shows ), Also it is using "TopMost" which means that the opaque window is on top of everything in Win 10, due to a slightly different message sequence !!

I propose to remove some values, change the explanation on the existing, and get it to work consistently amongst the supported OS's (i.e. Win 7 onwards)

Wagnerp:
> I propose to remove some values, change the explanation on the existing, and get it to work consistently amongst the supported OS's (i.e. Win 7 onwards)

I agree with that, since a few years ago support for Vista was dropped.

Smurf-IV:
So this has been a ride into the unknown.....
Some issues:
- Gaussian blur API now longer works on win 10
- TopMost windows make a mess of things
- Anything krypton handles painting so completely ignores the DWM mode of the form
[...]

Issue #69, opened 2021-04-04
[Bug]: The MessageBox display clears the blur underneath (Due to the code thinking it has lost focus, even when the messageBox is owned)


Smurf-IV:
**Describe the bug**
As found in #56 

https://github.com/Krypton-Suite/Standard-Toolkit/issues/56#issuecomment-808719860

**To Reproduce**
Use the Stand-Toolkit Examples II solution and pop the messagebox up

Smurf-IV:
![BlurIsOnTop](https://user-images.githubusercontent.com/2418812/151695112-8ee1c62a-797a-4444-8b4d-7c891195e04d.gif)

Issue #68, opened 2021-04-04
[Bug]: When Blurring on Windows 10 with a large screen, the offset of the blur overlay is wrong


Smurf-IV:
**Describe the bug**
As found in #56 

https://github.com/Krypton-Suite/Standard-Toolkit/issues/56#issuecomment-808719860

**To Reproduce**
Use the Stand-Toolkit Examples II solution and pop the messagebox up

Smurf-IV:
![BlurIsOnTop](https://user-images.githubusercontent.com/2418812/151695096-7f03422e-1cbd-4f3f-a553-214bb191dcd8.gif)

PR #618, opened 2022-01-30
- Make the blur work in the supported OS's

Smurf-IV:
- Ensure messagebox code works
- Grab image of source before loss of focus to ensure correct colouring
- Remove code that is no longer applicable
Resolves:
  - #70
  - #69
  - #68


Smurf-IV:
@Wagnerp Cannot load solution into VS2019 any more??
image
And it does not build from the command line.

Wagnerp:
> @Wagnerp Cannot load solution into VS2019 any more?? image And it does not build from the command line.
> 
> When re-adding in VS2019 It displays this: image

@Smurf-IV I'll take a look at it later :)

Wagnerp:
@Smurf-IV Just checked on VS2019, doesn't seem to have any issues?

image

Smurf-IV:
Just tried with the current alpha, and I get the same.
image

Wagnerp:
Now that's odd, I'm on the current alpha, and I get this:

image

Issue #603, opened 2022-01-20
[Bug]: Title Bar Images Stretched/Cropped (Regression)


sdreb3421:
The new canary build (60.22.1.17) has introduced stretching/cropping of the icons/images in the krypton form title bar. This could be related to #601.

Stable:
![image](https://user-images.githubusercontent.com/1911567/150245040-72a55450-8012-4e4e-a0a7-716d1d98e6a6.png)

Canary:
![image](https://user-images.githubusercontent.com/1911567/150244649-c98f7174-d5be-44ee-a476-01957ef7d727.png)

Smurf-IV:
@sdreb3421 I am unable to replicate this (At any of the scaling I have on my QHD monitor with Win 10 20H2)
Can you state what 
- size display you are using
- scale applied
- Version of windows

sdreb3421:
I was unable to resolve the issue with any scale setting unfortunately. The advanced scaling settings are disabled, but changing them didn't fix it either. Neither did changing the resolution of my monitor (native 1920x1080). The images below show my OS info, the results of various scaling (100%, 125%, 150%), and the images I used.

To replicate the issue, create an empty krypton form, no palette settings, and add button specs to the form using the three images below. The issue is most obvious when using smaller images that have rounded edges. If you're unable to replicate this I will go run some tests on other computers I can find.

![image](https://user-images.githubusercontent.com/1911567/150656246-31adadde-ca5e-4d11-a85c-1a6fbf7165de.png)

![Test](https://user-images.githubusercontent.com/1911567/150656154-5706e8e8-f844-4af0-889e-bae4cd04f6ad.png)

![help16](https://user-images.githubusercontent.com/1911567/150656173-7fd74a0c-b2d2-4ad7-9714-7c9e6fc87b4d.png)
![play20](https://user-images.githubusercontent.com/1911567/150656185-30d6bcf2-8363-447f-aa60-3be8df7c249c.png)
![recordredtop](https://user-images.githubusercontent.com/1911567/150656193-458642a4-94e0-4f7c-b877-03c8db618c1e.png)

Wagnerp:
Hi @sdreb3421 is it possible to use images, which are 24 x 24?

Smurf-IV:
> I was unable to resolve the issue with any scale setting unfortunately..... The images below show my OS info, the results of various scaling (100%, 125%, 150%), and the images I used.
Are you saying that this is a problem even when 100% on Native scaling ?

[...]

PR #627, opened 2022-02-06
- Do not resize buttons spec images, as that is handled by the Renderer

Smurf-IV:
#603

image


Wagnerp (approved):

Smurf-IV:
@Wagnerp 
The above fix has "re-broken" the upscaling when it is required ....
image

Wagnerp:
@Smurf-IV I think the `ButtonSpecView` is broken.

Issue #697, opened 2022-04-13
[Bug]: Number 9 is handled in ribbon textbox/richtextbox.


pgaraneta:
I can type numbers 123456780 in the ribbon textbox/richtextbox, however, 9 is handled.

I tried using other keyboards as well as on screen keyboard and typing 9 into the ribbon textbox/richtextbox, 9 is still handled. If I type number 9 into a normal textbox in a form in the same app, there is no problem. Notepad, just for verification, I can type 9.

![image](https://user-images.githubusercontent.com/85190700/163246562-cdfc305a-abe8-40c6-ae6b-c96fd7615724.png)

giduac:
Hi,

I had a look at this and found that next to Textbox and RichTextBox other controls have this:
```
KryptonRibbonGroupDomainUpDown
KryptonRibbonGroupNumericUpDown
KryptonRibbonGroupComboBox (DropdownStyle = DropDown)
KryptonRibbonGroupMaskedTextBox

```
As soon as you enter the number 9 the cursor/focus is "removed" from the control.

pgaraneta:
Yes, after hitting 9, ribbon textbox/richtextbox indeed, loses focus. Tested on an MDI with at least a child form opened.

Smurf-IV:
Might be something that has been "Introduced" but using an older toolkit version `9` works:
![image](https://user-images.githubusercontent.com/2418812/163667991-c42c3fda-9182-472a-a118-f2cbd50ec4ab.png)
When I have a building environment I'll test with the latest....

Smurf-IV:
I am able to reproduce this in the latest "Example" as shown above.
It appears that the base class for `RichText` is okay:
![image](https://user-images.githubusercontent.com/2418812/163705462-3e829be4-6737-4f01-b262-770e779f16e8.png)

I continue to investigate....

Smurf-IV:
Found the problem...
- Even tho the `Keys` type is a flags `enum` you cannot check for Tab and other key types at the same time!
- `Tab = 0x09`
- `NumPad9` or `D9` = `0x69`
- `if (keyData.HasFlag(Keys.Tab))` will be true when it should not be !

Smurf-IV:
Note: This was "Broken" on the 2021-12-28 !

PR #698, opened 2022-04-17
- Do not use Keys.Haskey as the flags enum is not accurate

Smurf-IV:
#697
build Options 2 -> 2:
image


Smurf-IV:
ChangeLog done in the next PR.

Wagnerp (approved):

Issue #669, opened 2022-03-11
[Feature Request]: `CueHint` features to be configured via the `KryptonPalette`


Wagnerp:
**Is your feature request related to a problem? Please describe.**
Can the `CueHint` features be configured via the `KryptonPalette`, i.e. colour and font?

Smurf-IV:
Is #535 related to this ?

Wagnerp:
> Is #535 related to this ?

See latest comment on #703

Smurf-IV:
Seems to be a thing now:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/1ae291a8-9c3b-445c-87c0-a88a4144274a)

PR #702, opened 2022-04-20
* Resolves #701

### LINKED ISSUE #701: [Bug]: `CueHint` in `KryptonPalette` does not work
Just started to use Palettes to override the look & feel of ButtonSpecs
Found a couple of issues 

Whenever I change a setting within the Palettes I get the following error
image
Although it doesnt seems to cause any problems, I would like to know the cause / workaround / fix

In addition, any changes I make to the Palette (in this case, ButtonStyles > ButtonCustom1>StateCommon>....) are ignored by the control next time I start VB studio, its an easy fix, I just need to make a change to anything within the Palette and all the settings magically appear in the relevant controls BUT I need to remember to do this every time VB starts

Wagnerp:
* Resolves #701 

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonPalette.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonPalette.cs
@@ -134,6 +134,7 @@ namespace Krypton.Toolkit
             TabStyles = new KryptonPaletteTabButtons(_redirectCommon, _needPaintDelegate);
             TrackBar = new KryptonPaletteTrackBar(_redirectCommon, _needPaintDelegate);
             ToolMenuStatus = new KryptonPaletteTMS(this, _basePalette.ColorTable, OnMenuToolStatusPaint);
+            CueHintText = new PaletteCueHintText(_redirector, _needPaintDelegate);
 
             // Hook into the storage change events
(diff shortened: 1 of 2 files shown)


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonPalette.cs
```diff
@@ -321,6 +321,8 @@ public KryptonPaletteCheckButtons ButtonStyles
         #endregion
 
         #region Cue Hint
+        /*
```
Smurf-IV: Do not do this, otherwise the CueHint text will not be stored in the designer into the relevant file

Wagnerp: Should I set it to `null`, as I think it's linked to #669?

Smurf-IV:
No, This stores the Text, Font (Style, colour, size etc). 
Should not have anything to do with the pallete, unless a partial style has been added and it has not competed through the code `switch` statements when it should be applying it.

Wagnerp:
> No, This stores the Text, Font (Style, colour, size etc). Should not have anything to do with the pallete, unless a partial style has been added and it has not competed through the code `switch` statements when it should be applying it.

So, leave it alone?

Wagnerp:
image

PR #726, opened 2022-06-04
* Themes

Wagnerp:
* Made classes public in order for additional themes to be made. See https://github.com/Krypton-Suite/Theme-Palettes/issues/34

image


Wagnerp:
Additional info, in order to reduce code duplication, it's better to use as much of the original code as feasibly possible

Issue #666, opened 2022-03-10
[Bug]: KryptonTextBox `Validate` / `Validating` / `KeyUp` events are invoked twice


overlinejota:
A simple form with a KrytonTextBox and a button.
The next code wave twice when KryptoTextbox1 loses focus:

```
namespace AppForm{
    public partial class Form1 : Krypton.Toolkit.KryptonForm{
        public Form1() {
            InitializeComponent();
        }            
        private void kryptonTextBox1_Validated(object sender, EventArgs e) {
            MessageBox.Show("Error","Hello");
        }
    }
}

private void InitializeComponent() {
	...
	this.kryptonTextBox1.Validated += new System.EventHandler(this.kryptonTextBox1_Validated);
	...
}

```
Thanks;

giduac:
KMaskedTextbox is also affected.
Also  happens with the "Validating" event.

Wagnerp:
@Smurf-IV Is this linked to the fixes that were merged yesterday?

Smurf-IV:
> @Smurf-IV Is this linked to the fixes that were merged yesterday?

Do not think so, Did not look at any impact in this area as I was in the Dock and ribbon projects

Wagnerp:
> > @Smurf-IV Is this linked to the fixes that were merged yesterday?
> 
> Do not think so, Did not look at any impact in this area as I was in the Dock and ribbon projects

Is this just missing the 'Focus()' call in the validate event?

thiagoraheem:
I have the same situation and I created a project to demonstrate

https://github.com/thiagoraheem/KryptonTest

Smurf-IV:
This also affects other controls like `KDomainUD`

It is because 
- the Kyrpton `VisualControlBase` is based on the winforms `Control`
- When "Adding" an event in the designer, it adds to the direct `Validated` EventHandler
- Inside Krypton it also has added an EventHandler to the control it is wrapping.
- The Krypton EventHandler calls the base controls `On#####()` version, thus triggering it twice.

This could take a while to fix, e.g. just "some" of the eents in the KDominUD:
``` c#
        private void OnDomainUpDownTextChanged(object sender, EventArgs e) => OnTextChanged(e);

        private void OnDomainUpDownScroll(object sender, ScrollEventArgs e) => OnScroll(e);

        private void OnDomainUpDownSelectedItemChanged(object sender, EventArgs e) => OnSelectedItemChanged(e);

[...]

PR #766, opened 2022-06-25
- Make sure that text events are only fired once

Smurf-IV:
- `KeyUp`
- `Validating`
- `Validated`

image

#666


Wagnerp (approved):

Smurf-IV:
Yes it "Should be"  possible,
But Winform Tooltips have shadows:
image

Or is this different in Win 11 ?

Smurf-IV:
Done. All things that have ToolTips "should" now have a `ToolTipShadow` Property:
image

1st page is off, 2nd is on (As default)
TTShadows

Issue #689, opened 2022-04-06
[Feature Request]: A way to remove border shadows on `KryptonToolTips`


Wagnerp:
As raised by @sdreb3421 on Discord

> Is there any way to get rid of the border shadow on tool tips? I see this on my context menus as well.

![image](https://user-images.githubusercontent.com/949607/161946398-a460c579-8ef2-4ad8-9e16-7d1860060af8.png)

![image](https://user-images.githubusercontent.com/949607/161946480-1ed25373-5657-4584-9ae7-645932a39496.png)

Smurf-IV:
Done. All things that have ToolTips "should" now have a `ToolTipShadow` Property:
![image](https://user-images.githubusercontent.com/2418812/175775267-b2c4d6a6-4894-445d-bed0-80fc973e852a.png)

1st page is off, 2nd is on (As default)
![TTShadows](https://user-images.githubusercontent.com/2418812/175775277-4b7c157f-60d5-47e2-8ee2-ae9597b8b6db.gif)

Smurf-IV:
Winform Tooltips have shadows:
![image](https://user-images.githubusercontent.com/2418812/175775372-9d6776ae-8867-4ad0-9cd8-a1f7c256c37c.png)

Or is this different in Win 11 ?

Note: "Material Design" suggests that tooltips should not have shadows (Or images, and lot's of other restrictions :-()

sdreb3421:
Sorry for the delay getting back to this. I was able to confirm that setting the ToolTipShadow property false on individual button specs works as expected. Would it be possible to set the default property value false? 

And were you able to look at context menus too? Is it possible to add a shadow property for them, or just remove the shadow? Thanks

Smurf-IV:
@sdreb3421 Thanks fro confirming. Please create 2 new "feature requests" for the above comments

PR #768, opened 2022-06-26
* Reinstated fixes for #666 & #689

Wagnerp:
* Reinstated fixes for #666 & #689

image


Source/Krypton Components/Krypton.Toolkit/Palette Base/PaletteDefinitions.cs
```diff
@@ -1068,9 +1068,9 @@ public interface IPalette
 
 		#region BaseFontSize
```
Smurf-IV, line 1069: If you undoing 703, then there other changes that need to be undone (i.e. Changelog.md)

Wagnerp: Forgot about that, thanks for reminding me :)

Issue #728, opened 2022-06-06
[Feature Request]: Bring MessageBox `States` inline with latest .Net 6 


Smurf-IV:
https://docs.microsoft.com/en-us/dotnet/desktop/winforms/whats-new/net60?view=netdesktop-6.0#updated-apis

![image](https://user-images.githubusercontent.com/2418812/172110241-715207b0-a14b-4052-b489-b55b4fe3272c.png)

===================
Edit: 
The "Idea" of these changes is to bring the functionality "Down" from .Net 6 / 7 into the earlier versions, so that a Single code base will work for multi-target builds.
===================

Wagnerp:
@Smurf-IV Just thinking about it, do these new features need the `#if` definition to work? If so, I'll try to tackle some of the messagebox portion of it today.

Smurf-IV:
`#if` would be one way to "Limit" the functionality,
But... It would be nice if it could be made "Backward compatible" (i.e all the way down down to .net 4.6.2)

Wagnerp:
Needs testing >= .162

Wagnerp:
@Smurf-IV Do we know if this is working, also what is `button4` used for?

Smurf-IV:
> @Smurf-IV Do we know if this is working, also what is `button4` used for?

![image](https://user-images.githubusercontent.com/2418812/179389608-5779f355-3555-45b4-abf7-aee683475547.png)

Wagnerp:
> > @Smurf-IV Do we know if this is working, also what is `button4` used for?
> 
> ![image](https://user-images.githubusercontent.com/2418812/179389608-5779f355-3555-45b4-abf7-aee683475547.png)

So, is there a `help` button now?

Wagnerp:
@Smurf-IV I assume the "new" button4 option would go to the left of the existing buttons _if_ it's a dedicated help button?

Smurf-IV:
You will have to create a demo app in .Net 6 and try....

Wagnerp:
I'm not sure if we even need this, as a button for help is already catered for

Wagnerp:
Got it!! PR coming soon

Smurf-IV:
@Wagnerp I think in order to get "Backward compatible with previous versions of TFM", _we_would need to use a krypton type for buttons etc. Like you did for the Icons. e.g.
`MessageBoxButtons` -> `KyrptonMessageBoxButtons`

Thuoghts ?

[...]

PR #784, opened 2022-07-16
* Implement point # 2 #728

Wagnerp:
- `MdiChildrenMinimizedAnchorBottom`

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
@@ -646,6 +646,24 @@ namespace Krypton.Toolkit
             }
         }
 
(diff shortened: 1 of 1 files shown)


Smurf-IV:
The "Idea" of these changes is to bring the functionality "Down" from .Net 6 / 7 into the earlier versions, so that a Single code base will work for multi-target builds.

Wagnerp:
> The "Idea" of these changes is to bring the functionality "Down" from .Net 6 / 7 into the earlier versions, so that a Single code base will work for multi-target builds.

We need to find out what the `MdiChildrenMinimizedAnchorBottom` option exactly does & has anyone modified it to work on older frameworks.

Smurf-IV:
@Wagnerp Is this still valid ?

Wagnerp:
> @Wagnerp Is this still valid ?

No, closing

PR #785, opened 2022-07-17
* Try to implement point 4 of #728

Wagnerp:
* Try to implement point 4 of #728

image


Smurf-IV:
Seems like the code for 4 buttons had been started, but now you have added button 5 ??

Wagnerp:
> Seems like the code for 4 buttons had been started, but now you have added button 5 ??

Yes, but I don't know what `_button4` is for, is it just for 'Ok'?

PR #807, opened 2022-08-23
* New `Apply` string for custom situations

Wagnerp:
* New `Apply` string for custom situations

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Utilities/General.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Utilities/General.cs
@@ -8,6 +8,8 @@
 
 namespace Krypton.Toolkit
 {
+    #region Enumerations
+
     /// <summary>Specifies constants defining the default button on a <seealso cref="KryptonMessageBox"/>.</summary>
     public enum KryptonMessageBoxDefaultButton
     {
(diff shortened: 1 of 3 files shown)


Source/Krypton Components/Krypton.Toolkit/Utilities/General.cs
```diff
@@ -29,4 +31,43 @@ public enum KryptonColorButtonCustomColorPreviewShape
...
+            var byteBuffer = new byte[2048];
+
+            using (var file = new FileStream(filePath, FileMode.Open, FileAccess.Read, FileShare.ReadWrite))
+            {
+                file.Read(byteBuffer, 0, byteBuffer.Length);
+            }
+
+            var headerPosition = BitConverter.ToInt32(byteBuffer, PE_HEADER_OFFSET);
+
+            var secondsSinceUNIXTimeStart = BitConverter.ToInt32(byteBuffer, headerPosition + LINKER_TIMESTAMP_OFFSET);
```
Smurf-IV, line 64: What is going to happen ion 2034 ?

Wagnerp: > What is going to happen ion 2034 ?

Need to find a better solution, as it defaults to 1/1/1970 at the moment

Issue #808, opened 2022-08-24
[Bug]: Button spacing is too small


kfcalf:
**Describe the bug**
The Button in Ribbon tab spacing is too small.
![Snipaste_2022-08-24_14-06-58](https://user-images.githubusercontent.com/28503355/186343157-ac3c95c8-f5ab-480c-bf83-d5a202540af8.png)

Smurf-IV:
There is now spacing around the image:
![image](https://user-images.githubusercontent.com/2418812/198869442-38688450-1c63-4b6d-aa05-4fc2b58f7678.png)

PR #825, opened 2022-10-30
- Make sure that the `Padding` is created before the control initialises with it

Smurf-IV:
#808
There is now pixel space "Around" the image:
image

Note:

What are these warnings??
image


Wagnerp:
@Smurf-IV The warnings are to do with .NET 5 not being supported anymore. Not sure if there's a way to ignore it... will look next week

Wagnerp:
Changelog?

Smurf-IV:
> Changelog?

Yupe.. Forgot again.. It's in the next PR for the disable Close button.

PR #832, opened 2022-10-31
* Fix configurations

Wagnerp:
* Fix configurations
* Removed .NET Core 3.1 for `Lite` configurations

image


Wagnerp:
* Updated log...

image

Issue #756, opened 2022-06-19
[Feature Request]: Add `[AllowNull]` to a controls `Text` field


Smurf-IV:
Part of the improvements in .net 6 / 7:
![image](https://user-images.githubusercontent.com/2418812/174472939-46629ccf-d934-416e-a646-a9aa2335cc21.png)

from here:
https://github.com/dotnet/winforms/commit/48616c47095149309b95caea9cc4e7c182313756

Wagnerp:
Making the project 'nullable' should help with this

Smurf-IV:
As part of this merge
https://github.com/Krypton-Suite/Standard-Toolkit/pull/933

The `AllowNull`able  attribute is now backward compatible with previous frameworks.

This will fix most of the following warnings in the compile:
![image](https://user-images.githubusercontent.com/2418812/216764619-6211bc9e-b4cc-4f16-aa3a-ca06c49a978c.png)

Wagnerp:
Reminder!!!!! This _should_ ought to be part of the extended toolkit as well ;)

Wagnerp:
![image](https://user-images.githubusercontent.com/949607/216767231-dc4a9623-c845-467c-90b8-a30a1a515059.png)

Wagnerp:
@Smurf-IV Disable `CS0436`?

Smurf-IV:
> ![image](https://user-images.githubusercontent.com/949607/216767231-dc4a9623-c845-467c-90b8-a30a1a515059.png)

The following should be preventing is from appearing in anything above core 3.1:
`#if NETFRAMEWORK || NETCOREAPP`
????

Smurf-IV:
> @Smurf-IV Disable `CS0436`?

I'll check what Frameworks I have installed; to make sure it is just the supported list ;-)

Smurf-IV:
Seems to be okay, I do not get the following

> @Smurf-IV Disable `CS0436`?

![image](https://user-images.githubusercontent.com/2418812/216771219-a72db341-809e-411f-954c-3342b505df59.png)

Smurf-IV:
@Wagnerp Why is there still netcoreapp3.1 in the Krypton support list ?

![image](https://user-images.githubusercontent.com/2418812/216771625-a268bcd2-c6a1-49ec-bea2-3b7018555243.png)

Smurf-IV:
> @Smurf-IV Disable `CS0436`?

This was caused by the following code
[...]

PR #944, opened 2023-02-12
- Changed All Standard toolkit projects to be

Smurf-IV:
- `<LangVersion>preview</LangVersion>`
  - `<Nullable>enable</Nullable>`
  - `<WarningLevel>6</WarningLevel>`
  - `<Optimize>True</Optimize>` for release style builds (i.e. nightly, canary, as well as normal release)
- Search for API's that return null and set there type to be nullable
- Trace back those API's and ensure nullable is applied to all derived API's Check all API's that Force return null and set to nullable type

Progress So far:
image

The number of warnings has gone up, because the other projects are now nullable and the level of warnings has "Gone up" to 6 on all

#756


Smurf-IV:
Notes:
- The is a work in progress
- I need to get the changes that you are dong into this ASAP so that they are not too divergent
- Ignore the branch name, it'll be changed on the next iteration
- I'll continue to work on this until the warnings are manageable

Wagnerp (approved):

PR #957, opened 2023-02-23
* Add .NET 8 support

Wagnerp:
* Add .NET 8 support
    - Instructions on how to enable beta/RC/preview versions of the SDK inside of stable versions of VS will come shortly


Wagnerp:
@Smurf-IV Only issue is: `error CS0117: 'FXAssembly' does not contain a definition for 'Version'`

PR #958, opened 2023-02-27
- Another (lite) weekends worth of fixing

Smurf-IV:
#756

image

I think I'm going to stop doing these, as it is a bit disheartening now that .net 8 is also included and made the numbers jump.

I'll keep plugging away at them as I work on "other" things instead. Thoughts ?


Wagnerp:
> #756
> 
> image
> 
> I think I'm going to stop doing these, as it is a bit disheartening now that .net 8 is also included and made the numbers jump.
> 
> I'll keep plugging away at them as I work on "other" things instead. Thoughts ?

@Smurf-IV Sure, as .NET 8 is unstable. I'm going to try to work out if we could use palette styles on a KForm, to change the background colour, rather than relying on KPanels to do the same thing. Thoughts?

Wagnerp (approved):

Smurf-IV:
>  I'm going to try to work out if we could use palette styles on a KForm, to change the background colour, rather than relying on KPanels to do the same thing. Thoughts?

Or, Just incorporate a Panel into the KForm by "Default", so when they are "Roundable" in a form, then it will be "Enabled" by default.

Wagnerp:
Will create a PR for it

Issue #872, opened 2022-12-17
[Feature Request]: ToolStrips in the non client area of a `KryptonForm`


Wagnerp:
Similar to how the quick access toolbar works in the `KryptonRibbon` control.

https://www.codeproject.com/Articles/55180/Extending-the-Non-Client-Area-in-Aero

https://www.codeproject.com/Articles/32623/Vista-Aero-ToolStrip-on-Non-Client-Area

Wagnerp:
Progress... 

![image](https://user-images.githubusercontent.com/949607/223206389-1404a319-db9e-4e4b-a7c9-e6f06cd2ee08.png)

@Smurf-IV Is there a way to 'shift' `ButtonSpecs` to the left after the window title?

Smurf-IV:
> @Smurf-IV Is there a way to 'shift' `ButtonSpecs` to the left after the window title?


Can you "Mock" something up please ?
Otherwise it's just the normal padding.
I think what you are asking is to have the following order:
1) App icon
2) App title
3) Buttons
4) Elastic space
5) Close window etc. buttons 

If so, then 3 and 4 cannot currently be done.

Wagnerp:
> Can you "Mock" something up please ?

Something like this...

![Mockup](https://user-images.githubusercontent.com/949607/223359150-ce98f8ce-495f-429c-936d-b3535b664bc7.png)

Wagnerp:
Completed 11/03/2023

PR #966, opened 2023-03-10
Alpha fr#872

Wagnerp:
* #872 
   - Theme specific icon sets coming soon

image


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonHelpCommand.cs
```diff
@@ -18,28 +18,29 @@ public class KryptonHelpCommand : KryptonCommand
...
 
         #endregion
 
         #region Public
 
-        public ButtonSpecAny HelpButton
+        /// <summary>Gets or sets the help button.</summary>
+        /// <value>The help button.</value>
+        [DefaultValue(null), Description(@"Access to the help button spec.")]
+        public ButtonSpecAny? HelpButton
```
Smurf-IV, line 30: This needs the `[AllowNull]` attribute to tell the compiler that it can accept null's but will not return them.

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonHelpCommand.cs
```diff
@@ -18,28 +18,29 @@ public class KryptonHelpCommand : KryptonCommand
...
         #region Identity
 
+        /// <summary>Initializes a new instance of the <see cref="KryptonHelpCommand" /> class.</summary>
         public KryptonHelpCommand()
         {
-            switch (KryptonManager.InternalGlobalPaletteMode)
-            {
-
-            }
+            _helpButtonSpec = null;
```
Smurf-IV, line 43: This is not needed as it already will be null when created.

Smurf-IV (changes requested):

Wagnerp:
> Will there also be a an example project going into the `Toolkit Demos` ?

Yes, just working on it : )

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonHelpCommand.cs
Wagnerp: @Smurf-IV Is there an example, so I can properly implement it?

Wagnerp: Addressed feedback

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonHelpCommand.cs
Wagnerp: Addressed feedback

PR #971, opened 2023-03-11
* `Office 2013 - White` now has correct control box icons

Wagnerp:
* `Office 2013 - White` now has correct control box icons

image

image


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonHelpCommand.cs
```diff
@@ -49,6 +49,50 @@ public class KryptonHelpCommand : KryptonCommand
          [AllowNull, Category(@"Appearance"), Description(@"State specific images for the help button."), DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
          public ButtonImageStates? ImageStates { get => _imageStates ?? new(); set { _imageStates = value; UpdateImageStates(KryptonManager.InternalGlobalPaletteMode); } }*/
 
+        [AllowNull]
```
Smurf-IV, line 52: These do not need `[AllonwNull]` because they can return null as well.
Also why are the `set`'s not expressions

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonHelpCommand.cs
Wagnerp: Addressed feedback and `set`'s are now `private` as images _should_ be theme based

Issue #1023, opened 2023-05-30
[Feature Request]: Please remove "sealed" from `KryptonWrapLabel` and `KryptonLinkWrapLabel`


MHartmann-427:
We are still using the original Krypton Toolkit coming from ComponentFactory/Phil Wright.
Because the original version written by Phil Wright has limitations and is not maintained anymore we plan to upgrade to Krypton-Suite/Standard-Toolkit.
Unfortunately we found one incompatibility to the original version, so we are not able to upgrade:
The control "KryptonWrapLabel" is implemented "sealed" in your version of Krypton ToolKit, so we can not inherit from this control (but we do).
Is there any need for sealing the class/control "KryptonWrapLabel"?

Wagnerp:
Hi @MHartmann-427 

Thanks for bringing this to our attention. I'll have a look into it as it might be an edit done by ReSharper. Will update you when  it's ready for testing.

Wagnerp:
Hi @MHartmann-427 

Please try this feature in build >= .156

Wagnerp:
Closing in 30 days (06/07/2023) if no response is recorded.

MHartmann-427:
I tried to test the changes you made for this request, but unfortunately version 80.23.6.156-alpha is not compatible with original Krypton Toolkit coming from ComponentFactory/Phil Wright anymore. 
The types "IPalette" and "KryptonPalette" do not exist, so it is not possible to update our application to version 80 of Krypton Toolkit. 
In version 70 the missing types "IPalette" and "KryptonPalette" still exist.
It seems that there will be breaking changes between version 70 and 80?

MHartmann-427:
I replaced all occurrences of KryptonPalette with KryptonCustomPaletteBase and all occurrences of IPalette with PaletteBase as mentioned in the documentation link provided.
Good news: Our solution compiles correctly without errors now.
Bad news: We will get a NullReferenceException on method "Import" of class KryptonCustomPaletteBase, so I still can't see if the actual controls render correctly.

Here's the code we use to import a palette XML-file that is embedded in the resources of the assembly:

        private static KryptonCustomPaletteBase ReadPalette(string paletteName)
        {
            KryptonCustomPaletteBase palette = new KryptonCustomPaletteBase();
            string resourceName = $"wb.resource.resources.{paletteName}.xml".ToLower();
            Assembly asm = typeof(Resources).Assembly;
[...]

PR #1024, opened 2023-06-04
* Implement #1023

Wagnerp:
* Implement #1023 

image


Source/Krypton Components/Krypton.Toolkit/Designers/Action Lists/KryptonTextBoxActionList.cs
```diff
@@ -40,13 +40,13 @@ public KryptonTextBoxActionList(KryptonTextBoxDesigner owner)
         /// <value>The Krypton Context Menu.</value>
         public KryptonContextMenu? KryptonContextMenu
         {
-            get => _textBox.KryptonContextMenu;
+            get => _textBox?.KryptonContextMenu;
 
             set
             {
-                if (_textBox.KryptonContextMenu != value)
+                if (_textBox?.KryptonContextMenu != value)
```
Smurf-IV, line 47: If text box is null, and value is not then line 51 will fail

Source/Krypton Components/Krypton.Toolkit/Designers/Action Lists/KryptonTextBoxActionList.cs
```diff
@@ -58,13 +58,13 @@ public KryptonTextBoxActionList(KryptonTextBoxDesigner owner)
         /// </summary>
         public PaletteMode PaletteMode
         {
-            get => _textBox.PaletteMode;
+            get => _textBox!.PaletteMode;
 
             set
             {
-                if (_textBox.PaletteMode != value)
+                if (_textBox?.PaletteMode != value)
```
Smurf-IV, line 65: if textbox is null and value is not, then line 68 will fail

Smurf-IV (changes requested):
The comment applies to all changes with the if statement detection, as it seems to be a pattern you have used

Source/Krypton Components/Krypton.Toolkit/Designers/Action Lists/KryptonTextBoxActionList.cs
Wagnerp: Reverted

Source/Krypton Components/Krypton.Toolkit/Designers/Action Lists/KryptonTextBoxActionList.cs
Wagnerp: Reverted

Wagnerp:
Reverted changes in `KryptonTextBoxActionList`

Issue #1020, opened 2023-05-22
[Bug]: Cannot add a `KryptonPage` to a `KryptonNavigator`


Wagnerp:
As found in today's nightly build, 142.

![KNavigator](https://github.com/Krypton-Suite/Standard-Toolkit/assets/949607/ec546cd2-72a7-4d2b-9222-533d61f6c53c)

Smurf-IV:
What version ov VS is that, perhaps they are starting to be more restrictive on the framework designer ??

Wagnerp:
> What version of VS is that?

Version 17.6.x, happens when using **both** .NET Framework & .NET

Smurf-IV:
I have been noticing that when trying to build / run the existing demo apps in the designer, that there are many errors.
I think it is an impact of the partial understanding of the `nullable` work that is still in progress.
OR, it is an impact of the designer moving forward forcing the rework required to use the new Visual Studio Way Of Working.

Wagnerp:
> I have been noticing that when trying to build / run the existing demo apps in the designer, that there are many errors. I think it is an impact of the partial understanding of the `nullable` work that is still in progress. OR, it is an impact of the designer moving forward forcing the rework required to use the new Visual Studio Way Of Working.

@Smurf-IV Think I've worked it out... _we're_ not checking for `null` when certain events are raised https://github.com/Krypton-Suite/Standard-Toolkit/issues/1023#issuecomment-1588810368

Smurf-IV:
Also found via the fix for https://github.com/Krypton-Suite/Theme-Palettes/issues/51

Smurf-IV:
It's gonna take a while to sort and fix all these ..
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/e3280827-9f58-4192-b42e-2aca78bb3d91)

Wagnerp:
I think @MHartmann-427 said that ReSharper can assist with this https://github.com/Krypton-Suite/Standard-Toolkit/issues/1023#issuecomment-1596772886

Smurf-IV:
It can, but first you have to modify all the `eventhandlers` to be nullable....

MHartmann-427:
No, ReSharper can assist you with this issue out-of-the-box.
Here's a small example:

![ReSharper](https://github.com/Krypton-Suite/Standard-Toolkit/assets/102691912/b719c033-656f-414e-92d7-337a15232447)


[...]

PR #1046, opened 2023-07-09
- completing checks for `?.Invoke(` usages

Smurf-IV:
- more nullable work
- 
#1020 

Note: I'll start to work on the bug when this has been merged (It'll probably be next weekend now)

Canary Build:
image

Alpha Build:
image


Smurf-IV:
The Canary build also complains about not have .net 47 etc. But they are installed:
image

PR #1048, opened 2023-07-15
* Alpha integratedtoolbarsptiii

Wagnerp:
* Only MS 365 themes work for the moment

image

ToolBar


Wagnerp:
@Smurf-IV Just picked up your question. This should not affect ribbon QAT features, as it only adds the 'new' buttonspec types to the KForm. Going to tidy up the new `KryptonIntegratedToolBarManager` more next week.

Issue #1050, opened 2023-07-15
[Feature Request]: Automatically upgrade incompatible theme (XML) files


Wagnerp:
Currently, there is a separate binary for  'Palette Upgrade Tool' (PUT) that the user must install to 'upgrade' palette XML files. To streamline things, it would be beneficial to bring this functionality directly into the toolkit, rather than having a standalone binary.

Smurf-IV:
I think it should just "Automatically" perform the upgrade within the toolkit usage.
And place the upgrade also into the "Palette Designer".
Leave the exceptions etc, so that the Toolkit "User" can catch them and then perform a designer upgrade.

Wagnerp:
> I think it should just "Automatically" perform the upgrade within the toolkit usage. And place the upgrade also into the "Palette Designer". Leave the exceptions etc, so that the Toolkit "User" can catch them and then perform a designer upgrade.

@Smurf-IV I'll work on these features ASAP

Smurf-IV:
> Automatic upgrading should be done within the `Import()` function, **but** we need to warn the user so that they can cancel it.

Only if there is a problem, otherwise it should be UI silent (maybe log out) and continue on it's way.

Wagnerp:
> > Automatic upgrading should be done within the `Import()` function, **but** we need to warn the user so that they can cancel it.
> 
> Only if there is a problem, otherwise it should be UI silent (maybe log out) and continue on it's way.

So, it needs to go in the `Import(bool silent)` function. Should the upgraded file go in the application's executable path, if so, that may cause issues if the system is locked down?

Smurf-IV:
> it needs to go in the Import(bool silent) function. 

Ok, That could be a way around the auto messagebox thing, that you are creating


> Should the upgraded file go in the application's executable path, if so, that may cause issues if the system is locked down?

Correct, it needs to be in memory, and then passed into what the import should have been using to activate the theme.

Wagnerp:
@Smurf-IV Might need some thinking, as I thought that I could compare the version with the defined 'maximum version', but then realised that you had to do XML parsing in the `Import()` function to get the right node (see lines 3028 - 3030 in `KryptonCustomPaletteBase` for an example). Feel free to give it a go.

Smurf-IV:
Will perform changes / tests via the following app 
[...]

PR #1054, opened 2023-07-16
* Begin work on #1050

Wagnerp:
* Begin work on #1050

image


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
```diff
@@ -3028,7 +3034,25 @@ private void ImportFromXmlDocument(XmlDocument doc)
 
                 if (version < CURRENT_PALETTE_VERSION)
                 {
-                    throw new ArgumentException($"Version '{version}' number is incompatible, only version {CURRENT_PALETTE_VERSION} or above can be imported.\nUse the PaletteUpgradeTool from the Application tab of the KryptonExplorer to upgrade.");
+                    //throw new ArgumentException($"Version '{version}' number is incompatible, only version {CURRENT_PALETTE_VERSION} or above can be imported.\nUse the PaletteUpgradeTool from the Application tab of the KryptonExplorer to upgrade.");
+
+                    DialogResult result = KryptonMessageBox.Show(
```
Smurf-IV, line 3039: I do not think it should do this, otherwise, every time a customer app is started it will be blocked until all this is run through.
The experience should be-
Load "Old xml file"
If version is "Old", run auto in memory upgrade.
If exception, then Messagebox and switch theme to default blue.

Therefore the customer user app will always run as expected, without the ned for weird Saving reloading. (i.e. the xml might be embedded and cannot be loaded !)

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
Wagnerp: Or just use the 'new' KMbox link label feature to redirect the user to the PD installer download?

Wagnerp:
- If a custom palette is deemed to be incompatible, it'll prompt you to upgrade it from the palette designer
- If you need to install the palette designer, you can download it straight from the prompt

image

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
Wagnerp: See provided animation

Smurf-IV:
> * If a custom palette is deemed to be incompatible, it'll prompt you to upgrade it from the palette designer
> 
>     * If you need to install the palette designer, you can download it straight from the prompt


No, this PR is missing the point.
It is supposed to be adding the "Auto upgrade" of a palette that is being passed into the import functionality.
- The xml may (should) be embedded in the application, therefore cannot be upgraded externally
- The workflow you have provided make this into a popup which will stop the app from loading and working normally
  - i.e. embedded never can be upgraded, and therefore the app is stuck
  - it's should not be enforced into the apps workflow, as this is supposed to be a framework, not an application suite.

Wagnerp:
> > * If a custom palette is deemed to be incompatible, it'll prompt you to upgrade it from the palette designer
> >   
> >   * If you need to install the palette designer, you can download it straight from the prompt
> 
> No, this PR is missing the point. It is supposed to be adding the "Auto upgrade" of a palette that is being passed into the import functionality.
> 
> * The xml may (should) be embedded in the application, therefore cannot be upgraded externally
> * The workflow you have provided make this into a popup which will stop the app from loading and working normally
>   
>   * i.e. embedded never can be upgraded, and therefore the app is stuck
>   * it's should not be enforced into the apps workflow, as this is supposed to be a framework, not an application suite.

Closing PR & reviewing

PR #1077, opened 2023-08-04
- Implemented via a new API, that takes a "Positionable" stream

Smurf-IV:
- So that if the xml is an application resource it, can still be upgraded in memory.
- Some Updates for nullables
- Update the PaletteExampleApp

#1050 

image


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
```diff
@@ -10,6 +10,12 @@
  */
 #endregion
 
+using System.Xml.Xsl;
+
+using Krypton.Toolkit.Properties;
+
+using static Krypton.Toolkit.PI;
+
```
Wagnerp, line 18: `using` statements should be moved to 'Global\GlobalDeclarations' (apart from 'static' ones) :).

Smurf-IV: These are "Only" used here, So will not put into global usings

Wagnerp (approved):

Issue #943, opened 2023-02-12
[Feature Request]: `ThemeManager` should be in a single file


Wagnerp:
The `ThemeManager` should be in a single file, as per discussion [here](https://github.com/Krypton-Suite/Standard-Toolkit/pull/936#discussion_r1103623023)

Smurf-IV:
Coming to a PR tomorrow..
![SingleThemName](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/4a560c8e-570f-4255-bf8d-72531d6f46f1)

PR #1078, opened 2023-08-04
- Make the `StringLookupConverter` typesafe

Smurf-IV:
- Start to Add code to have a single source of Theme strings.

#943

build report to come.


Smurf-IV:
@Wagnerp Why are the designer strings `Localizable`

These will never be seen by the App user, only by the developer, so I do not think that all the work that has been done to make these accessible via public interfaces will actually be seen by a Toolkit application !

Wagnerp (approved):

Wagnerp:
> @Wagnerp Why are the designer strings `Localizable`
> 
> These will never be seen by the App user, only by the developer, so I do not think that all the work that has been done to make these accessible via public interfaces will actually be seen by a Toolkit application !

@Smurf-IV I'll make them constants in a future PR

PR #1079, opened 2023-08-05
- Use a single file for the built in Theme strings, that also maps to the enums

Smurf-IV:
- Some spelling
- Fix a few more `null`able things

#943

image


Smurf-IV:
@Wagnerp 
What does this mean?
Why does the suite have any assets that need to be pulled in ?
image

And why now ?

Wagnerp:
> 

@Smurf-IV I tend to get this, running it a few times usually resolves it. Or 'flush' it before building.

Smurf-IV:
> @Smurf-IV I tend to get this, running it a few times usually resolves it. Or 'flush' it before building.


FYI:
I Deleted the *.json files to allow build to occur 
Seems like they are were old because they mentioned `lite` a lot.
They get recreated the next time VS is restarted.. but are not part of the Git repo.

Wagnerp (approved):

Issue #909, opened 2023-01-21
[Feature Request]: Can the colour picker support alpha blends


Wagnerp:
Can the colour picker used in some controls support alpha blends https://www.codeproject.com/Articles/14823/ColorEditorEx-An-extension-to-the-ColorEditor-to-s

Smurf-IV:
e.g.: another slider next to the existing one ?
![image](https://user-images.githubusercontent.com/2418812/213859254-4d1125d8-8571-4473-a2b8-d0616da8290e.png)
And an "option in the designer" to enable it, and programmatically as well ?

The trouble - with the layout - will be adding the new entry values above the "Add to Custom colours" button, and still getting the font usage to look good.

Smurf-IV:
A thought:
It might be possible to replace this area, with a masked edit (4 values) / text / and a slider
![image](https://user-images.githubusercontent.com/2418812/213859420-84980caa-fe16-4637-9f40-1e061be62378.png)

Reason being, is that the existing slider uses whatever space is available to draw the arrow, and the hit context for moving it.

Wagnerp:
I was thinking about the one found in the KColorButton, ContextMenus etc?

Wagnerp:
i.e. A slider in here maybe?

![image](https://user-images.githubusercontent.com/949607/213859859-5108fef5-95b8-4959-adea-6a212cd1aa96.png)

Smurf-IV:
> i.e. A slider in here maybe?


Deffo - not.

Wagnerp:
Is this something for extended?

Smurf-IV:
> Is this something for extended?

Normally I would have said yes, but now, so many colours are alpha blended I think it could / should be in standard

Wagnerp:
> A thought: It might be possible to replace this area, with a masked edit (4 values) / text / and a slider ![image](https://user-images.githubusercontent.com/2418812/213859420-84980caa-fe16-4637-9f40-1e061be62378.png)
> 
> Reason being, is that the existing slider uses whatever space is available to draw the arrow, and the hit context for moving it.

Maybe a better idea

Smurf-IV:
Coming to a PR soon:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/941073de-6588-4fcf-9c6e-f88f1cb4fb3f)

PR #1084, opened 2023-08-05
- Implement the alpha blending and return of colours via the `KryptonColourDialog`

Smurf-IV:
- Implement the alpha blending and return of colours via the `KryptonColourDialog`
- 
#909

image


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonColorDialog.cs
```diff
@@ -6,6 +6,9 @@
  */
 #endregion
 
+using System.ComponentModel;
+using System.Windows.Forms;
```
Wagnerp, line 10: Are these class specific, or can they go into `GlobalDeclarations`?

PR #1086, opened 2023-08-06
* Some refining

Wagnerp:
* Some refinements

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/General/KryptonCommandType.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/General/KryptonCommandType.cs
@@ -2,7 +2,7 @@
 /*
  *
  *   BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
- *  Modifications by Peter Wagner(aka Wagnerp) & Simon Coghlan(aka Smurf-IV), et al. 2017 - 2023. All rights reserved.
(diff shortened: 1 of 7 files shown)


Source/Krypton Components/Krypton.Toolkit/General/KryptonCommandType.cs
```diff
@@ -0,0 +1,31 @@
+﻿#region BSD License
+/*
+ *
+ *   BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
+ *  Modifications by Peter Wagner(aka Wagnerp) & Simon Coghlan(aka Smurf-IV), et al. 2017 - 2023. All rights reserved.
```
Smurf-IV, line 5: Incorrect start year

Smurf-IV (changes requested):

Smurf-IV:
Will / Are These buttons accessible in the palette designer as well ?

Wagnerp:
> Will / Are These buttons accessible in the palette designer as well ?

The ability to set custom images was done in the last PR, but I'll bring it back again.

Source/Krypton Components/Krypton.Toolkit/General/KryptonCommandType.cs
Wagnerp: Corrected start year

PR #1150, opened 2023-10-16
* Fixes the picturebox alignment

Wagnerp:
image

image

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonToolkitPoweredByControl.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonToolkitPoweredByControl.cs
@@ -43,8 +43,14 @@ namespace Krypton.Toolkit
 
         //public bool ShowWorkspaceVersion { get => _showWorkspaceVersion; set { _showWorkspaceVersion = value; Invalidate(); } }
 
-        public bool ShowThemeOption { get => _showThemeOption; set { _showThemeOption = value; Invalidate(); } }
+        /// <summary>Gets or sets a value indicating whether [show theme option].</summary>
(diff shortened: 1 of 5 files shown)


Smurf-IV:
Why is there "SO Much" space around the text ?

Smurf-IV:
Then they need to have null strings to make them "Invisible"
And set the height to be "Auto" as well

Wagnerp:
> Then they need to have null strings to make them "Invisible" And set the height to be "Auto" as well

Fixed sizing issues

Animation

Smurf-IV:
In the demo, When the "Them" option is chosen, it is crammed to the bottom, and does not follow the spacing of the previous intergaps.

Wagnerp:
> In the demo, When the "Them" option is chosen, it is crammed to the bottom, and does not follow the spacing of the previous intergaps.

Adjusted margins to give more space

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonToolkitPoweredByControl.cs
```diff
@@ -244,6 +244,11 @@ protected override void OnLoad(EventArgs e)
             base.OnLoad(e);
         }
 
+        protected override void OnCreateControl()
```
Smurf-IV, line 258: Why is this here

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonToolkitPoweredByControl.cs
Wagnerp: Was testing sizing. Will be cleaned up in next commit :)

PR #1160, opened 2023-10-29
Alpha

Wagnerp:
* Catchup


Smurf-IV:
@Wagnerp This has broken the "Test Form" project in Alpha.
i.e. the `kcmbBorderStyle` in Form1 is no longer initialised, and there fore throws an exception when starting the app!

Wagnerp:
> @Wagnerp This has broken the "Test Form" project in Alpha. i.e. the `kcmbBorderStyle` in Form1 is no longer initialised, and there fore throws an exception when starting the app!

@Smurf-IV Will remove in next PR

Issue #1091, opened 2023-08-08
[Bug]: Krypton File Dialogs Missing Buttons


Wagnerp:
As found in todays `alpha`

![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/949607/f39a5be1-6fc7-4835-b4f9-1ff5d4c1990e)

Smurf-IV:
Hmm..
Can you state what OS this is on please ?

Wagnerp:
> Hmm.. Can you state what OS this is on please ?

Using latest Windows 11

Wagnerp:
![1091](https://github.com/Krypton-Suite/Standard-Toolkit/assets/949607/41036289-be69-4304-98c2-ec9ec09be8cb)

Smurf-IV:
I have just tried this with todays alpha on Win 10 H22 x64 and various scaling sizes, and it appears to be okay.

Wagnerp:
> I have just tried this with todays alpha on Win 10 H22 x64 and various scaling sizes, and it appears to be okay.

Weird, wonder if it's Win 11 specific?

Smurf-IV:
> Weird, wonder if it's Win 11 specific?

I think so, What does a Win 11 Winform "File open dialog" look like?

Wagnerp:
> > Weird, wonder if it's Win 11 specific?
> 
> I think so, What does a Win 11 Winform "File open dialog" look like?

Like the Win 10 one...

![Win11OpenFileDialog](https://github.com/Krypton-Suite/Standard-Toolkit/assets/949607/da5f2d57-73fe-4cff-9c7b-1ad394598cbb)

Smurf-IV:
I'm Creating a VM for Win 11 then will start this investigation.

Smurf-IV:
@Wagnerp can you run the sample, because it appears to be "Okay" with the current "Alpha" builds:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/b0c920e0-fcc6-47bf-a6cd-24f6fc0dbf70)

Smurf-IV:
Mine is 22H2:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/4ac290d6-e482-425d-ab82-13b7500e11a5)

Wagnerp:
What happens if you call `Import/Export` via code?

Smurf-IV:
[...]

PR #1179, opened 2023-11-18
Alpha kmanstrings

Wagnerp:
* Strings + #1091

image


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonManager.cs
```diff
@@ -10,6 +10,8 @@
  */
 #endregion
 
+using System;
```
Smurf-IV, line 13: Why is this here ?

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonManager.cs
Wagnerp: Think it has been added by ReSharper, removed now

Source/Krypton Components/Krypton.Toolkit/Controls Visuals/KryptonAboutToolkitForm.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
Smurf-IV: These should not be needed

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Visuals/KryptonAboutToolkitForm.cs
Wagnerp: Not sure why they are showing up here, as they were all removed in the latest commit?

Smurf-IV: Just going through each change in order ....

PR #1182, opened 2023-11-19
* Some tidy up

Wagnerp:
* Some tidy up

image


Source/Krypton Components/Krypton.Toolkit/ButtonSpec/ButtonSpecView.cs
```diff
@@ -379,21 +379,19 @@ private void OnClick(object sender, MouseEventArgs e)
...
+                                             && ViewButton.SplitRectangle.Contains(e.Location));
 
                 if (performDefaultClick)
                 {
                     // Fire the event handlers hooked into the button spec click event
                     ButtonSpec.PerformClick(e);
                 }
 
-                if (showMenu)
+                // ToDo: Is this 'else' or remove brackets?
```
Smurf-IV, line 394: Remove brackets, so that it performs as per V70

Wagnerp: Removed brackets

Issue #1161, opened 2023-10-30
[Feature Request]: A proper `About` box


Wagnerp:
Krypton does not currently have a proper `About` dialog. Some items seen in #1009 could be used in a proposed `KryptonAboutBox` API, such as design, theming and the ability to get details of referenced assemblies. The question is, what to include?

PR #1186, opened 2023-11-19
* 1161 feature request a proper about box

Wagnerp:
* Work on #1161 

image


Source/Krypton Components/Krypton.Toolkit/View Base/InternalAssemblyDetails.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
```
Smurf-IV, line 1: These are not needed here

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/View Base/InternalAssemblyDetails.cs
Wagnerp: Done :)

Issue #1166, opened 2023-11-04
[Feature Request/Question]: Use a struct to contain `KryptonMessageBox` data


Wagnerp:
Is it possible to use a struct to contain `KryptonMessageBox` data, rather than having several `Show()` entries to choose from? Yes, this would be a big change from the standard WinForms API, but it might result in a cleaner codebase.

PR #1199, opened 2023-11-29
* Some cleanup + #1166

Wagnerp:
* Some cleanup + #1166

image


Source/Krypton Components/Krypton.Toolkit/Controls Visuals/VisualMessageBoxForm.cs
```diff
@@ -123,6 +126,52 @@ public VisualKryptonMessageBoxForm()
             UpdateSizing(showOwner);
         }
 
+        public VisualMessageBoxForm(KryptonMessageBoxData messageBoxData)
+        {
+            // Store incoming values
+            _text = messageBoxData.MessageText;
```
Smurf-IV, line 132: Why not just use the messageBox data struct as the storage ?

Source/Krypton Components/Krypton.Toolkit/View Base/TaskDialogMessageButton.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
```
Smurf-IV, line 1: Needs License statement and move the usings out

Smurf-IV (changes requested):

Wagnerp:
@Smurf-IV I think that we should keep the old `Show()` functions for compatibility, but maybe add the `Obsolete` flag to them. Thoughts?

Smurf-IV:
> @Smurf-IV I think that we should keep the old `Show()` functions for compatibility, but maybe add the `Obsolete` flag to them. Thoughts?

The ones that represent a winform MessageBox should be kept, And then they re-route to the struct call.
The ones that have been added to "Extend" the Param list should be set to obsolete.

Source/Krypton Components/Krypton.Toolkit/General/KryptonMessageBoxData.cs
```diff
@@ -18,7 +18,7 @@ public struct KryptonMessageBoxData
 
         /// <summary>Gets or sets the owner window.</summary>
         /// <value>The owner window.</value>
-        public IWin32Window? Owner { get; set; }
+        public IWin32Window? Owner { get; set; } = null;
```
Smurf-IV, line 21: That should not be necessary as the `?` defaults to null

Source/Krypton Components/Krypton.Toolkit/General/KryptonMessageBoxData.cs
```diff
@@ -30,75 +30,84 @@ public struct KryptonMessageBoxData
...
 
         /// <summary>Gets or sets the help information.</summary>
         /// <value>The help information.</value>
-        public HelpInfo? HelpInfo { get; set; }
+        public HelpInfo? HelpInfo { get; set; } = null;
 
         /// <summary>Gets or sets the show control copy.</summary>
         /// <value>The show control copy.</value>
-        public bool? ShowCtrlCopy { get; set; }
+        public bool? ShowCtrlCopy { get; set; } = false;
```
Smurf-IV, line 53: Why be a `?` if you are initing to a solid value ?

Source/Krypton Components/Krypton.Toolkit/Controls Visuals/VisualMessageBoxForm.cs
```diff
@@ -954,6 +956,47 @@ private void UpdateHelp()
             }
         }
 
+        /// <summary>
+        /// When the user clicks the Help button, the Help file specified in the helpFilePath parameter
+        /// is opened and the Help keyword topic identified by the keyword parameter is Displayed.
+        /// The form that owns the message box (or the active form) also receives the HelpRequested event.
+        /// </summary>
+        private void LaunchHelp(IWin32Window? owner)
```
Smurf-IV, line 964: Is there a test for this? There should already be a [Request] to have this function ality somewhere.

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/General/KryptonMessageBoxData.cs
Wagnerp: Removed default values

Source/Krypton Components/Krypton.Toolkit/Controls Visuals/VisualMessageBoxForm.cs
Wagnerp: Should be in the 'new' constructor.

Wagnerp:
> The ones that have been added to "Extend" the Param list should be set to obsolete.

New 'feature request' as a reminder?

Smurf-IV:
> > The ones that have been added to "Extend" the Param list should be set to obsolete.
> 
> New 'feature request' as a reminder?

Yes, go for it.

Issue #1187, opened 2023-11-20
[Feature Request]: Bring over the `KryptonCommandLinkButtons`


Wagnerp:
Is it time to bring over the `KryptonCommandLinkButtons` from the Extended Toolkit?

Wagnerp:
> Please re-open this, so that the "Class" is actually called `KryptonCommandLinkButton` please ?

@Smurf-IV There were 2 types, what should the second one be named as? Currently it's `KryptonAlternateCommandLinkButton` & the second one is `KryptonCommandLinkButton`.

Smurf-IV:
I thought only one of them was actually worth using in the Ext kit.
Does the demo use both ?

Wagnerp:
> I thought only one of them was actually worth using in the Ext kit. Does the demo use both ?

@Smurf-IV Yes, the one which looks like a normal KButton is the `KryptonAlternateCommandLinkButton` and vice-versa. I can switch the names around?

Smurf-IV:
I think it should replace the one that does not do stuff in the demo.
i.e. the one that works should have the expected name.

Wagnerp:
> I think it should replace the one that does not do stuff in the demo. i.e. the one that works should have the expected name.

I've 'hidden' the `KryptonAlternateCommandLinkButton` version, so it can be revisited at a later date if needed.

PR #1203, opened 2023-12-01
* Implement #1187

Wagnerp:
* Implement #1187 

image

image

Demos to come soon :)

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Designers/Designers/KryptonCommandLinkButtonDesigner.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Designers/Designers/KryptonCommandLinkButtonDesigner.cs
@@ -1,0 +1,47 @@
+﻿#region BSD License
+/*
+ *
+ *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
+ *  Modifications by Peter Wagner(aka Wagnerp) & Simon Coghlan(aka Smurf-IV), et al. 2023 - 2023. All rights reserved.
+ *
+ */
+#endregion
+
+namespace Krypton.Toolkit
+{
+    internal class KryptonCommandLinkButtonDesigner : ControlDesigner
+    {
+        #region Identity
+        /// <summary>
+        /// Initialize a new instance of the KryptonButtonDesigner class.
+        /// </summary>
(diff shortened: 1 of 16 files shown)


Source/Krypton Components/Krypton.Toolkit/Designers/Designers/KryptonCommandLinkButtonDesigner.cs
```diff
@@ -0,0 +1,47 @@
+﻿#region BSD License
+/*
+ *
+ *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
```
Smurf-IV, line 4: As these had already been developed in the Ext kit, then the start date year should be that.

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
```diff
@@ -268,65 +268,7 @@ public ButtonStyle ButtonStyle
 
         private void ResetButtonStyle() => ButtonStyle = ButtonStyle.Standalone;
 
-        [DefaultValue(false),
```
Smurf-IV, line 271: Is this a "breaking change"? and needs to be documented

Smurf-IV (changes requested):

Smurf-IV:
This will also need an equivalent in Ext kit to remove it from there, and document a breaking change

Source/Krypton Components/Krypton.Toolkit/Designers/Designers/KryptonCommandLinkButtonDesigner.cs
Wagnerp: Dates are now correct.

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
Wagnerp: Has now been put in `Readme`

Wagnerp:
> This will also need an equivalent in Ext kit to remove it from there, and document a breaking change

Will do, when next `nightly` is released

Issue #1148, opened 2023-10-15
[Feature Request]: MessegeBox CheckBox option


Wagnerp:
Currently testing in `Extended-Toolkit`, but will be implemented in v90

![Animation](https://github.com/Krypton-Suite/Standard-Toolkit/assets/949607/b35ea23d-abf7-4080-a692-979a1b473da2)

Wagnerp:
@Smurf-IV I was wondering if I should create 2 more `ShowCore` methods, `bool` & `CheckState` to return the value of the checkbox. Thoughts?

Wagnerp:
Maybe 'hook' into the existing `ShowCore` method to do a if/else?

Smurf-IV:
Or 
- extend the DialogResult to be a masked enum and have a state for the check state.
- Have an accessor for the CheckState
- No more parameters, there are already a load that should not be there (i.e forced optional and it confuses the intellisense !)

Wagnerp:
Might be the completely wrong way to do this, but:

```cs
 [Flags]
    public enum KryptonMessageBoxResult
    {
        None = DialogResult.None,
        Ok = DialogResult.OK,
        Cancel = DialogResult.Cancel,
        Abort = DialogResult.Abort,
        Retry = DialogResult.Retry,
        Ignore = DialogResult.Ignore,
        Yes = DialogResult.Yes,
        No = DialogResult.No,
        Checked = 8,
        CheckState = 9
    }
```

Smurf-IV:
Flags have to be bitfields, so yes that's wrong.
If the "other" fields are sequential, then they will need to be converted to bit as well.
And I would stick the "New ones" right at the end of the int field.

BTW what is "CheckState" supposed to be different from "Checked" ?

Wagnerp:
> BTW what is "CheckState" supposed to be different from "Checked" ?

Yes, the KCheckBox can have 3 states, `Unchecked`, `Intermediate` & `Checked`. Checked will be a 'normal' `bool` value`.

Wagnerp:
@Smurf-IV Should the KMessageBox be extended to support this? The current implementation only returns a `DialogResult`, maybe a custom enum with various options is needed, i.e. bool, three state, `DialogResult`, string, timeout etc, & call it `KryptonMessageBoxResponseType`?

Smurf-IV:
[...]

PR #1251, opened 2023-12-24
* Basic UI work for #1148

Wagnerp:
* Basic UI work for #1148

image


Source/Krypton Components/Krypton.Toolkit/General/Definitions.cs
```diff
@@ -2250,4 +2250,22 @@ public enum FadeSpeedChoice
...
+    {
+        None = DialogResult.None,
+        Ok = DialogResult.OK,
+        Cancel = DialogResult.Cancel,
+        Abort = DialogResult.Abort,
+        Retry = DialogResult.Retry,
+        Ignore = DialogResult.Ignore,
+        Yes = DialogResult.Yes,
+        No = DialogResult.No,
+        Checked = 8,
```
Smurf-IV, line 2266: Move these to higher numbers,
And Changed CheckState to be Intermediate>
Currently only 1 of these states can be returned

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/General/Definitions.cs
Wagnerp: @Smurf-IV Any numbers to avoid, i.e. can they be 31 & 32?

Smurf-IV: It's an integer, so go big ;-)

Wagnerp: @Smurf-IV Should it be a `uint` as all default values come from [here](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-messagebox)?

Smurf-IV: No, As they are Return values, and as such should be `int` according to that link.
But consider this scenario..
- Question is `MB_CANCELTRYCONTINUE` based
- Checkbox is set as an option
- How does it return a Continue with a checkbox intermediate state set ?

In other words, I do not think returning the checkstate is going to work.
And only the `get` CheckState is required as an API !

Wagnerp: There might be another way, just use `Checked` as it's a bool. Checked would be yes & unchecked would be no?

Wagnerp: Used higher numbers (1006 & 1007)

Smurf-IV:
Remove the ProgressBar updates, as they belong to another Request, that is now in Ext ??

Wagnerp:
> Remove the ProgressBar updates, as they belong to another Request, that is now in Ext ??

Ready to go

Wagnerp:
@Smurf-IV This is now ready to merge :)

PR #1267, opened 2024-01-05
* Resolves #1249

### LINKED ISSUE #1249: [Bug]: Ribbon Form Bars are no longer Drawn with the theme colouring
**Describe the bug**
image

**To Reproduce**
It's been like this since 2023-05-21 (The last time I created a release build of the Ribbon Gallery Example)

- In the design (Using todays Alpha) you can see the Theme colouring,
- When the application is run, it has the OS's Control Title frame BUT with the Orb drawn over it !

Wagnerp:
* Resolves #1249 

image

Needs to be tested

### DIFF
--- a/Source/Krypton Components/Krypton.Ribbon/Controller/GalleryButtonController.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/Controller/GalleryButtonController.cs
@@ -25,7 +25,7 @@ namespace Krypton.Ribbon
         private bool _pressed;
         private bool _mouseOver;
         private NeedPaintHandler? _needPaint;
-        private readonly Timer? _repeatTimer;
+        private readonly Timer _repeatTimer;
 
         #endregion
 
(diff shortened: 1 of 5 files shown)


Source/Krypton Components/Krypton.Ribbon/Controller/GalleryButtonController.cs
```diff
@@ -25,7 +25,8 @@ internal class GalleryButtonController : GlobalId,
         private bool _pressed;
         private bool _mouseOver;
         private NeedPaintHandler? _needPaint;
-        private readonly Timer _repeatTimer;
+        private readonly Timer? _repeatTimer;
```
Smurf-IV, line 28: Why is this nullable ?
If it is set to null in Dispose, then that's okay !

Wagnerp: It crashed in testing. Maybe use `!`?

Smurf-IV: That means something else is wrong.

Wagnerp: Now fixed, the timer wasn't disposed of

PR #1307, opened 2024-01-30
* Add missing case to `PaletteMicrosoft365SilverDarkMode.cs`

Wagnerp:
* Add missing case to `PaletteMicrosoft365SilverDarkMode.cs`
* Some Ribbon designer implementations

image

### DIFF
--- a/Source/Krypton Components/Krypton.Ribbon/Palette/RibbonAppButton.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/Palette/RibbonAppButton.cs
@@ -408,7 +408,7 @@ namespace Krypton.Ribbon
 
                     if (_ribbon.CaptionArea != null)
                     {
-                        _ribbon.TabsArea.AppButtonVisibleChanged();
+                        _ribbon.TabsArea?.AppButtonVisibleChanged();
                         _ribbon.CaptionArea.AppButtonVisibleChanged();
                         _ribbon.CaptionArea.PerformFormChromeCheck();
                     }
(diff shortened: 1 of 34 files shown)


Source/Krypton Components/Krypton.Ribbon/Palette/RibbonAppButton.cs
```diff
@@ -437,7 +437,7 @@ public bool FormCloseBoxVisible
                     if (_ribbon.CaptionArea != null)
                     {
                         _ribbon.CaptionArea.KryptonForm!.CloseBox = value;
-                        _ribbon.CaptionArea.PerformFormChromeCheck();
+                        _ribbon.CaptionArea?.PerformFormChromeCheck();
```
Smurf-IV, line 440: Why was this change necessary, as the `CaptionArea` is used on the previous line "Without" the null check.

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Ribbon/Palette/RibbonAppButton.cs
tobitege: Actually the "KryptonForm" in line 439 should be checked for null at that location, not the CaptionArea as the outer "if" already assures that.

Wagnerp: Undone, but what's causing #1308?

tobitege: > Undone, but what's causing #1308?

Exactly what I wrote in my above comment and in the issue report:
**_ribbon.CaptionArea.KryptonForm** is null, thus needs to be checked for null before assignment

PR #1317, opened 2024-02-05
Fix KryptonCustomPaletteBase import's image assignment

tobitege:
Fix exception in ImportObjectFromElement: do not assign string to image prop, but the image referenced by its name, if it exists in the imageCache dictionary.


Wagnerp:
Hi @tobitege 

Thanks for your contribution, would you mind pushing the changes to the `alpha` branch please, so that they can undergo testing?

tobitege:
Sure thing

Issue #1316, opened 2024-02-05
[Bug]: [master] KryptonCustomPaletteBase Import fails if XML contains images


tobitege:
# **Issue**

Trying to import an XML theme file that has images defined in it for e.g ButtonSpecs, the import fails with below exception:  
"Object of type "System.String" cannot be converted to type "System.Drawing.Image""

Stacktrace (sorry, in German)
```
   bei System.RuntimeType.TryChangeType(Object value, Binder binder, CultureInfo culture, Boolean needsSpecialCast)
   bei System.Reflection.MethodBase.CheckArguments(Object[] parameters, Binder binder, BindingFlags invokeAttr, CultureInfo culture, Signature sig)
   bei System.Reflection.RuntimeMethodInfo.InvokeArgumentsCheck(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   bei System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
   bei System.Reflection.RuntimePropertyInfo.SetValue(Object obj, Object value, Object[] index)
   bei Krypton.Toolkit.KryptonCustomPaletteBase.ImportObjectFromElement(XmlElement element, ImageReverseDictionary imageCache, Object obj) in D:\github\Standard-Toolkit\Source\Krypton Components\Krypton.Toolkit\Controls Toolkit\KryptonCustomPaletteBase.cs:Zeile 3388.
   bei Krypton.Toolkit.KryptonCustomPaletteBase.ImportObjectFromElement(XmlElement element, ImageReverseDictionary imageCache, Object obj) in D:\github\Standard-Toolkit\Source\Krypton Components\Krypton.Toolkit\Controls Toolkit\KryptonCustomPaletteBase.cs:Zeile 3361.
   bei Krypton.Toolkit.KryptonCustomPaletteBase.ImportObjectFromElement(XmlElement element, ImageReverseDictionary imageCache, Object obj) in D:\github\Standard-Toolkit\Source\Krypton Components\Krypton.Toolkit\Controls Toolkit\KryptonCustomPaletteBase.cs:Zeile 3361.
[...]

PR #1315, opened 2024-02-04
* Resolved #1280

Wagnerp:
* Resolves #1280 

image

PR #1318, opened 2024-02-05
Fix KryptonCustomPaletteBase import's image assignment

tobitege:
Fix exception in ImportObjectFromElement: do not assign string to image prop, but the image referenced by its name, if it exists in the imageCache dictionary.

### DIFF
--- a/Documents/Help/Changelog.md
+++ b/Documents/Help/Changelog.md
@@ -3,6 +3,7 @@
 =======
 
 ## 2024-11-xx - Build 2411 - November 2024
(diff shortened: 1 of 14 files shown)


tobitege:
relates to #1316

Wagnerp (approved):

Wagnerp:
@Smurf-IV I've updated the changelog in PR #1315 

cc. @tobitege

tobitege:
wth did I do? I tried to get this removed on my local machine only :(

PR #1331, opened 2024-02-12
Fix white themes, theme combo and propertygrid.

tobitege:
Hey @Wagnerp @Smurf-IV 
please forgive some code reformattings in between, mostly indentations.
This should fix 3 issues:
- White themes (2010, 2013, 365) had white menu text. This was a bit tricky as I had to change the color table "_table" to become protected to be able to override the method for menu item text color in new derived classes for the white themes to use.
- The PropertyGrid showed white-on-white/unreadable color combos in several themes. Adapated to use for the given background a contrast color as per method from another class. It's either black or white font and I tested it with all builtin palettes and they now all show readable content.
- Revised the ThemeComboBox' handling of the DefaultPalette as the index was operating on 2 disjunct sources and thus causing a mismatch. Also made it UI safer to wait for handle to exist.
- See the TestForm's "Theme" dialog for above: once you open it the themes correctly synchronize across the 2 forms as they should and the property grid shows up fine in all themes, too.

### DIFF
--- a/Documents/Help/Changelog.md
+++ b/Documents/Help/Changelog.md
@@ -3,6 +3,7 @@
 =======
 
 ## 2024-11-xx - Build 2411 - November 2024
+* Resolved [#1331](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1331), Fix white menu text in White themes (2010, 2013, 365); fixes to `KryptonPropertyGrid` and `KryptonThemeComboBox` with regard to theme switching
 * Resolved [#1313](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1313), White background in tabs area
 * Implement [#1309](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1309), Is it time to bring over `KryptonOutlookGrid`
 * Resolved [#1316](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1316), `KryptonCustomPaletteBase` Import fails if XML contains images (fix courtesy of [tobitege](https://github.com/tobitege))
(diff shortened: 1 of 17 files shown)


Wagnerp (approved):

Wagnerp:
Hi @tobitege 

Looks good, would you mind updating the changelog to include the changes that you've made? (basically copy & paste what you've wrote in the description)

tobitege:
> Hi @tobitege
> 
> Looks good, would you mind updating the changelog to include the changes that you've made? (basically copy & paste what you've wrote in the description)

Hmm I'd rather not mess with that, tbh. The only changelog file I see is from last November.

There are still annoying issues with the ThemeComboBox to be looked into, I hope you guys can take another peek:
- serialization of the DefaultPalette just doesn't work bugfree
- constructor seems to reset it each time you open the form designer
- if it was saved, it might not load on next attempt to open the form due to an exception

tobitege:
@Wagnerp 
about the ThemeComboBox: right now I don't see the use for the ThemeSelectedIndex property and if that shouldn't just be removed and its references just be replaced with the _selectedIndex variable?
I know it isn't hidden in the master branch (yet), but it has no convincing purpose as its value is read-only and as we saw, it's not in a one-to-one relationship with the PaletteMode.

tobitege:
About latest commit for KryptonPropertyGrid:
In InitColours() commented out the assignment of the normalFont as the reproducably causes unrecoverable exceptions (crash to desktop) when switching current theme back and forth.
Across the builtin palettes, I didn't observe a noticable variation of font settings that would suffer from this change for this one component.

Wagnerp:
Hi @tobitege 

Yes, it should be `SelectedIndex`

Wagnerp (approved):

Issue #1314, opened 2024-02-04
[Bug]: CheckedListBox CheckedIndices NullRef


iwarp:
**Describe the bug**
Just converting an existing app using Krypton to the standard-toolkit, when looping thru the CheckedIndices of a CheckedListBox I get a null ref issue in multiple places across my code base. 

**StackTrace:**
   at Krypton.Toolkit.KryptonCheckedListBox.InternalCheckedListBox.InnerArrayIndexOfIdentifier(Object identifier, Int32 stateMask) in C:\git\Krypton-Standard-Toolkit\Source\Krypton Components\Krypton.Toolkit\Controls Toolkit\KryptonCheckedListBox.cs:line 759
   at Krypton.Toolkit.KryptonCheckedListBox.CheckedIndexCollection.InnerArrayIndexOfIdentifier(Object identifier, Int32 stateMask) in C:\git\Krypton-Standard-Toolkit\Source\Krypton Components\Krypton.Toolkit\Controls Toolkit\KryptonCheckedListBox.cs:line 153
   at Krypton.Toolkit.KryptonCheckedListBox.CheckedIndexCollection.get_Item(Int32 index) in C:\git\Krypton-Standard-Toolkit\Source\Krypton Components\Krypton.Toolkit\Controls Toolkit\KryptonCheckedListBox.cs:line 127
   at Krypton.Toolkit.KryptonCheckedListBox.CheckedIndexCollection.CopyTo(Array array, Int32 index) in C:\git\Krypton-Standard-Toolkit\Source\Krypton Components\Krypton.Toolkit\Controls Toolkit\KryptonCheckedListBox.cs:line 70
   at Krypton.Toolkit.KryptonCheckedListBox.CheckedIndexCollection.GetEnumerator() in C:\git\Krypton-Standard-Toolkit\Source\Krypton Components\Krypton.Toolkit\Controls Toolkit\KryptonCheckedListBox.cs:line 81
   at KryptonTest.Form1.kryptonButton1_Click(Object sender, EventArgs e) in 

**To Reproduce**
Simple form with CheckedListBox with Designer added items.

`
foreach(var item in this.kryptonCheckedListBox1.CheckedIndices) {
    MessageBox.Show(item.ToString());
}
`

**Expected behavior**
Not to crash and return the selected indexes back to my foreach

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version: 10.0.22621.3007
[...]

PR #1368, opened 2024-03-03
- Sort out access to the correct named static fields that have been renamed in `netCore`

Smurf-IV:
- Sort out access to the correct named static fields that have been renamed in `netCore`

- Sort usage of the `#if NET6...` define usages

#1314

image


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonSaveFileDialog.cs
```diff
@@ -25,7 +25,7 @@ public class KryptonSaveFileDialog : FileSaveDialogWrapper, IDisposable
         /// <inheritdoc />
         protected override DialogResult ShowActualDialog(IWin32Window? owner) => _internalSaveFileDialog.ShowDialog(owner);
 
-#if NET7_0 || NET8_0
+#if NET7_0 || NET8_0_OR_GREATER
```
Wagnerp, line 28: Should this be `#if NET7_0_OR_GREATER`?

Wagnerp (changes requested):

Wagnerp (approved):

Issue #1281, opened 2024-01-11
[Feature Request]: Implement User Input Types


Wagnerp:

Smurf-IV:
Can you put in an example pictures and a task for each of the "Types" please ?

Wagnerp:
> Can you put in an example pictures and a task for each of the "Types" please ?

Yes, sure, it'll be stuff like textbox, combobox, numeric up/down etc

Smurf-IV:
@Wagnerp This has a merge. Is it still open ?
And there are no images added !

Wagnerp:
> @Wagnerp This has a merge. Is it still open ? And there are no images added !

Still open, I haven't done the RTL versions yet

Wagnerp:
Working on this next week :)

PR #1391, opened 2024-04-04
* Some testing

Wagnerp:
* Some testing for #1281 

image

**Note:** Cannot currently define a user input to use, will be fixed in a future PR


Source/Krypton Components/TestForm/UserInputToastNotificationTest.cs
```diff
@@ -1,12 +1,274 @@
-﻿using Krypton.Toolkit;
+﻿using System;
```
Smurf-IV, line 1: Why are these here ?

Smurf-IV: Also These new files needs to have the License header added.

Smurf-IV (changes requested):

Wagnerp:
Will review on Thursday, as I'm currently away until Wednesday

Source/Krypton Components/TestForm/UserInputToastNotificationTest.cs
Wagnerp: Sorted

Source/Krypton Components/TestForm/BasicToastNotificationTest.cs
```diff
@@ -1,4 +1,16 @@
-﻿namespace TestForm
+﻿#region BSD License
+/*
+ * 
+ * Original BSD 3-Clause License (https://github.com/ComponentFactory/Krypton/blob/master/LICENSE)
```
Smurf-IV: Does not need the `ComponentFactory` part, as these are all new

Smurf-IV (changes requested):

Issue #1356, opened 2024-02-26
[Bug]: `AppButton` colours don't change while switching themes


Wagnerp:
While testing #1343, it has been discovered that `AppButton` colours don't change while switching themes. Looks as if it's not repainting somewhere.

![AppButton Colours](https://github.com/Krypton-Suite/Standard-Toolkit/assets/949607/a9e26be6-dcc4-474c-91a8-b3a05b4a0439)

Smurf-IV:
I would expect the following to have an input:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/1fc9550e-0557-4468-b581-fb4c2bc9dc95)

But even when they are empty a colour is still being found from somewhere.

Smurf-IV:
Narrowed it down to these Palette values:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/93071c3c-d6fc-4623-bd6e-fd486946c339)

Nothing to do with the designer values above?!?

Smurf-IV:
To get this:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/cd219f7f-e60c-490d-a12c-e4335982e0dd)

Set these things in the designer:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/b9fde4cd-7820-48eb-916d-aad916d5bb3f)

Smurf-IV:
To get this:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/cc11d693-9c56-43d5-b2f2-2200ffc51d3f)

Set the following in the designer:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/a9060ac8-6825-4d9b-8b6a-e0c9246c33f2)

Which means that the `RibbonAppButton` is really the QATButton !
So the Explanation text needs to change because changing the Designer text will break all the serialisation of that object !!

Smurf-IV:
@Wagnerp Proposal:
Because the `##Dark` and `##Light` values are currently not overridden (Used) in the serialisation of the Palettes:
- [] Remove from their current "General" location in the designer
- [] Add new Area and values to the `State###` area with an appropriate named Key: i.e. `RibbonFileAppButton`
  - Use the existing values as the defaults, but not "Hard Code" them !
[...]

PR #1392, opened 2024-04-05
Done so far:

Smurf-IV:
- [x] Remove from their current "General" location in the designer
- [x] Add new Area and values to the `State###` area with an appropriate named Key: i.e. `RibbonFileAppButton`
  - [x] Use the existing values as the defaults, but not "Hard Code" them !
- [x] Wire up the usage of Disabled etc usages to use the `##Dark` and `##Light` overrides to follow the correct pattern
- [x] Update the Breaking Changes area to state correct usages of the `RibbonFileAppButton` values
- [x] Fix Examples
- [x] Fix Theme Template editor

#1356

image


Smurf-IV:
Converted to draft: As the Toolkit examples have issues (Not sure if they existed before, but sorting anyway !)

Smurf-IV:
image
Ready now

PR #1395, opened 2024-04-06
- Alpha: Restore the code to Propogate the serilasation steps back into the UI

Smurf-IV:
#1394

image


Wagnerp:
Somthing's not right here, as it's blocking the merge

Smurf-IV:
> Somthing's not right here, as it's blocking the merge

Conflict due to 2 updates to the Changelog ! - Sorted now

Issue #1418, opened 2024-04-22
[Other Issues]: Cleanup warnings KT/Ribbon/Controller


giduac:
Keeping track of files handled in the topic start and possible discussion about this can be done in this thread.

2024-05-22 | 1418-cleanup-warnings-0043 | PR 
Warnings: 

```
1418-cleanup-warnings-0042 | PR 1487 | Warnings 4 | Removes the publickey warnings CS1700
1418-cleanup-warnings-0041 | PR 1486 | Warnings 32 | Cleanup remaining warnings.
1418-cleanup-warnings-0031 | PR 1474 | Warnings 81 Revisits previously disabled throw statements.
1418-cleanup-warnings-0040 | PR 1472 | Warnings 402 | Check on CommonHelper
1418-cleanup-warnings-0030 | PR 1471 | Warnings 420 | Several warnings removed
1418-cleanup-warnings-0027 | PR 1459 | Warnings 429 | Toolkit third part
1418-cleanup-warnings-0026 | PR 1458 | Warnings 3014 | Toolkit second part
1418-cleanup-warnings-0025 | PR 1457 | Warnings 12922 | Toolkit first part
1418-cleanup-warnings-0020 | PR 1453 | Warnings 17614 | Workspace complete
1418-cleanup-warnings-0012 | PR 1451 | Warnings 19721 | Navigator second part (count is off, build without merging alpha)
1418-cleanup-warnings-0011 | PR 1446 | Warnings 19273 | Navigator first part
1418-cleanup-warnings-0010 | PR 1445 | Warnings 20224 | Docking module
1418-cleanup-warnings-0007 | PR 1444 | Warnings before: 20258 | Warnings after: 20061
1418-cleanup-warnings-0006 | PR 1443 | Warnings before: 26097 | Warnings after: 20258
1418-cleanup-warnings-0005 | PR 1440 | Warnings before: 29131 | Warnings after: 26097
1418-cleanup-warnings-0004 | PR 1438 | Warnings before: 29869 | Warnings after: 29131
1418-cleanup-warnings-0003 | PR 1437 | Warnings before: 30125 | Warnings after: 29869
1418-cleanup-warnings-0002 | PR 1429 | Warnings before: 30386 | Warnings after: 30164
1418-cleanup-warnings-0001 | PR 1421 | Warnings before: 30666 | Warnings after :30386
```

giduac:
Hi @Wagnerp 
[...]

PR #1421, opened 2024-04-23
1418-cleanup-warnings-0001

giduac:
First serious PR :)

Before the changes there were 30666 warnings and 30386 afterwards = -280.
compile-results

Probably you have a look at the added string that is used in the throw.
Possibly that can be done via a globally used one with a param position to format....

Your comments please.

### DIFF
--- a/Source/Krypton Components/Krypton.Ribbon/Controller/AppButtonController.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/Controller/AppButtonController.cs
@@ -30,6 +30,7 @@ namespace Krypton.Ribbon
         private bool _fixedPressed;
         private bool _hasFocus;
         private readonly Timer _updateTimer;
+        protected readonly string ribbonTabsAreaCannotBeNull = "Property ribbon.TabsArea cannot be null";
         #endregion
 
         #region Events
@@ -250,10 +251,15 @@ namespace Krypton.Ribbon
         /// <param name="e">A KeyEventArgs that contains the event data.</param>
         public void KeyDown(Control c, KeyEventArgs e)
         {
-            ViewBase newView = null;
+            ViewBase? newView = null;
             var ribbon = (KryptonRibbon)c;
 
-            switch (e.KeyData)
+            if ( ribbon.TabsArea is null )
+            {
+                throw new NullReferenceException( this.ribbonTabsAreaCannotBeNull );
+            }
+
+            switch ( e.KeyData)
             {
                 case Keys.Tab:
                 case Keys.Right:
@@ -373,6 +379,11 @@ namespace Krypton.Ribbon
         /// <param name="ribbon">Reference to owning ribbon instance.</param>
         public void KeyTipSelect(KryptonRibbon ribbon)
         {
+            if ( ribbon.TabsArea is null )
+            {
+                throw new NullReferenceException( this.ribbonTabsAreaCannotBeNull );
+            }
+
             // We leave key tips usage whenever we use the application button
             ribbon.KillKeyboardKeyTips();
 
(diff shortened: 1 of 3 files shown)


Wagnerp (approved):

Source/Krypton Components/Krypton.Ribbon/Controller/AppButtonController.cs
```diff
@@ -250,10 +251,15 @@ public virtual void LostFocus([DisallowNull] Control c)
         /// <param name="e">A KeyEventArgs that contains the event data.</param>
         public void KeyDown(Control c, KeyEventArgs e)
         {
-            ViewBase newView = null;
+            ViewBase? newView = null;
             var ribbon = (KryptonRibbon)c;
```
Smurf-IV, line 255: Use `as`

Source/Krypton Components/Krypton.Ribbon/Controller/AppButtonController.cs
```diff
@@ -250,10 +251,15 @@ public virtual void LostFocus([DisallowNull] Control c)
...
 
-            switch (e.KeyData)
+            if ( ribbon.TabsArea is null )
```
Smurf-IV, line 257: Because `ribbon` may be null now (Due to `as`) then also check it in this `if` check please ?

Apply these changes to the other locations as well

Smurf-IV (changes requested):

giduac:
Hi @Smurf-IV , @Wagnerp 
Leaving the ribbon out of the check does create a warning in this case, but:
In the first approach I had two ifs, one for the ribbon and next the tabarea so there are two throws with a distinctive each

Using two ifs like that might be clearer when the ribbon would be null at first. Like this:

```cs
            if ( ribbon is null )
            {
                throw new NullReferenceException( this.ribbonCannotBeNull );
            }

            if ( ribbon.TabsArea is null )
            {
                throw new NullReferenceException( this.ribbonTabsAreaCannotBeNull );
            }
``
??

**Comment by giduac** [comment]
Hi @Smurf-IV ,

If you have time can you please have a look at this to move forward with this?

File: Source/Krypton Components/Krypton.Ribbon/Controller/AppButtonController.cs [hunk 3]
```diff
@@ -379,9 +383,14 @@ public void KeyPress(Control c, KeyPressEventArgs e)
         /// <param name="ribbon">Reference to owning ribbon instance.</param>
         public void KeyTipSelect(KryptonRibbon ribbon)
```
Smurf-IV, line 384: If ribbon can truely be null at this point, then change the parameter to be a nullable type

Source/Krypton Components/Krypton.Ribbon/Controller/AppTabController.cs
```diff
@@ -369,9 +373,14 @@ public void KeyPress(Control c, KeyPressEventArgs e)
         /// <param name="ribbon">Reference to owning ribbon instance.</param>
         public void KeyTipSelect(KryptonRibbon ribbon)
         {
-            if ( ribbon.TabsArea is null )
+            if (ribbon is null)
```
Smurf-IV, line 376: If truely nullabel at this point, then change the parameter type to be nullable

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Ribbon/Controller/AppTabController.cs
giduac: @Smurf-IV 
Both files updated.

giduac: @Smurf-IV 
Thank you!

Wagnerp:
> Hi @Smurf-IV ,
> 
> If you have time can you please have a look at this to move forward with this?

Hi @giduac 

It has now been merged

PR #1443, opened 2024-04-30
1418 cleanup warnings 0006

giduac:
1418-cleanup-warnings-0006
This concludes the Ribbon.
Overall warnings down by a third.

compile-results

### DIFF
--- a/Source/Krypton Components/Krypton.Ribbon/Palette/ContextTabSet.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/Palette/ContextTabSet.cs
@@ -21,7 +21,7 @@ namespace Krypton.Ribbon
     {
         #region Instance Fields
 
-        private ViewDrawRibbonTab? _lastTab;
+        private ViewDrawRibbonTab _lastTab;
 
         #endregion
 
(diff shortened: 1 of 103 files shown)


Source/Krypton Components/Krypton.Ribbon/Palette/ContextTabSet.cs
```diff
@@ -21,7 +21,7 @@ internal class ContextTabSet
     {
         #region Instance Fields
 
-        private ViewDrawRibbonTab _lastTab;
+        private ViewDrawRibbonTab? _lastTab;
```
Smurf-IV, line 24: I do not like these sort of changes.. Because if the class is just ignoring the nullable by the use of `!`. then why have it has a nullable object ?

Please back these out.

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Ribbon/Palette/ContextTabSet.cs
giduac: Thanks for reviewing, code corrected as requested including a couple of additional modifications I came across.
Warnings slightly reduced. 

See https://github.com/Krypton-Suite/Standard-Toolkit/pull/1443/commits/a63cae17d8b1c6a031da7d6c97b54002b87bb8f9

compile-results

PR #1451, opened 2024-05-05
1418-cleanup-warnings-0012

giduac:
1418-cleanup-warnings-XXXX

Part 2  Navigator module
Concludes this module.

compile-results


giduac:
Hang on please, just looking into the merge conflict. (weird since i just fetch alpha before restoring the stash....)

giduac:
Not sure why this is conflicting. Just updated the branch from upstream alpha, restored the stash and made the commit without problems. None of the adaptations were there before???

But to resolve this is, here's the top part of the Layout method as it should be.

```cs
        public override void Layout([DisallowNull] ViewLayoutContext context)
        {
            Debug.Assert(context is not null);

            if ( context is null)
            {
                throw new ArgumentNullException(nameof(context));
            }
```

Which is then followed by the exiting code:
```cs
            // We take on all the available display area
            ClientRectangle = context.DisplayRectangle;
            . . . . 
````

giduac:
Wasn't directly aware that I can edit the conflicting code myself...

PR #1453, opened 2024-05-06
1418-cleanup-warnings-0020

giduac:
1418-cleanup-warnings-0020

This concludes the Workspace module.

compile-results

### DIFF
--- a/Source/Krypton Components/Krypton.Workspace/Controls Workspace/KryptonWorkspace.cs
+++ b/Source/Krypton Components/Krypton.Workspace/Controls Workspace/KryptonWorkspace.cs
@@ -2804,7 +2804,7 @@ namespace Krypton.Workspace
                                      Point splitter)
         {
             // Get the sequence that contains the items moved
-            var parentSequence = separator.WorkspaceItem.WorkspaceParent as KryptonWorkspaceSequence;
+            var parentSequence = (KryptonWorkspaceSequence)separator.WorkspaceItem.WorkspaceParent!;
             SeparatorToItems(separator, out IWorkspaceItem after, out IWorkspaceItem? before);
 
             // At design time we can get null references
@@ -2825,7 +2825,7 @@ namespace Krypton.Workspace
                 if (offset != 0)
                 {
                     // Update the sizing value for each item in the sequence
-                    foreach (Component child in parentSequence!.Children!)
+                    foreach (Component child in parentSequence.Children!)
                     {
                         // Can only process IWorkspaceItem items
                         if (child is IWorkspaceItem item)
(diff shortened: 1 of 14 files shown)


Source/Krypton Components/Krypton.Workspace/Controls Workspace/KryptonWorkspace.cs
```diff
@@ -2804,7 +2804,7 @@ internal Rectangle SeparatorMoveBox(ViewDrawWorkspaceSeparator separator)
                                      Point splitter)
         {
             // Get the sequence that contains the items moved
-            var parentSequence = (KryptonWorkspaceSequence)separator.WorkspaceItem.WorkspaceParent;
+            var parentSequence = separator.WorkspaceItem.WorkspaceParent as KryptonWorkspaceSequence;
```
Smurf-IV, line 2807: If the var is just going to be ignoring the nullable possibility from using the `as`
then use a cast, and remove the `!` usage later on.

Source/Krypton Components/Krypton.Workspace/Controls Workspace/KryptonWorkspace.cs
```diff
@@ -2899,7 +2899,7 @@ internal void InternalPageDragStart(object sender, KryptonNavigator? navigator,
                     }
 
                     // Calling DragStart will cause the drag targets to be created from the target providers
-                    e.Cancel = !_dragManager.DragStart(e.ScreenPoint, new PageDragEndData(sender as KryptonNavigator, e.Pages));
+                    e.Cancel = !_dragManager.DragStart(e.ScreenPoint, new PageDragEndData((sender as KryptonNavigator)!, e.Pages));
```
Smurf-IV, line 2902: Either 
- check the nullability from the `as` before sending
- DragStart to accept null's
- Use a cast.

Option 2 preferred.

Smurf-IV (changes requested):
Apply comments on casting / null check / null ignore `!` override to all code changes please ?

Source/Krypton Components/Krypton.Workspace/Controls Workspace/KryptonWorkspace.cs
giduac: @Smurf-IV 
Dragstart already does a null check on the dragEndData param and will throw if null.
The PageDragEndData class expects a valid source param.

This way the sender is checked and dragstart handles the throw
```cs
PageDragEndData? dragEndData = sender is null
    ? null
    : new PageDragEndData((sender as KryptonNavigator)!, e.Pages);

e.Cancel = !_dragManager.DragStart(e.ScreenPoint, dragEndData);
```

Your thoughts please ?

Smurf-IV: The `sender is null` should have the following check in it `(sender as KryptonNavigator)`
ReSharper will then change it to a null check pattern along the lines of `(sender is KryptonNavigator kn)` and pass the `kn` into the constructor.

PR #1459, opened 2024-05-10
1418-cleanup-warnings-0027

giduac:
1418-cleanup-warnings-0027

Toolkit third part
This is the last bulk of warnings removed.

The last 115 will need closer attention to be removed.

compile-results

### DIFF
--- a/Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingFloatingWindow.cs
+++ b/Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingFloatingWindow.cs
@@ -213,7 +213,7 @@ namespace Krypton.Docking
         protected override void LoadDockingElement(XmlReader xmlReader, KryptonPageCollection pages)
         {
             // Grab the requested size and location
-            Point location = CommonHelper.StringToPoint(xmlReader.GetAttribute(@"L"));
+            Point location = CommonHelper.StringToPoint(xmlReader.GetAttribute(@"L")!);
             Size clientSize = CommonHelper.StringToSize(xmlReader.GetAttribute(@"S")!);
 
             // Find the size of the floating window borders
--- a/Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingFloatingWindow.cs
+++ b/Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingFloatingWindow.cs
@@ -213,7 +213,7 @@ namespace Krypton.Docking
         protected override void LoadDockingElement(XmlReader xmlReader, KryptonPageCollection pages)
         {
             // Grab the requested size and location
-            Point location = CommonHelper.StringToPoint(xmlReader.GetAttribute(@"L")!);
+            Point location = CommonHelper.StringToPoint(xmlReader.GetAttribute(@"L"));
             Size clientSize = CommonHelper.StringToSize(xmlReader.GetAttribute(@"S")!);
 
             // Find the size of the floating window borders
--- a/Source/Krypton Components/Krypton.Navigator/View Draw/ViewDrawNavCheckButtonBase.cs
+++ b/Source/Krypton Components/Krypton.Navigator/View Draw/ViewDrawNavCheckButtonBase.cs
@@ -114,9 +114,9 @@ namespace Krypton.Navigator
             : base(stateDisabled, stateNormal, stateTracking,
                    statePressed, null, null, orientation, true)
         {
-            Debug.Assert(navigator != null);
+            Debug.Assert(navigator is not null);
 
-            Navigator = navigator;
+            Navigator = navigator ?? throw new ArgumentNullException(nameof(navigator));
             _page = page;
             _lastClick = DateTime.Now.AddDays(-1);
 
@@ -400,8 +400,9 @@ namespace Krypton.Navigator
             KeyController = _buttonController;
 
             // Create two decorators in order to support tooltips and hover events
-            var toolTipController = new ToolTipController(Navigator.ToolTipManager, this, _buttonController);
-            var hoverController = new ToolTipController(Navigator.HoverManager, this, toolTipController);
+            var toolTipController = new ToolTipController(Navigator.ToolTipManager!, this, _buttonController);
+            var hoverController = new ToolTipController(Navigator.HoverManager!, this, toolTipController);
+
             return hoverController;
         }
         #endregion
--- a/Source/Krypton Components/Krypton.Ribbon/Group Contents/KryptonRibbonGroupComboBox.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/Group Contents/KryptonRibbonGroupComboBox.cs
@@ -413,8 +413,7 @@ namespace Krypton.Ribbon
         [Category(@"Appearance")]
         [Description(@"Text associated with the control.")]
         [Editor(typeof(MultilineStringEditor), typeof(UITypeEditor))]
-        [AllowNull]
-        public string? Text
+        public string Text
         {
             get => ComboBox.Text;
             set => ComboBox.Text = value;
--- a/Source/Krypton Components/Krypton.Ribbon/View Draw/ViewDrawRibbonGroupClusterButtonText.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/View Draw/ViewDrawRibbonGroupClusterButtonText.cs
@@ -100,7 +100,17 @@ namespace Krypton.Ribbon
         /// <param name="context">Layout context.</param>
         public override Size GetPreferredSize([DisallowNull] ViewLayoutContext context)
         {
-            Debug.Assert(context != null);
+            Debug.Assert(context is not null);
+
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (context.Renderer is null)
+            {
+                throw new ArgumentNullException(nameof(context.Renderer));
+            }
 
             // Validate incoming reference
             if (context == null)
@@ -148,6 +158,16 @@ namespace Krypton.Ribbon
         {
             Debug.Assert(context is not null);
 
+            if (context is null)
+            {
+                throw new ArgumentNullException(nameof(context));
+            }
+
+            if (context.Renderer is null)
+            {
+                throw new ArgumentNullException(nameof(context.Renderer));
+            }
+
             // We take on all the available display area
             ClientRectangle = context!.DisplayRectangle;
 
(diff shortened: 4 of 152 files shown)


Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingFloatingWindow.cs
```diff
@@ -213,7 +213,7 @@ public override void SaveElementToXml(XmlWriter xmlWriter)
         protected override void LoadDockingElement(XmlReader xmlReader, KryptonPageCollection pages)
         {
             // Grab the requested size and location
-            Point location = CommonHelper.StringToPoint(xmlReader.GetAttribute(@"L"));
+            Point location = CommonHelper.StringToPoint(xmlReader.GetAttribute(@"L")!);
```
Smurf-IV, line 216: Change `CommonHelper`s to take a null item, and they then return a default if is null

Source/Krypton Components/Krypton.Navigator/View Draw/ViewDrawNavCheckButtonBase.cs
```diff
@@ -114,9 +114,9 @@ internal abstract class ViewDrawNavCheckButtonBase : ViewDrawButton,
             : base(stateDisabled, stateNormal, stateTracking,
                    statePressed, null, null, orientation, true)
         {
-            Debug.Assert(navigator != null);
+            Debug.Assert(navigator is not null);
 
-            Navigator = navigator;
+            Navigator = navigator ?? throw new ArgumentNullException(nameof(navigator));
```
Smurf-IV, line 119: I prefer this style to the `if( ## is null)` usage

Source/Krypton Components/Krypton.Ribbon/Group Contents/KryptonRibbonGroupComboBox.cs
```diff
@@ -413,8 +413,7 @@ public Size MaximumSize
         [Category(@"Appearance")]
         [Description(@"Text associated with the control.")]
         [Editor(typeof(MultilineStringEditor), typeof(UITypeEditor))]
-        [AllowNull]
-        public string? Text
+        public string Text
```
Smurf-IV, line 416: If you look in the Decompiled source for net 6 onwards, I think this can return null.

Source/Krypton Components/Krypton.Ribbon/View Draw/ViewDrawRibbonGroupClusterButtonText.cs
```diff
@@ -100,7 +100,17 @@ public void MakeDirty()
         /// <param name="context">Layout context.</param>
         public override Size GetPreferredSize([DisallowNull] ViewLayoutContext context)
         {
-            Debug.Assert(context != null);
+            Debug.Assert(context is not null);
+
+            if (context is null)
```
Smurf-IV, line 105: Why throw an exception, when the null condition is being dealt with below ?

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonComboBox.cs
```diff
@@ -1357,8 +1357,9 @@ public new Padding Padding
         /// <summary>
         /// Gets and sets the text associated with the control.
         /// </summary>
-        public override string? Text
+        public new virtual string Text
         {
+            // Replaced override with "new virtual" since inheritance keeps producing null warnings which ever way
             get => _comboBox.Text;
```
Smurf-IV, line 1363: This can be null in the newer frameworks...

Smurf-IV (changes requested):
Got 1/2 way through.
Please review the overrides that have been modified to return base items. Because in the newer frameworks they can return nullable items.

Smurf-IV:
@giduac 
Please also revisit "**ALL**" designers and remove the exception throwing code
Because when opening a designer (Lot's of things will be null)
And the designers are _**unusable**_:

image

Smurf-IV:
Just open the `main.cs` from inside the TestForm project in the suite, and you will get things like this:
image

Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingFloatingWindow.cs
giduac: Changed as requested.

The line (217) below this one is done similarly
```cs
Size clientSize = CommonHelper.StringToSize(xmlReader.GetAttribute(@"S")!);
```
Now CommonHelper has more of these one-liners that work in a similar manner.
I leave that for now and put CommonHelper on the list to be revisted later on for a closer look.

Source/Krypton Components/Krypton.Navigator/View Draw/ViewDrawNavCheckButtonBase.cs
giduac: Only applies of course if there's an assignment in the code. 
Mostly done in this way. Will take this further.

Source/Krypton Components/Krypton.Ribbon/Group Contents/KryptonRibbonGroupComboBox.cs
giduac: Underlying base properties are all non-nullable string. Older code of course.

Winforms.Control.Text 
	non-nullable
	gets: always a string
	sets: an empty string if value is null
	
Winforms.ComboBox: (inherits from Control.Text)
	non-nullable
	gets: always a string
	sets: can take a null value (which make no sense in relation to Control.Text's behaviour)

Either leave it as is now, since under water it will always set to a string of return at least an empty string.
So null make no sense. And other solutions are imo merely a work-around.

Or allow null input on a inherited non-nullable type like this.
```cs
        /// <summary>
        /// Gets and sets the text associated with the control.
        /// </summary>
        [Category(@"Appearance")]
        [Description(@"Text associated with the control.")]
        [Editor(typeof(MultilineStringEditor), typeof(UITypeEditor))]
        [AllowNull]
        public string Text
        {
            get => ComboBox.Text;
            set => ComboBox.Text = value!;
        }
```

Source/Krypton Components/Krypton.Ribbon/View Draw/ViewDrawRibbonGroupClusterButtonText.cs
giduac: All the DebugAssert statement, throughout the code base, cause [DisallowNull] to be invalidated and the warnings reappear, that X could dereference a possible null value.

using context.Renderer causes warnings without a null check.
```cs
        public override Size GetPreferredSize([DisallowNull] ViewLayoutContext context)
        {
            Debug.Assert(context is not null);

            // Validate incoming reference
            if (context is null)
            {
                throw new ArgumentNullException(nameof(context));
            }

            if (context.Renderer is null)
            {
                throw new ArgumentNullException(nameof(context.Renderer));
            }
```

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonComboBox.cs
giduac: This hooks up with my earlier explanation on 
`[Source/Krypton Components/Krypton.Ribbon/Group Contents/KryptonRibbonGroupComboBox.cs](https://github.com/Krypton-Suite/Standard-Toolkit/pull/1459/files/1c2450229996836dbc697259fc8b7034a0d5cf4d#diff-cd2e7f5863063dd2fd805078ad0f5c60ac4a82b081311edef75278f303ec47cd)`

giduac:
> @giduac Please also revisit "**ALL**" designers and remove the exception throwing code Because when opening a designer (Lot's of things will be null) And the designers are _**unusable**_:
> 
> image

Will do.


> @giduac Please also revisit "**ALL**" designers and remove the exception throwing code Because when opening a designer (Lot's of things will be null) And the designers are _**unusable**_:
> 
> image

Will do, and add these to the commit addressing the earlier corrections.

giduac:
> > @giduac Please also revisit "**ALL**" designers and remove the exception throwing code Because when opening a designer (Lot's of things will be null) And the designers are _**unusable**_:
> 
> Will do, and add these to the commit addressing the earlier corrections.
> 
> @Smurf-IV, Do you mean designers or palettes like the one in the screenshot?

I've been looking into this. 
It is already present in alpha and running some tests in a branch from alpha.
Not sure if the main.cs (form) in testform is not having some difficulties of it's own, we'll see.

All needpaint "throws" have been disabled.
Testform runs, only main form has some problems with the main panel,.
Clicking "ignore & continue" does show the form. although it won't work at runtime.
Which if another form is added and given a main panel, all is well....
But I'm getting there (I guess) :)

giduac:
The [last commit](https://github.com/Krypton-Suite/Standard-Toolkit/pull/1459/commits/39a4734361355b8f5d3c749683712621753c7b50) fixes the breaking and contains the corrections.

giduac:
Got my head around all the nullable attributes and reverted the earlier property headers that had been changed to `public new virtual` back to `public override`. Applied attributes as needed.
All in the [this commit](https://github.com/Krypton-Suite/Standard-Toolkit/pull/1459/commits/3ce977ea9bc17a6bd2244d76cdf19025b92b496e)

Smurf-IV:
> Main.cs from TestForm has a kcombobox on it that seems to cause the problems. If you ignore the error in the designer and remove the combo and add a new combobox, all is fine. All other forms run from the startscreen.cs work.

This is not really acceptable for opening existing projects.
So it will need to be sorted ASAP. Please raise a new bug so that it can be worked on, outside the scope of the warnings fixing(s)

giduac:
The error in testform is back.
Is connected (at this point) to the kcombobox on the main form. If KCB is added to startscreen the same error occurs.
Tried the following:
- add a button to startscreen that launches main form.
- if the active theme is: MS Off 365 Blue it crashes.
- If the theme is changes to something else it doesn't.

> > Main.cs from TestForm has a kcombobox on it that seems to cause the problems. If you ignore the error in the designer and remove the combo and add a new combobox, all is fine. All other forms run from the startscreen.cs work.
> 
> This is not really acceptable for opening existing projects. So it will need to be sorted ASAP. Please raise a new bug so that it can be worked on, outside the scope of the warnings fixing(s)

Will do.

giduac:
> > Main.cs from TestForm has a kcombobox on it that seems to cause the problems. If you ignore the error in the designer and remove the combo and add a new combobox, all is fine. All other forms run from the startscreen.cs work.
> 
> This is not really acceptable for opening existing projects. So it will need to be sorted ASAP. Please raise a new bug so that it can be worked on, outside the scope of the warnings fixing(s)

Will do.

Issue #1475, opened 2024-05-22
[Bug]: Build Scripts will run when no suitable environment is detected. | Add BinLog option to build-nightly.cmd


giduac:
**Describe the bug** [PR 1476](https://github.com/Krypton-Suite/Standard-Toolkit/pull/1476)
When no suitable environment is detected by the build scripts will still try to run.

Excerpt from `build-nightly.cmd`.
```batch
@echo off

if exist "%ProgramFiles%\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin" goto vs17prev
if exist "%ProgramFiles%\Microsoft Visual Studio\2022\Enterprise\MSBuild\Current\Bin" goto vs17ent
if exist "%ProgramFiles%\Microsoft Visual Studio\2022\Professional\MSBuild\Current\Bin" goto vs17pro
if exist "%ProgramFiles%\Microsoft Visual Studio\2022\Community\MSBuild\Current\Bin" goto vs17com
if exist "%ProgramFiles%\Microsoft Visual Studio\2022\BuildTools\MSBuild\Current\Bin" goto vs17build

echo "Unable to detect suitable environment. Check if VS 2022 is installed."

pause

:vs17prev
set msbuildpath=%ProgramFiles%\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin
goto build

```
### Proposed solution:
Add a `goto exitbatch` statement after `pause`.
Add the `:exitbatch` label at the end of the file.

### Add the "/bl" parameter to the msbuild line of  `build-nightly.cmd`
```batch
"%msbuildpath%\msbuild.exe" -t:%targets% nightly.proj /fl /flp:logfile=build.log /bl
```
This will output, next to the build.log, a msbuild.binlog that can be viewed/examined with ["MSBuild Binary and Structured Log Viewer"](https://msbuildlog.com/).

giduac:
Hi @Wagnerp,
Please assign to me.
TY!

Wagnerp:
> Hi @Wagnerp, Please assign to me. TY!

Hi @giduac 

Done, you might want to bump the version up. Currently it's 1.9.

giduac:
@Wagnerp 
Where is the version located and what would the next version number be?
Then I'll add it to the PR.

Wagnerp:
> @Wagnerp Where is the version located and what would the next version number be? Then I'll add it to the PR.

Hi @giduac 

[...]

PR #1476, opened 2024-05-22
1475-Build Scripts will run when no suitable environment is detected

giduac:
Build Scripts will run when no suitable environment is detected.
Add the /BL parameter to build-nightly.cmd.
Issue 1475: https://github.com/Krypton-Suite/Standard-Toolkit/issues/1475

compile-results

### DIFF
--- a/Scripts/build-nightly.cmd
+++ b/Scripts/build-nightly.cmd
@@ -9,6 +9,7 @@ if exist "%ProgramFiles%\Microsoft Visual Studio\2022\BuildTools\MSBuild\Current
 echo "Unable to detect suitable environment. Check if VS 2022 is installed."
 
 pause
+goto exitbatch
 
 :vs17prev
 set msbuildpath=%ProgramFiles%\Microsoft Visual Studio\2022\Preview\MSBuild\Current\Bin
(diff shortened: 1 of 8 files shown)


Scripts/build-nightly.cmd
```diff
@@ -39,7 +40,7 @@ for /f "tokens=* usebackq" %%A in (`tzutil /g`) do (
 @echo
 set targets=Build
 if not "%~1" == "" set targets=%~1
-"%msbuildpath%\msbuild.exe" -t:%targets% nightly.proj /fl /flp:logfile=build.log 
+"%msbuildpath%\msbuild.exe" -t:%targets% nightly.proj /fl /flp:logfile=build.log /bl
```
Wagnerp, line 43: Should the `/bl` flag also be in the `build-canary` & `build-stable` scripts?

giduac: Can be done of course.
Objective was to start with Nightly.

giduac: @Wagnerp 
On second thought it wont hurt putting it in there straight away.
So imo, let's put the switch in those two as well...

giduac: @Wagnerp 
Added both.

Issue #1478, opened 2024-05-22
[Bug]: Toolkit Nightly Alpha - KryptonHeaderGroup doesn't disappear


JeanPasquini:
### **Describe the bug**
Error message when I try to delete the KryptonHeaderGroup. After to try delete, the component doesn't disapper in the form, so I need to open again the Visual Studio.

Screenshots:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/126198701/00105614-b5e7-4289-a5a6-b879aecb1b76)
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/126198701/4187032d-22c8-4d66-a48a-9d894e3dfb60)
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/126198701/9551e962-c8a7-4c69-85e6-2265a1947079)

-------------

### **To Reproduce**
1. Put the KryptonHeaderGroup in form
2. Try exclude the component
3. Error message

--------------

### **Expected behavior**
Delete the component without leaving Visual Studio.

--------------
### **Desktop:**

 - Visual Studio (17.9.7)
 - OS: Windows 11 (V23H2)
 - Framework/.NET Version: 4.8
 - Toolkit Nightly Version: 90.24.5.141-alpha

giduac:
@JeanPasquini 

I'm having a look into this. It's the result of ongoing work.
It can take a bit before a change is merged.

giduac:
Hi @JeanPasquini,

I got it fixed so far and will do some more testing tomorrow.
Probably will be in next week's Nightly build.
I'll let you know here when the PR goes in.

giduac:
@Wagnerp 
Please assign to me.
TY!

Wagnerp:
> @Wagnerp Please assign to me. TY!

@giduac 

Now assigned to you

JeanPasquini:
Is this 'Nightly' version available to contribute?

Wagnerp:
> Hi @Wagnerp,
> 
> The KListview class uses the KTreeviewDesigner, this KTreeviewDesigner has it's own action list which does not seem compatible with KListview. I tried an older stable release (80.23.11.321) and there the action list doesn't show in the designer as well.
> 
> The Class header as it is now:
> 
> ```cs
>     [ToolboxItem(true)]
>     [ToolboxBitmap(typeof(ListView))]
>     [DefaultEvent("AfterSelect")]
[...]

PR #1480, opened 2024-05-23
1478-KryptonHeaderGroup-doesnt-disappear

giduac:
[1478-KryptonHeaderGroup-doesnt-disappear](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1478#issuecomment-2127572426)
- Throw statements removed. Even when vars are non-nullable that doesn't mean they can't be null. Inherited GetService() method can return null, and sometimes does. That's what happened in this call.
- Approach (casting and nullabilty) throughout the Action Lists slightly differed, all made equal. Applying null-forgiving and null-conditional operators.
- All cast in the constructors are done via AS.
- KListview was using the KTreeviewDesigner, now set to KlistviewDesigner and the corresponding action list
- Tested all controls on TestForm to see if they are removable now.

compile-results


Source/Krypton Components/Krypton.Navigator/Navigator/KryptonPageActionList.cs
```diff
@@ -144,13 +144,13 @@ public string ToolTipBody
         /// </summary>
         public Bitmap? ImageSmall
         {
-            get => _page.ImageSmall;
+            get => _page!.ImageSmall;
 
             set
             {
-                if (_page.ImageSmall != value)
+                if (_page!.ImageSmall != value)
```
Smurf-IV, line 151: if page is null and value is null, then line 154 will still throw it's toys out of the pram !

Smurf-IV (changes requested):
As the statement made, "All follow the same pattern" then the review comment will need to be checked in all areas.
Or, Page is not nullable, as it is being ignored everywhere any way !

Wagnerp:
Please also update the changelog, so that fixes can be tracked. Thanks.

Source/Krypton Components/Krypton.Navigator/Navigator/KryptonPageActionList.cs
giduac: I also noticed that , the code has been setup like that from the start and will fail in most areas if the cast from `owner.Component` to KControl should return null. Seems that this has always worked so far.

So `owner.Component` always has to be a valid reference for the action list code to run.
Putting in a check on the type cast of 'owner.Component' and throw will make use of the null-forgiving operator redundant. The control var can become non-nullable. Will process all Action Lists Like in the example below.
```cs
        private readonly KryptonPage _page;
        private readonly IComponentChangeService? _serviceComponentChange;
        private DesignerActionItemCollection _actions;
        #endregion

        #region Identity
        /// <summary>
        /// Initialize a new instance of the KryptonPageActionList class.
        /// </summary>
        /// <param name="owner">Designer that owns this action list instance.</param>
        public KryptonPageActionList(KryptonPageDesigner owner)
            : base(owner.Component)
        {
            // Remember designer and actual component instance being designed
            _page = owner.Component as KryptonPage ?? throw new NullReferenceException(GlobalStaticValues.ParameterCannotBeNull(nameof(owner.Component)));

            // Cache service used to notify when a property has changed
            _serviceComponentChange = GetService(typeof(IComponentChangeService)) as IComponentChangeService;
        }
        #endregion
        
        #region Public
        /// <summary>
        /// Gets and sets the page text.
        /// </summary>
        public string TextShort
        {
            get => _page.Text;

            set
            {
                if (!_page.Text.Equals(value))
                {
                    _serviceComponentChange?.OnComponentChanged(_page, null, _page.Text, value);
                    _page.Text = value;
                }
            }
        }

```

giduac:
> Please also update the changelog, so that fixes can be tracked. Thanks.

Will update the changelog when this is signed off.

Smurf-IV:
> > Please also update the changelog, so that fixes can be tracked. Thanks.
> 
> Will update the changelog when this is signed off.

Please add the update for the changelog at the same time as when the code changes are done, then it is not forgotten or split into 2 PR's for the same issue

Thanks

giduac:
> > > Please also update the changelog, so that fixes can be tracked. Thanks.
> > 
> > 
> > Will update the changelog when this is signed off.
> 
> Please add the update for the changelog at the same time as when the code changes are done, then it is not forgotten or split into 2 PR's for the same issue
> 
> Thanks

Yes, just waiting till the changes are acccepted.
Change log added.

Smurf-IV:
@giduac Conflicts ?

giduac:
> @giduac Conflicts ?
The other PR interfered...
Should work now!

Issue #1414, opened 2024-04-18
[Bug / Enhancement]: `SetDate` API is missing from `KryptonMonthCalendar`


Wagnerp:
There are some missing elements in the `KryptonMonthCalendar`, which are present in the WinForms version. See https://github.com/Krypton-Suite/Standard-Toolkit/discussions/1412 for details.

Missing Method:
- `public void SetDate(DateTime date)` : Winform has it _only_ in `MonthCalendar`

Possible affected classes:
- `KryptonDateTimePicker`
- `KryptonContextMenuMonthCalendar`
- `KryptonDataGridViewDateTimePickerColumn`
- `KryptonMonthCalendar`

Notes:
- `public event DateRangeEventHandler DateSelected` is done in `KryptonMonthCalendar` as per Winform equivalent
- `public DateTime Value`: Done in `KryptonDateTimePicker` as per Winform equivalent.

PR #1495, opened 2024-05-28
- Add missing function(s)

Smurf-IV:
- Update Warning to latest
- Fix a few Resharper warnings along the way

#1414


Smurf-IV:
@Wagnerp Just a note.. The timestamp being produced in the output stated `GMT` but it's not !

Issue #1388, opened 2024-04-02
[Bug]: Todays alpha: Using the Toolkit "TestForm" shows that the drop button icon is not drawn


Smurf-IV:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/b4969b63-f48e-47ef-998c-90769260f331)

giduac:
Hi @Smurf-IV 

Also KDropButton was affected by this and the change to ButtonValues.cs will make KDB behave the same as KButton.
PR underway.

Smurf-IV:
@Wagnerp Should this go into 8.5 as well ?

Wagnerp:
> @Wagnerp Should this go into 8.5 as well ?

@Smurf-IV Yes it should

Smurf-IV:
@giduac Please recreate the change against the v85 branch and create a PR for that (Also include updates to the changelog)

giduac:
@Smurf-IV 
Can I pop the same commits in there as in alpha ?

Smurf-IV:
Yes, of course. (If they fit ;-))

giduac:
@Smurf-IV 
PR posted: https://github.com/Krypton-Suite/Standard-Toolkit/pull/1503

giduac:
Hi @Smurf-IV & @Wagnerp 

FYI.
In this fix there will several alterations. 
One of the problems with selecting themes are inconsistencies in the order of the `PaletteMode` `Enum` and the `PaletteModeStrings.SupprotedThemes` Dictionary.
This order will be corrected and the SupportedThemes order will be leading.

I would like to ask to make no changes to the two until this issue is complete.

This is the order of the two currently with the discrepancies between the two.

![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/96108132/414e1443-6417-4f62-accf-01b485d447ec)

Smurf-IV:
I prefer the Right list, as it shows the "New" themes being added to the end of the list.
Alphabetical sorting can be done in the combo, as it will still (should) take the tag numbers for the offsets.

Smurf-IV:
@giduac Also, remember that this has to be placed into 8.5, and still work in V90 without any changes to a designer made with V80 (i.e. enum offsets etc.)

Smurf-IV:
A Bi Dectional map should take care of identifying what is what (They have been used for other "Sources of truth" as well)

giduac:
[...]

PR #1500, opened 2024-05-30
1388-drop-button-icon-is-not-drawn

giduac:
1388-drop-button-icon-is-not-drawn
[issue 1388](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1388)

Kbutton has been changed so the arrow color reacts to theme changes.
Also KDropButton was affected by this, but only needed the change in ButtonValues.cs.

On the way the code block in `KButton.cs` method `OnPaint` of multiple if statement testing for a `DialogResult` value has been transformed into an if else if block.

https://github.com/Krypton-Suite/Standard-Toolkit/assets/96108132/25b7be13-9e6c-46aa-91e0-2f039df26677

compile-results

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
@@ -689,34 +689,17 @@ namespace Krypton.Toolkit
         {
             if (Values.UseAsADialogButton)
             {
-                if (DialogResult == DialogResult.Abort)
+                Text = DialogResult switch
                 {
-                    Text = KryptonManager.Strings.GeneralStrings.Abort;
-                }
-                else if (DialogResult == DialogResult.Cancel)
-                {
-                    Text = KryptonManager.Strings.GeneralStrings.Cancel;
-                }
-                else if (DialogResult == DialogResult.OK)
-                {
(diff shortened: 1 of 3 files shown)


Smurf-IV:
@giduac FYI:
image

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
```diff
@@ -693,33 +693,27 @@ protected override void OnPaint(PaintEventArgs? e)
                 {
                     Text = KryptonManager.Strings.GeneralStrings.Abort;
                 }
-
-                if (DialogResult == DialogResult.Cancel)
+                else if (DialogResult == DialogResult.Cancel)
```
Smurf-IV, line 696: This is horrible=..
Please make it a `switch`

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonButton.cs
giduac: I actually tried those:
- A normal switch is 3 lines longer?
- A switch expression lacks a default value, which generates a warning?

Which one would you like?

Issue #1328, opened 2024-02-11
[Bug]: KryptonThemeComboBox: invalid designer code generation [alpha-1309]


tobitege:
**KryptonThemeComboBox generates wrong code in designer file**

Saving a form with such component writes a line like below to the designer.cs file:  

`this.kryptonThemeComboBox1.ThemeSelectedIndex = 24;`  

Compiler errors then as ThemeSelectedIndex is not writable.  
Removing that line manually fixes the error.

@Wagnerp I'm currently working on the alpha-derived branch named 1309-xxx, but I'd guess it is happening in alpha as well, if not also in the Canary branch.

**Expected behavior**
The above property should either not be written to the designer file or be made accessible.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Version: 23H2
 - Framework/.NET Version: 4.8.x
 - Toolkit Version: current alpha

Wagnerp:
Hi @tobitege 

It's set at the wrong index, the default should be `Microsoft 365 - Blue` at 3x.

Wagnerp:
This  _should_ stop it

```[DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]```

tobitege:
When manually in form design mode switching the DefaultPalette to a different theme like e.g. Office2010Black it correctly writes this line to the designer:  
this.kryptonThemeComboBox1.DefaultPalette = Krypton.Toolkit.PaletteMode.Office2010Black;  
(that's where the mentioned ThemeSelectedIndex line came from).

However, there is still the issue, that changing the property in the Properties at design time, it still displays the wrong text inside the combo:  
![grafik](https://github.com/Krypton-Suite/Standard-Toolkit/assets/10787084/d07dba7c-7662-44bd-88c8-35a4a8054d5b)

Form designer:
![grafik](https://github.com/Krypton-Suite/Standard-Toolkit/assets/10787084/70537838-ea8a-4bfe-95c9-8ce16f69a060)

Wagnerp:
It's not pairing up for some reason? Its always off by 1 or 2

tobitege:
> It's not pairing up for some reason? Its always off by 1 or 2

Yes, it is off by 1 in this example. I thought that was already fixed last week? 🤔

tobitege:
[...]

PR #1505, opened 2024-06-02
1328-ThemeCBB-invalid-designer-code

giduac:
1328-ThemeCBB-invalid-designer-code
[issue 1328](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1328)
Adjusts PaletteMode enum and SupportedThemes list.

This corrects this issue.

Further actions towards the Theme selector controls will be done in a separate issue since those will take more time to implement.

> I left the disabled entries in their place because the dictionary has them listed in the same order. If one is re-enabled again then it's in the right spot at once.
> 
> Keeping PaletteMode and the Dictionary in the same order will make it easier to maintain consistency.
> 
> In the future PaletteMode will not be used as an index anymore. Like you pointed out the needed values can be retrieved from the BiDirectional map in either way.
> 
> PaletteMode will then only serve as a unique identifier. That can be different per theme throughout builds.
> 
> That all will be implemented in the follow-up on this where all three selectors will be revised.
> 
> I kept this brief so the reporter can move on.

Will be continued in: https://github.com/Krypton-Suite/Standard-Toolkit/issues/1507

compile-results


Wagnerp (approved):

Source/Krypton Components/Krypton.Toolkit/Palette Base/PaletteMode.cs
```diff
@@ -89,17 +90,19 @@ public enum PaletteMode
         /// </summary>
         Office2007BlackDarkMode,
 
-        // ToDo: Re-enable when the gray themes are completed
-        /*
-         /// <summary>
-        /// Specifies the dark Gray color variant of the Office 2010 appearance.
+        /* ToDo: Re-enable when the gray themes are completed
```
Smurf-IV, line 93: These commented out elements need to be moved to the end,
Because if my designer has selected 24, and then later on I have a nuget package update, then 24 will be a different number !!!
Or, The enums have explicit numbers, that are stored in the ComboBox Item tag, which are related in a Map, and they never change, regardles of commented out, OR filtered out mebers

giduac: I left the disabled entries in their place because the dictionary has them listed in the same order. If one is re-enabled again then it's in the right spot at once. 

Keeping PaletteMode and the Dictionary in the same order it will make it easier to maintain consistency.

In the future PaletteMode will not be used as an index anymore. Like you pointed out the needed values can be retrieved from the BiDirectional map in either way.

PaletteMode will then only serve as a unique identifier.  That can be different per theme throughout builds.

That all will be implemented in the follow-up on this where all three selectors will be revised.

I kept this brief so the reporter can move on.

Issue #1556, opened 2024-06-22
[Bug]: Launching TestForms -> Controls Test results in Exception message box


Smurf-IV:
Using today's alpha:

![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/861a5dc1-4c89-4517-9f31-aef65e29a5ce)

@giduac
Appears to be theme selection related

Smurf-IV:
And when launching Calendar:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/131b7b32-905d-40ab-8bde-ec1c0fd67412)

giduac:
Hi @Wagnerp,

Please assign to me, ty.

Wagnerp:
> Hi @Wagnerp,
> 
> Please assign to me, ty.

Hi @giduac 

Done

giduac:
@Smurf-IV & @Wagnerp 

These messages do not happen consistently. 
I had it one time yesterday. Then tried another 10 times or so with problems.

Seems that somewhere a setting gets changed that triggers a debugtools warning/message...
That gets triggered then when a theme changed  is performed....

Next to that: 
- Outlookgrid.cs has problems
- Fade form hangs the program when the form is closed

giduac:
Hi @Wagnerp & @Smurf-IV,

A question about form fading.

Which was the reason to add the `ModernFadeFormIn` and `ModernFadeFormOut` methods to `KryptonFormFadeController`?

Wagnerp:
> Hi @Wagnerp & @Smurf-IV,
> 
> A question about form fading.
> 
> Which was the reason to add the `ModernFadeFormIn` and `ModernFadeFormOut` methods to `KryptonFormFadeController`?

Hi @giduac 

They were designed specifically to work with .NET. Problem with fading though is there is a bug where the KForm 'pops' in at the very last second.

giduac:
@Wagnerp 
Ok, thanks.

The modern fadeout method doesn't work since the loop is infinite. Making all programs using this and .NET hang on fade out.
The < .NET and this version  also has some flickering when fading.

I'll make the FadeForm with some test controls to show a more fine grained fade in and out which is more fluent. But that will always be a bit speculative since not all computers are the same and also the load on a system can influence that.

With the findings of that FadeForm we can then adjust those methods and maybe the .NET Framework ones also.

giduac:
@Smurf-IV & @Wagnerp 

[...]

PR #1575, opened 2024-06-26
1556-Launching-TestForms-results-in-exception-messagebox

giduac:
[1556-Launching-TestForms-results-in-exception-messagebox](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1556)

Tentative FadeForm update to test fade settings.
[See Comment and Explanation](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1556#issuecomment-2192297916)

compile-results

### DIFF
--- a/Source/Krypton Components/TestForm/FadeFormTest.cs
+++ b/Source/Krypton Components/TestForm/FadeFormTest.cs
@@ -7,6 +7,8 @@
  */
 #endregion
 
+using System.Threading;
+
 namespace TestForm
 {
     public partial class FadeFormTest : KryptonForm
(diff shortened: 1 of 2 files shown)


Wagnerp (approved):

Source/Krypton Components/TestForm/FadeFormTest.cs
```diff
@@ -7,13 +7,71 @@
...
 {
     public partial class FadeFormTest : KryptonForm
     {
         public FadeFormTest()
         {
             InitializeComponent();
+            cbtnShowImage.Checked = true;
+        }
+
+        private void FadeFormTest_FormClosing(object sender, FormClosingEventArgs e)
```
Smurf-IV, line 22: So this is a workaround just for the test - app ?
Doesn;t solve the usage of the FadeForm at it's core at all !

Smurf-IV (changes requested):

Source/Krypton Components/TestForm/FadeFormTest.cs
giduac: > So this is a workaround just for the test - app ? Doesn;t solve the usage of the FadeForm at it's core at all !
Doesn't resolve it yet. 
This way you both can test this and see what works / give good results.
From there on the controller routines can be adjusted as a final solution.

Issue #1613, opened 2024-07-08
[Bug]: KryptonMessageBox text is not centered vertically 


Ahmed-Abdelhameed:
Because KMB now uses a KTextBox to display the text, the text doesn't get centered vertically when the height of the icon is larger, which doesn't look good if compared with the original MB or older versions of KMB:

![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/18140178/5170d9c5-25dc-498c-84b9-af4c9f5c5ee5) vs ![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/18140178/10d7e30b-f933-4eb2-8e0c-8b3a508ab944)

A workaround would be to calculate the size difference and use it to offset the text box but there's a simpler solution, IMO. That is, to just increase the top and bottom margin of the text area because I think it needs more margin anyway, especially after fixing #1599.

We just need to change one line in `KryptonMessageBoxForm.Designer.cs`:
```
this._panelContentArea.Margin = new System.Windows.Forms.Padding(4);
```
to
```
this._panelContentArea.Margin = new System.Windows.Forms.Padding(4, 12, 4, 12);
```

### Before:

| One line | Multiple lines | No icon |
|--------|--------|--------|
| ![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/18140178/6a59e3da-a446-48f5-8ff7-60c889ad2276) | ![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/18140178/a5c55e6a-fbf3-4336-a616-6f6869c34ef7) | ![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/18140178/4356858a-3954-4385-8308-dd517a14ee20) |

### After:
| One line | Multiple lines | No icon |
|--------|--------|--------|
| ![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/18140178/9d0c7f2c-a01a-44d8-8df7-3588a1bd822c) | ![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/18140178/b39bae31-e808-4837-9020-5c89ca9618f6) | ![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/18140178/cec648b6-01a5-4e06-b086-d9398b24796b) |

giduac:
@Wagnerp 
Please assign to me and @Ahmed-Abdelhameed , ty.

Wagnerp:
[...]

PR #1618, opened 2024-07-09
1613-V85-KryptonMessageBox-text-is-not-centered-vertically

Ahmed-Abdelhameed:
[Issue 1613-KryptonMessageBox-text-is-not-centered-vertically](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1613)

- Increased the top and bottom margin of the panel to fix the referenced issue for single-line texts and also to avoid having the text too close to the borders.

image

### DIFF
--- a/Documents/Help/Changelog.md
+++ b/Documents/Help/Changelog.md
@@ -3,6 +3,7 @@
 =======
 
 ## 2024-07-xx - Build 2407 (Patch 1) - July 2024
+* Resolved [#1613](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1613), `KryptonMessageBox` text is not centered vertically.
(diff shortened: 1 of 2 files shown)


Wagnerp (approved):

Smurf-IV:
@Ahmed-Abdelhameed Conflicts.. Why is the github comparer so silly !

Ahmed-Abdelhameed:
Hi @Smurf-IV 

It's because I created my 3 branches based on the V85 branch. So, they all have one entry in `Changeslog.md` that doesn't exist in the other branches. In any case, I resolved the conflict in #1619. Once that's merged, we can resolve this one too (because if we fix it now, it'll cause another conflict after the merge).

P.S. If there's a better way to do what I did, please let me know. Thank you!

Issue #1614, opened 2024-07-08
[Bug]: KryptonMessageBox sometimes throws an exception after Esc key is pressed


Ahmed-Abdelhameed:
This is not an easy one to reproduce because it happens randomly but when the Esc key is pressed, sometimes KMessageBox throws an exception of type `System.Runtime.InteropServices.ExternalException` and the bug is in [this part of the code](https://github.com/Krypton-Suite/Standard-Toolkit/blob/4c38c8492a2290291047e2f668fa17efe8417931/Source/Krypton%20Components/Krypton.Toolkit/Controls%20Visuals/KryptonMessageBoxForm.cs#L705):

```
private void AnyKeyDown(object sender, KeyEventArgs e)
{
    // Escape key kills the dialog if we allow it to be closed
    if (ControlBox
        && (e.KeyCode == Keys.Escape)
       )
    {
        Close();
    }
    else if (!e.Control
             || (e.KeyCode != Keys.C)
            )
    {
        return;
    }

    // [Irrelevant code omitted]

    Clipboard.SetText(sb.ToString(), TextDataFormat.Text);
    Clipboard.SetText(sb.ToString(), TextDataFormat.UnicodeText);
}
```

The problem is that after `Close()` is called, the method doesn't return, allowing the execution to continue while the form is being disposed of. To fix this, we could either add `return;` after `Close();` or to avoid having two `return;` statements, we can turn the `else if` into its own `if` statement:

```
private void AnyKeyDown(object sender, KeyEventArgs e)
{
    // Escape key kills the dialog if we allow it to be closed
    if (ControlBox
        && (e.KeyCode == Keys.Escape)
       )
    {
        Close();
    }

    if (!e.Control || (e.KeyCode != Keys.C)) return;

    // [Irrelevant code omitted]

    Clipboard.SetText(sb.ToString(), TextDataFormat.Text);
    Clipboard.SetText(sb.ToString(), TextDataFormat.UnicodeText);
}
```

Wagnerp:
> @Wagnerp Please assign to me and @Ahmed-Abdelhameed, ty.

@Ahmed-Abdelhameed & @giduac 

Now assigned to both of you

giduac:
@Ahmed-Abdelhameed 

Btw.: we always use curly braces also on control statements with one line.
I think this is fine. 

```cs
            if (ControlBox && e.KeyCode == Keys.Escape))
            {
                Close();
                return;
            }
            
[...]

PR #1619, opened 2024-07-09
1614-V85-KryptonMessageBox-throws-an-exception-after-Esc-key-is-pressed

Ahmed-Abdelhameed:
[Issue 1614-KryptonMessageBox-throws-an-exception-after-Esc-key-is-pressed](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1614)

- Move the copying logic inside an `else if` branch to avoid executing the code after Close().
- Use `e.KeyData` for more accurate key combinations.

image

### DIFF
--- a/Documents/Help/Changelog.md
+++ b/Documents/Help/Changelog.md
@@ -3,6 +3,7 @@
 =======
 
 ## 2024-07-xx - Build 2407 (Patch 1) - July 2024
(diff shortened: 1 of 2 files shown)


Wagnerp (approved):

Smurf-IV:
@Ahmed-Abdelhameed Conflicts...

Issue #615, opened 2022-01-28
[Feature Request]: Controls should have "min design heights" be the same.


MattH-Work:
It would be beneficial to allow 'input' controls (ComboBoxes, TextBoxes, NumericUpDown, DateTimePicker etc) to have their heights modifiable independently of the Font size.

Textbox height can change if set to multiline, whereas ComboBoxes (my main reason for this post) and other controls are linked to their font size

Default Heights
Textbox                    23
NumericUpDown     22
Combox                   21
DateTimePicker        21

It’s only a pixel here and there ( slight OCD triggering, 1st world problems and all! ), but its noticeable when controls don't line up on forms

I do note that the height increases when using ButtonSpec, or adding Rounding to the borders, so it feels the code for manipulating the heights exists in the background

Wagnerp:
@Smurf-IV Is this possible?

Smurf-IV:
Probably, But I think having them all the same size would be a good start !

MattH-Work:
Recently found that changing Minimum & Maximum size 

![image](https://user-images.githubusercontent.com/84494183/172795767-bebc5b05-718e-4c57-ad3d-e884ae91562d.png)

allows me to set the height of the control, which solves this particular issue for me

Smurf-IV:
@MattH-Work Great find.
@Wagnerp Controls should have "min design heights" be the same.

Smurf-IV:
https://docs.microsoft.com/en-us/dotnet/core/compatibility/winforms#change-description
Make all controls 27% bigger !

Smurf-IV:
> https://docs.microsoft.com/en-us/dotnet/core/compatibility/winforms#change-description Make all controls 27% bigger !

I'll make min heights to be **_30_**, and that will cover both the above (i.e. 23 * 1.27 = 29.###) and when corners are used, and when a button spec is inserted.

Smurf-IV:
The Problem is now that Up Downers and Drop Downs DO NOT FIT, into the larger space:
![image](https://github.com/user-attachments/assets/b546d3c3-4ddc-4acc-95c5-79be823bdf64)

Smurf-IV:
Done and also make the NumUpDowner center when using a smaller font:
![image](https://github.com/user-attachments/assets/a941fbb3-7d75-4c7e-b442-b01af76129fd)

Smurf-IV:
Note: Won;t fix for V85 as there are tooo many changes

[...]

PR #1636, opened 2024-07-14
- Make all controls follow MinHeight of 30 (Apart from Min Trackbar :-)

Smurf-IV:
#615

Something has "Gone wrong"  ??

image

Not sure why these nullables are being reported now ?

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,7 @@
 =======
 
 ## 2024-11-xx - Build 2411 - November 2024
+* Resolved [#615](https://github.com/Krypton-Suite/Standard-Toolkit/issues/615), Controls should have "min design heights" be the same.
 * Resolved [#1628](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1628), Some themes do not render the "ToolStrip" Correctly
 * Implemented [#632](https://github.com/Krypton-Suite/Standard-Toolkit/issues/632), **[Breaking Change]** `KryptonPropertyGrid` should have a customisable back colour.
 * Resolved [#1564](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1564), Disabled Button Text in Ribbons is not visible in some themes
--- a/Source/.editorconfig
+++ b/Source/.editorconfig
@@ -49,5 +49,5 @@ dotnet_diagnostic.CS8073.severity = None
 dotnet_diagnostic.CS8618.severity = None
 
 #Warning	CS8622	Nullability of reference types in type of parameter 'sender' of ####
-dotnet_diagnostic.CS8622.severity = None
+#dotnet_diagnostic.CS8622.severity = None
 
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCheckedListBox.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCheckedListBox.cs
@@ -161,6 +161,7 @@ namespace Krypton.Toolkit
             }
 
             #endregion
+
         }
 
         /// <summary>
@@ -525,6 +526,11 @@ namespace Krypton.Toolkit
             #endregion
 
             #region Protected
+
+            /// <summary>Gets the length and height, in pixels, that is specified as the default minimum size of a control.</summary>
+            /// <returns>A <see cref="T:System.Drawing.Size" /> representing the size of the control.</returns>
+            protected override Size DefaultMinimumSize => GlobalStaticValues.DefaultMinimumSize;
+
             /// <summary>
             /// Creates a new instance of the item collection.
             /// </summary>
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonColorButton.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonColorButton.cs
@@ -123,7 +123,7 @@ namespace Krypton.Toolkit
             AutoRecentColors = true;
             SchemeThemes = ColorScheme.OfficeThemes;
             SchemeStandard = ColorScheme.OfficeStandard;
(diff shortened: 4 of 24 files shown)


Smurf-IV:
> Not sure why these nullables are being reported now ?

This is due to the change in the `editorConfig` I suspect it was masking a lot of warnings !!

Wagnerp (approved):

Ahmed-Abdelhameed:
FYI: This change completely messed up KMessageBox and KInputBox as well as other visual forms due to KBorderEdge's (new) min height.

Changing the min height for that specific control would likely fix this particular issue but I can see other issues with KComboBox, KNumericUpDown and I'm 99% sure that this change will cause many other unexpected side effects and should be reversed.

I read the linked issue and I don't agree with the author. Certain controls in WinForms (like ComboBox, NumericUpDown, etc.) have always had their height managed internally. Many controls (including ComboBox, Label, etc.) have a default height that is different from, say, TextBox. and I don't think we should try to change that for the sake of the alignment argument. The WinForms designer has alignment tools (read: Format menu) to help with that. Personally, I have shortcuts like `Ctrl+Shift+A, Ctrl+Shift+M` and `Ctrl+Shift+A, Ctrl+Shift+C` mapped to 'Align Middles' and 'Align Centers', which allows me to align controls pretty quickly.

TL;DR: No need to try and unify the sizing of all controls to make control alignment (arguably) marginally easier as this will cause many other problems.

giduac:
@Smurf-IV, @Wagnerp, @Ahmed-Abdelhameed & @MattH-Work,

Ahmed's statement about Winform controls NOT having a standard size is true., The image below shows most of them aligned to the same top value.

image

Wagnerp:
Hi @Ahmed-Abdelhameed & @giduac 

I don't think the changes made to `editorconfig` should **not** be reverted, as prior to this it was hiding 7000+ nullable warnings. Plus the toolkit has non-standard WinForms features, i.e. ButtonSpecs which need to be considered.

cc. @Smurf-IV

Ahmed-Abdelhameed:
Hi @Wagnerp 

Agreed. I was mainly talking about the changes to Min Height.

> Plus the toolkit has non-standard WinForms features, i.e. ButtonSpecs which need to be considered.

I realize that. I wasn't suggesting that everything should be exactly the same as in the original WinForms controls. I was just asserting that control alignment shouldn't be the deciding factor when it comes to the sizing of controls (in my opinion, at least).

MattH-Work:
image

From the above image, we can see that the original WinForms default heights are close to consistent ( combo box being the outlier in the 'box' category )

The krypton default heights are a little more of a mixed bag, but even though it's only a pixel here and a pixel there, it can still be a little jarring when they controls are aligned

The lower strip shows my preferred 25 height ( set via Minimum Size control )
image

The drop-down on the Date Picker stretches but the Combo Box does not, should it not be the same?
NumberUpDown & Combo Box maintain the drop-down height with a constant gap
I would like them all to stretch, but would accept them being fixed height, but consistent through all control types
image
( I also note that text does not consistently place itself, some are centres some are positioned at the top of the control )

To my mind, the way Date Pickers behave seems correct, stretch & centre
image


Check Box, Radio Button & Labels seem to follow their own rules. The top two rows ( Winforms & Krypton ) are aligned by tops, the 3rd row ( default height ) are all aligned by middle
We can see that Winforms is just wrong on all 3, whereas Krypton at least align with each other, however, accepting the control itself is shorter ( Check Box's are only 13 pixels high ) vs the box height of 20, 22, 25 etc
Is there a way to make the top of the Check Box's etc align with the top of the Combo Box ''correctly''
( see below, top row current result, bottom row 'expected, but still 'off' as the square & cirlce are aligned, but the text is not !! )
image

So to wrap it up, give us a consistent _default_ height, but still allow it to be altered ( reduced & increased ) via Minimum & Maximum height setting in the designer, which would then give correct 'alignment' results

Whatever happens, we still love you! 
We appreciate all the work done on these Krypton controls and I have noticed the increase in updates recently with the onboarding of new Devs, thankyou

giduac:
@MattH-Work 

My 1.5 cents on this.

- If you go for a fixed initial height existing interfaces need to be reformatted, imo that's undesired.
- Making all controls sizeable would be a great plus.
- If a proof of concept is done minimum / maximum size and corresponding font heights can be explored and tested.

As for the onboarding. We still have plenty of work so if you like to tuck in....you're welcome.... :)

MattH-Work:
> As for the onboarding. We still have plenty of work so if you like to tuck in....you're welcome.... :)

Unfortunately, I am a VB hobbyist, my understanding of C, C#, C++ etc only goes so far as the third letter of the alphabet!
I have 'rolled' my own Krypton based controls ( multi column combobox is my current pride and joy ) but they are far from production ready, even embarrassing compared to the work of you 'professionals" !

image

giduac:
> > As for the onboarding. We still have plenty of work so if you like to tuck in....you're welcome.... :)
> 
> Unfortunately, I am a VB hobbyist, my understanding of C, C#, C++ etc only goes so far as the third letter of the alphabet! I have 'rolled' my own Krypton based controls ( multi column combobox is my current pride and joy ) but they are far from production ready, even embarrassing compared to the work of you 'professionals" !
> 
> image

There's always time to start. C# as a language isn't particularly difficult and since you do VB you already have knowledge of the .NET framework/api.

If you do java(script) or php for example, the control structures and formatting are pretty much the same.

If you don't try then you'll never know... ;)

Wagnerp:
Hi @giduac & @MattH-Work 

> If you do java(script) or php for example, the control structures and formatting are pretty much the same.

I've started to learn how to use TypeScript over the last 24 hours (for work) and it's surprisingly easy to learn. Now I know where C# has inherited the `null` ideas from!

Smurf-IV:
Locking this, as these conversations should be the bug not the merge request ;-)

Wagnerp:
> Locking this, as these conversations should be the bug not the merge request ;-)

I agree

Issue #1373, opened 2024-03-14
[Bug]: Getting an application crash when pressing the Delete key.


Tape-Worm:
**Describe the bug**
In my application, I have a grid containing a list of files. I also have a Krypton ribbon present. When I press my delete key, I expect the application to ask me if I want to delete the file.  

Instead, what happens is an unhandled exception is thrown. This is weird, because my code should have trapped the exception. It did not.

Here is the stack trace that is given on the crash:
```
Error Message: Unable to cast object of type 'System.Collections.Generic.Dictionary`2[System.Windows.Forms.Keys,System.Windows.Forms.ToolStripMenuItem]' to type 'System.Collections.Hashtable'.
Exception Type:  System.InvalidCastException
Source:  System.Private.CoreLib
Target Site:  System.Runtime.CompilerServices.CastHelpers.ChkCast_Helper
Stack trace
   at Gorgon.Editor.Program.Main(String[] args) in Program.cs:line 197
   at Gorgon.UI.GorgonApplication.Run(ApplicationContext context, Func`1 idleMethod) in GorgonApplication.cs:line 782
   at System.Windows.Forms.Application.ThreadContext.RunMessageLoop(msoloop reason, ApplicationContext context)
   at System.Windows.Forms.Application.ThreadContext.RunMessageLoopInner(msoloop reason, ApplicationContext context)
   at System.Windows.Forms.Application.ComponentManager.Microsoft.Office.IMsoComponentManager.FPushMessageLoop(UIntPtr dwComponentID, msoloop uReason, Void* pvLoopData)
   at System.Windows.Forms.Application.ThreadContext.Microsoft.Office.IMsoComponent.FPreTranslateMessage(MSG* msg)
   at System.Windows.Forms.Application.ThreadContext.PreTranslateMessage(MSG& msg)
   at System.Windows.Forms.Application.ThreadContext.ProcessFilters(MSG& msg, Boolean& modified)
   at Krypton.Ribbon.KryptonRibbon.PreFilterMessage(Message& m)
   at Krypton.Ribbon.KryptonRibbon.ProcessCmdKey(Message& msg, Keys keyData)
   at Krypton.Ribbon.KryptonRibbonTab.ProcessCmdKey(Message& msg, Keys keyData)
   at Krypton.Ribbon.KryptonRibbonGroup.ProcessCmdKey(Message& msg, Keys keyData)
[...]

PR #1639, opened 2024-07-16
1373-Getting-an-application-crash-when-pressing-the-Delete-key

giduac:
[Issue 1373-Getting-an-application-crash-when-pressing-the-Delete-key](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1373)
- `KT.CommonHelper.CheckContextMenuForShortcut()` handles direct type casts differently from .NET 8.0 onward. 
- Solution courtesy of @Tape-Worm
- And the change log

compile-results

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,7 @@
 =======
 
 ## 2024-11-xx - Build 2411 - November 2024
+* Resolved [#1373](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1373), `KT.CommonHelper.CheckContextMenuForShortcut()` handles direct type casts differently from .NET 8.0 onward. Solution courtesy of @Tape-Worm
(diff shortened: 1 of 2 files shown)


Documents/Changelog/Changelog.md
```diff
@@ -3,6 +3,7 @@
 =======
 
 ## 2024-11-xx - Build 2411 - November 2024
+* Resolved [#1373](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1373), `KT.CommonHelper.CheckContextMenuForShortcut()` handles direct type casts differently from .NET 8.0 onward. Solution courtesy of @Tape-Worm
```
Wagnerp, line 6: Please move to V85 patch section to reduce duplication, thanks!

giduac: @Wagnerp 
Where would that be?

Wagnerp: Should be line 136

giduac:
@Wagnerp 
Done

PR #1671, opened 2024-07-22
* 1661 bug office 2007 dark blue theme form text is hard to read when app loses focus

### LINKED ISSUE #1661: [Bug]: Office 2007 Dark Blue theme: Form Text is hard to read when app loses focus
Using todays alpha
image

Same applies to other "dark" themes as well

Also when the Form does have focus the "Extra text" area is still hard to read!

Wagnerp:
* Resolved #1661 - Darker grey
* Some tweaks to `ExceptionHandler`

image


Source/Krypton Components/Krypton.Toolkit/Tooling/ExceptionHandler.cs
giduac, line 1: Shouldn't this better be handled through a separate ticket.?
If the pr would need reverting then those changes that have nothing to do with the ticket it handles are also backed out....

Wagnerp: Reverted

giduac (approved):

Issue #1675, opened 2024-07-24
[Bug]: Catastrophic failure wherever KryptonGroupPanel is used in V85


Ahmed-Abdelhameed:
This is a regression that was apparently caused by [this fix](https://github.com/Krypton-Suite/Standard-Toolkit/commit/e62aa59942b1560274eecfffd30c78eda826bb16).

To reproduce:

- Add a `KryptonGroup` control on any form.
- Run the program and open that form.
- The program crashes due to a NullReferenceException.
- The bug can also be viewed in the designer if the form is reopened after adding the KryptonGroup control.

PR coming right up.

Ahmed-Abdelhameed:
Hi @Wagnerp

I'd appreciate it if you could update the NuGet package (again 😅) after merging the PR because anyone updating to build 407 is likely to be faced with this critical bug.

PR #1676, opened 2024-07-24
Fix NullReferenceException in KryptonGroupPanel

Ahmed-Abdelhameed:
Issue [1675-V85-Catastrophic-failure-wherever-KryptonGroup-is-used](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1675)

Replaced the null-forgiving operator `!` with a null-conditional operator `?.` to prevent throwing a NullReferenceException when the KryptonGroupPanel is about to be shown.

image

### DIFF
--- a/Documents/Help/Changelog.md
+++ b/Documents/Help/Changelog.md
@@ -2,6 +2,11 @@
 
 =======
 
+## 2024-07-xx - Build 2408 (Patch 2) - July 2024
+* Resolved [#1675](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1675), Catastrophic failure wherever `KryptonGroupPanel` is used.
+
+=======
+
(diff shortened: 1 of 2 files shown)


giduac (approved):

Wagnerp (approved):

Wagnerp:
Please make sure that the changelog is updated, so we can track the changes. Thanks:)

Ahmed-Abdelhameed:
@Wagnerp

Oh, I meant to leave a comment about this but forgot. The reason I didn't update the changelog this time is because I'm not sure where put the changes after the last build was shipped. Should I add a new heading and name it "2024-07-**24** - Build **2408** (Patch **2**) - July 2024"? Not sure what the standard is in this case.

Ahmed-Abdelhameed:
@Wagnerp Changelog updated.

giduac (approved):

PR #1702, opened 2024-08-03
* Expanded build system

Wagnerp:
* Expanded build system capabilities


giduac (commented):
My point of view on this.

First of all , we are not in the business of software deployment in this project. 

It might be a handy thing to have at first sight but personally I see a slew of problems coming our way which we don't have time for and are, surely, not in the scope of the Controls Toolkit.

Problems and/or errors installing software through the build system will land on our plate. 

I'd like to stick to the project's "Core-Business"...

Scripts/install-prerequisites.cmd
```diff
@@ -0,0 +1,250 @@
+@echo off
```
Smurf-IV, line 1: Please remove this file and the processes that call it..

Smurf-IV (changes requested):

Scripts/install-prerequisites.cmd
Wagnerp: Removed files

giduac (approved):

Issue #1734, opened 2024-08-18
[Feature Request]: Add `ForceDesignerDPIUnaware` option


Wagnerp:
Add `ForceDesignerDPIUnaware` option to `csproj` files

giduac:
@Wagnerp 

Still having problems on TestForm after this change. 
When opening a KForm in the designer there's an error:

![image](https://github.com/user-attachments/assets/e5f69707-e096-427f-a761-503e6bc9289a)

Wagnerp:
> @Wagnerp
> 
> Still having problems on TestForm after this change. When opening a KForm in the designer there's an error:
> 
[...]

PR #1735, opened 2024-08-18
* Implemented #1734

Wagnerp:
* Implemented #1734 - new for .NET 9 (must be using VS 17.8 or higher)

image

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,7 @@
 =======
 
 ## 2024-11-xx - Build 2411 - November 2024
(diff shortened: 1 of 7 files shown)


giduac:
@Wagnerp,

Building from script works, but cannot build alpha inside VStudio. All kind of errors on TestForm...

image


Does this have todo with this change?

Wagnerp:
> @Wagnerp,
> 
> Building from script works, but cannot build alpha inside VStudio. All kind of errors on TestForm...
> 
> image
> 
> Does this have todo with this change?

@giduac 

Yes it does, fix coming in next PR

PR #1736, opened 2024-08-19
* Add missing dependency

Wagnerp:
* Add missing dependency - no build log needed, as it's for the test app

### DIFF
--- a/Source/Krypton Components/TestForm/TestForm.csproj
+++ b/Source/Krypton Components/TestForm/TestForm.csproj
@@ -45,6 +45,7 @@
             <Version>4.0.0.0</Version>
             <!-- Designers for the `TFMs != net4` are "Pulled in" via Visual studios ".nuget\packages\microsoft.windowsdesktop.app.ref" -->
         </Reference>
+        <Reference Include="System.Windows.Forms" />
     </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


giduac:
By exception I will merge this since it is a small change on testform, so we can move on.

Source/Krypton Components/TestForm/TestForm.csproj
```diff
@@ -45,6 +45,7 @@
             <Version>4.0.0.0</Version>
             <!-- Designers for the `TFMs != net4` are "Pulled in" via Visual studios ".nuget\packages\microsoft.windowsdesktop.app.ref" -->
         </Reference>
+        <Reference Include="System.Windows.Forms" />
```
Smurf-IV, line 48: Is this because the `usingWinforms` was removed from the top part ?

giduac:
And related to this:
https://github.com/Krypton-Suite/Standard-Toolkit/issues/1734#issuecomment-2297994704

Source/Krypton Components/TestForm/TestForm.csproj
Wagnerp: Yes, though more investigation is needed

Issue #1755, opened 2024-08-29
[Bug]: V85 Ribbon, moving the mouse over the dropdown arrow of a gallery causes an exception.


giduac:
**Describe the bug**
The timer in `GalleryButtonController` is dereferenced when a timer is not requested and therefore not instantiated.

**To Reproduce**
N/A

**Expected behavior**
N/A

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: [e.g. Windows 11]
 - Version: Latest V85 build / branch master-v85
 - Framework/.NET 4.8.x / Net 6.0 (tested)
 - Toolkit Version: Latest V85 as of this date

**Additional context**
V90 is not affected.

PR #1756, opened 2024-08-29
1755-V85-Ribbon-moving-the-mouse-over-the-dropdown-arrow-of-a-gallery…

giduac:
[Issue 1755-V85-Ribbon-moving-the-mouse-over-the-dropdown-arrow-of-a-gallery-causes-an-exception](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1755)
- Fixes the issue
- And the change log

compile-results

### DIFF
--- a/Source/Krypton Components/Krypton.Ribbon/Controller/GalleryButtonController.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/Controller/GalleryButtonController.cs
@@ -73,7 +73,7 @@ namespace Krypton.Ribbon
                 _pressed = false;
                 _mouseOver = false;
                 UpdateTargetState(new Point(int.MaxValue, int.MaxValue));
-                _repeatTimer.Stop();
+                _repeatTimer?.Stop();
             }
         }
         #endregion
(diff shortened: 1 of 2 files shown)


Source/Krypton Components/Krypton.Ribbon/Controller/GalleryButtonController.cs
```diff
@@ -116,7 +116,7 @@ public virtual bool MouseDown(Control c, Point pt, MouseButtons button)
                 if (Target.Enabled)
                 {
                     OnClick(new MouseEventArgs(MouseButtons.Left, 1, pt.X, pt.Y, 0));
-                    _repeatTimer.Start();
+                    _repeatTimer?.Start();
```
Smurf-IV, line 119: If this is doing a `start`, then should it not have previously checked to see if timer should be created ? And create if Null ?

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Ribbon/Controller/GalleryButtonController.cs
giduac: Naah, The Constructor allows for the timer to be used or not (boolean param).
So it's either a valid reference or null.

Issue #1715, opened 2024-08-06
[Bug]: Not Implemented Exception thrown for `GetRibbonBackColorStyle` `PaletteOffice2010Base.cs`


DigitalMills:
**Describe the bug**
Not Implemented Exception thrown for GetRibbonBackColorStyle for PaletteOffice2010Base.cs

**To Reproduce**
1. Attempt to render and paint a Ribbon Group component

**Expected behavior**
Ribbon Group renders and paints

**Actual behavior**
Error is thrown, suggesting part of the flow is not yet implemented. Stack trace:

System.ComponentModel.Win32Exception (2): An error occurred trying to start process 'https://github.com/Krypton-Suite/Standard-Toolkit/issues/new/choose' with working directory 'C:\Git\Aireforge\Aireforge Suite\OmniCompare\bin\Debug\net8.0-windows'. The system cannot find the file specified.
  at System.Diagnostics.Process.StartWithCreateProcess(ProcessStartInfo startInfo)
  at System.Diagnostics.Process.Start(ProcessStartInfo startInfo)
  at Krypton.Toolkit.DebugTools.NotImplemented(String outOfRange, String callingFilePath, Int32 lineNumber, String callingMethod)
  at Krypton.Toolkit.PaletteOffice2010Base.GetRibbonBackColorStyle(PaletteRibbonBackStyle style, PaletteState state)
  at Krypton.Toolkit.PaletteRedirect.GetRibbonBackColorStyle(PaletteRibbonBackStyle style, PaletteState state)
  at Krypton.Toolkit.PaletteRibbonBackInheritRedirect.GetRibbonBackColorStyle(PaletteState state)
  at Krypton.Toolkit.PaletteRibbonBack.GetRibbonBackColorStyle(PaletteState state)
  at Krypton.Toolkit.PaletteRibbonBack.GetRibbonBackColorStyle(PaletteState state)
  at Krypton.Ribbon.ViewDrawRibbonGroup.RenderNormalBefore(RenderContext context)
  at Krypton.Ribbon.ViewDrawRibbonGroup.RenderBefore(RenderContext context)
  at Krypton.Toolkit.ViewComposite.Render(RenderContext context)
  at Krypton.Toolkit.ViewComposite.Render(RenderContext context)
  at Krypton.Toolkit.ViewControl.OnPaint(PaintEventArgs e)
[...]

PR #1763, opened 2024-09-07
* Expand `TestApp`

PWagner1:
* Expand `TestApp`

image


giduac (commented):
Shouldn't the changes to the theme files be part of 1715 changes ?

PWagner1:
> Shouldn't the changes to the theme files be part of 1715 changes ?

No, as the changes to the TestApp is for validating the potential fixes for #1715

giduac:
> > Shouldn't the changes to the theme files be part of 1715 changes ?
> 
> No, as the changes to the TestApp is for validating the potential fixes for #1715

Then this PR is best referenced in the 1715 in case some things might need rolling back in the future.

PR #1773, opened 2024-09-15
* 1771 bug net 9 rc 1 throws wfo1000

PWagner1:
* Disabled WFO1000 (for now)

{230E39BE-1C81-48EE-9B9D-F1DF9BF213B6}


Smurf-IV:
Warnings would be best in the `.editorconfig`
Also need to update the docs to show the removal of `.Net6` from V90

PWagner1:
> Warnings would be best in the `.editorconfig` Also need to update the docs to show the removal of `.Net6` from V90

@Smurf-IV & @giduac 

I don't think we should completely disable `WFO1000`, so I've set it to warning only.

PWagner1:
Dealt with in PR #1775

Issue #1757, opened 2024-08-29
[Bug]: KForm has a thin magenta border after the fix of #1749


giduac:
**Describe the bug**
See #1749 for extended information.
Since this has been fixed KForm has a thin magenta border.
Only when the border width is set to 0 the border is not visible.

**To Reproduce**
N/A

**Screenshots**
N/A

**Desktop (please complete the following information):**
- Framework/.NET Version: current alpha

**Additional context**
N/A

PWagner1:
Hi @giduac 

What happens if you set it to transparent?

Smurf-IV:
> Hi @giduac
> 
> What happens if you set it to transparent?

It should already be "Transparent" from the change that I made, so is something resetting it to a different colour ?

Smurf-IV:
> > Hi @giduac
> > What happens if you set it to transparent?
> 
> It should already be "Transparent" from the change that I made, so is something resetting it to a different colour ?

FYI:
The transparency no longer applies, because this is a 0.5 pixel offset, therefore the colour is slightly merged with the border, thus not making it "Pure" magenta" anymore !
Then due to HDpi monitors and scaling, it makes it appear to be a single pixel border added to the Form Non-Calc region(s)
This is mainly caused by the introduction of the `float` rounding, and the fallout of using `float` calculations for borders that are drawn into `int` regions. It was fine until windows started to respect half-pixels, and then scaled them !!

giduac:
> > > Hi @giduac
> > > What happens if you set it to transparent?
> > 
> > 
> > It should already be "Transparent" from the change that I made, so is something resetting it to a different colour ?
> 
> FYI: The transparency no longer applies, because this is a 0.5 pixel offset, therefore the colour is slightly merged with the border, thus not making it "Pure" magenta" anymore ! Then due to HDpi monitors and scaling, it makes it appear to be a single pixel border added to the Form Non-Calc region(s) This is mainly caused by the introduction of the `float` rounding, and the fallout of using `float` calculations for borders that are drawn into `int` regions. It was fine until windows started to respect half-pixels, and then scaled them !!

Then I would suggest to set the default border width for KForm States to zero as a part of this correction.

giduac:
@Smurf-IV 
This might be of interest:
https://reza-aghaei.com/panel-border-size-and-border-color-customize-nonclient-area/

Smurf-IV:
Sorry it took so long!
[...]

PR #1782, opened 2024-09-22
- Remove the Single `0.5f` pixel from the outside Form Borders

Smurf-IV:
- Make sure that the little transparency elements on the curves do not show up for Form Borders

#1757

image

I still have the Old branch, therefore this build report is worthless !
image


giduac:
@Smurf-IV 

Still some strange behaviour when adjusting border settings (that was there before already)

https://github.com/user-attachments/assets/1cf230c0-d7a9-462d-97b8-728192a314bd

PWagner1 (approved):

Smurf-IV:
> @Smurf-IV
> 
> Still some strange behaviour when adjusting border settings (that was there before already)

That does not look like my code change has been applied to that code base, because the Curvature of the corners is "Too" smooth, and the pink pixel still exists around the whol form.

Smurf-IV:
> @Smurf-IV Weird I used that exact branch "1757-KFormThinPinkBorder", tried again from alpha after the merge and that looks like you said.

Yay ;-)

> 
> Still one thing that might need some attention: When borders are set to "None" there's still a partial grey border on the top
> The rest of the form is surround by a thin white border.

Hmm, That might be the case for V85 as well then !

Please raise another ticket for that.

Issue #66, opened 2021-04-01
[Bug]: Cannot Add Ribbon-Buttons-Container (KryptonRibbonGroupTripple) when using .netcore onwards [Returns error due to `abstract class`]


DarthVader85PA:
**Describe the bug**
Fresh Form in VB.Net (.NET 5.0) with NuGet Krypton.Toolkit Standard-Edition latest version.
Add KryptonRibbon-Control from toolbox to Form. I searched for the smart tags on Ribbon-Control to Add Tabs but not available. So I used the RibbonTabs-Property (Collection) and added a Tab, in Tab-Propertys I added a new group and in the group I want add new Items (Collections). After click on "Add" Button I get the following error.

**Screenshots**
![grafik](https://user-images.githubusercontent.com/32876506/113298447-211e9b80-92fc-11eb-8800-0a870cd9c7ac.png)

**Desktop (please complete the following information):**
 - OS: Windows 10 Pro
 - Version: 20H2 - 19042.867
 - Visual Studio 2019 - 16.9.3

sabrinanurhidayah:
Look at YOURFORM.Designer.cs, do manually code on there

```
       private void InitializeComponent()
        {
            this.MainRibbon = new Krypton.Ribbon.KryptonRibbon();
            this.DispenserTab = new Krypton.Ribbon.KryptonRibbonTab();
            this.AccountGroup = new Krypton.Ribbon.KryptonRibbonGroup();
            this.AccountGroupTriple = new Krypton.Ribbon.KryptonRibbonGroupTriple();
            this.AccountGroupButton = new Krypton.Ribbon.KryptonRibbonGroupButton();

            // 
            // MainRibbon
            // 
            this.MainRibbon.HideRibbonSize = new System.Drawing.Size(0, 0);
            this.MainRibbon.InDesignHelperMode = false;
            this.MainRibbon.Name = "MainRibbon";
            this.MainRibbon.QATLocation = Krypton.Ribbon.QATLocation.Hidden;
            this.MainRibbon.RibbonAppButton.AppButtonImage = null;
            this.MainRibbon.RibbonAppButton.AppButtonMenuItems.AddRange(new Krypton.Toolkit.KryptonContextMenuItemBase[] 
            {
            this.ExitContextMenuItem});
            this.MainRibbon.RibbonAppButton.AppButtonShowRecentDocs = false;
[...]

PR #1793, opened 2024-09-29
- Use the winform base type for the designer editor

Smurf-IV:
#66

image


Smurf-IV:
> Note: Still got to find what makes the ribbon **0** height in the designer

Or should I make that s separate bug for V85 ?

PWagner1 (approved):

PR #1814, opened 2024-10-12
Correct Testform csproj.

giduac:
No ticket since this is only for TestForm
Removes the hintpath reference.


giduac:
Merging this so work can move on.

PR #1855, opened 2024-11-16
* Add .NET 9 to `all` TFM

PWagner1:
* Add .NET 9 to `all` TFM

image

No changelog needed


giduac (approved):

giduac:
@PWagner1 
Seems that alpha needs to also...

Issue #1871, opened 2024-11-21
[Bug]: HeaderGroups are 'clipped' after upgrade to 90.24.11.317


MattH-Work:
After upgrading from 85...... to 90.24.11.317, the right-hand edge / border of all Headers & HeaderGroups have been 'clipped'

HeaderGroups

![image](https://github.com/user-attachments/assets/31a3a555-0625-45b2-a863-f26c09037a4d)

![image](https://github.com/user-attachments/assets/d85966d2-ed07-455b-b4c6-d53d4b7b0dc8)

![image](https://github.com/user-attachments/assets/ef3494f7-d5ed-4100-8a53-23fe6ec08593)

Headers

![image](https://github.com/user-attachments/assets/02932aea-1194-45c6-8ca5-58f59bb11728)

MattH-Work:
Navigator pages also suffer
![image](https://github.com/user-attachments/assets/b75d076c-a5de-4349-9fd2-5a3ece6d54df)

Smurf-IV:
This work is being sorted via #1807
I'll keep this open just in case I cannot sort it out with the above tho!

MattH-Work:
I use a KryptonCustomPalette to edit the global pallete, when making the same adjustments directly to the control it works as expected ( no clipping, consistent border width all round ) Issue might stem from the KryptonCustomPalette

MattH-Work:
Stupid question, where do I find UseChrome?

Smurf-IV:
> Stupid question, where do I find UseChrome?

Sorry, I shortened it !

![image](https://github.com/user-attachments/assets/efb4eacb-fb65-46c4-bdc4-6ba215b0eb5c)

MattH-Work:
No dice, and I now have the Magenta border issue #1757   #1749

Smurf-IV:
So far it appears to be something to do with the addition of the "CheckButton" style, even when it is in a stack:
![image](https://github.com/user-attachments/assets/5fd55591-3fc2-4fee-8907-78d424b03bde)

And Roll over (Hover) is even worse !

Smurf-IV:
@MattH-Work AS a possible workaround, can you try the following settings ?
![image](https://github.com/user-attachments/assets/26a77e91-e1d3-484e-9c01-ef8c6f20e878)

MattH-Work:
More evidence of clipping weirdness ( on a button )

[...]

Issue #1783, opened 2024-09-23
[Bug]: KForm borders incorrect


giduac:
[**Describe the bug**
see: [https://github.com/Krypton-Suite/Standard-Toolkit/pull/1782#issuecomment-2367371932](https://github.com/Krypton-Suite/Standard-Toolkit/pull/1782#issuecomment-2367371932)

Smurf-IV:
It appears that the 2 issues that are mentioned:
- Grey around title bar
- White border
Have been there for a very long time:
![image](https://github.com/user-attachments/assets/a02d9853-6044-447e-ac0d-e5973b1939b3)

Smurf-IV:
And it's even worse in the "Old" version, when `Draw: False` is used:
![image](https://github.com/user-attachments/assets/569bc21a-9a03-47df-85c3-f80cf41e5005)

Whereas the current alpha does have that part "Sorted":
![image](https://github.com/user-attachments/assets/64906147-8f60-4f06-9f17-597c9cfe6a8c)

Smurf-IV:
Workaround for the whiteline issue:
- Set the Width to be **-2**

![image](https://github.com/user-attachments/assets/adbde68e-0201-41b0-98d5-ec6458742c85)

Smurf-IV:
Workaround for the Left bit of grey:
- Rounding must be -1
- Width must be 0
- DrawBorder must be Bottom, left, Right

![image](https://github.com/user-attachments/assets/4fd64263-b015-41f0-a862-73318b6950b2)

Smurf-IV:
FYI: The "Grey Lines appear for the Office2010 derived themes.
I'm Checking the others
![image](https://github.com/user-attachments/assets/8f0d4150-5d5a-4a78-a558-35dc19af9701)

Smurf-IV:
V95 PR incoming

PR #1879, opened 2024-11-29
- Fix Borders: all over the places...

Smurf-IV:
- Attempt to make `GetWindowBorders` the single source of truth.
- Make use of `GlobalStaticValues.TRANSPARENCY_KEY_COLOR`
- Pass `KryptonForm? owningForm` into appropriate functions to determine actual borders in use
- Make sure `MdiParent` is taken care of in `KryptonForm` reuse
- Fix some of the overflow when a form is maximised
- Make sure changes to Layout still allow Tooltips to work.
- Fix Usages of borders when they are Not to be drawn
- Update themes to return the usage of the actual borders rather the hardcoded values
- Fix Designer "Misuse" from the `DropDownArrowColor`

#1871
#1783

No warnings ??

image

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,8 @@
 =======
 
 ## 2025-02-01 - Build 2502 (Version 90 - Patch 1) - February 2025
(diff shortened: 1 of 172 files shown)


Smurf-IV:
Never mind, After the merge the `WFO1000` warnings are there ;-(

image

PWagner1 (approved):

Issue #1840, opened 2024-11-09
[Other Issues]: Remove DN9 WFO1000 Warnings


giduac:
### First 
~~Edit: `Krypton-Suite/Standard-Toolkit/Source/.editorconfig`~~ Enabled in PR #1854 & #1856
~~Set: `dotnet_diagnostic.WFO1000.severity` from: `suggestion` to `warning`~~

From the build menu run: Clean and rebuild that will bring forward the WFO1000 warnings

Practically this can be done from the VS 2022 Preview with DN9-RC2

### **The Visibility can differ between the type and use of the properties brought forward through the warnings. So each property has to be evaluated on its own.**

**Use the explanation below to apply the correct attribute**

The `DesignerSerializationVisibility` enum in .NET is used to control how a property or event is serialized by the designer (like the Visual Studio designer) into the source code. It has three options:

### 1. `Visible` (`DesignerSerializationVisibility.Visible`)
   - The property or event is serialized by the designer.
   - This is the default setting and is used for properties that should appear in the designer’s generated code. 
   - When set to `Visible`, the designer serializes the property's value, enabling it to be saved in the form's code and restored at runtime.

### 2. `Hidden` (`DesignerSerializationVisibility.Hidden`)
   - The property or event is not serialized by the designer.
   - Properties set to `Hidden` are ignored by the designer, meaning their values will not be saved to the form’s generated code, even if you set them in the designer.
   - This setting is used for properties that are calculated or transient, or for those that don’t need to be persisted.

### 3. `Content` (`DesignerSerializationVisibility.Content`)
   - The designer serializes the contents of the property, rather than the property itself.
   - This is typically used for complex properties like collections or custom types with nested properties. When set to `Content`, the designer serializes each item or subproperty within the property instead of the entire property.
   - This is useful for properties that contain a collection of items (e.g., a list of buttons within a panel), allowing each item to be individually serialized.

**Example Usage:**

```csharp
[DesignerSerializationVisibility(DesignerSerializationVisibility.Content)]
public List<string> MyListProperty { get; set; } = new List<string>();
```

Part of: <https://github.com/Krypton-Suite/Standard-Toolkit/issues/1777>

giduac:
Alpha is in the works

giduac:
Alpha first and then we do a couple of nugets there before proceeding to V95

giduac:
For V95 this is the log with all WFO10100 warnings that need attention.

[...]

PR #1894, opened 2024-12-04
1840-V95-DN9-changes-to-nullability-annotations

giduac:
[Issue 1840-remove-dn9-warnings](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1840)
- Changes to nullability annotations
- https://learn.microsoft.com/en-us/dotnet/core/compatibility/windows-forms/9.0/nullability-changes
- No change log provided.

compile-results


Smurf-IV (changes requested):
These are not `nullability` change fixes.
Code style for Krypton, is to have vars on the left, and solid types on the right.

giduac:
> These are not `nullability` change fixes. Code style for Krypton, is to have vars on the left, and solid types on the right.

Which results in a nullable type...?

PR #1960, opened 2024-12-17
Fixes-nullables and warnings and build errors.

giduac:
Fixes-nullables and warnings and build errors.
No ticket associated.

@Smurf-IV, @Ahmed-Abdelhameed  & @PWagner1 

The warnings have been removed and also the build problems.

It was the `HintPath` again. Anybody knows where this comes from and how to get rid of it permanently?

```xml
    <Reference Include="Krypton.Toolkit">
      <HintPath>obj\Debug\net462\Krypton.Toolkit.dll</HintPath>
    </Reference>
```

This fix should be part of the canary build that was merged lately.

compile-results


giduac:
> > It was the HintPath again. Anybody knows where this comes from and how to get rid of it permanently?
> 
> I think that is added when VS (Or ReSharper) add the reference, rather than the developer adding it explicitly

@Smurf-IV 
It's becoming a recurring feature...

Issue #2125, opened 2025-02-21
[Feature Request]: Adding NotificationIcon Size in ToastNotification


mohatchaks:
Hi any chance to have a way to change Notificationicon Size
 it a bit big 

![Image](https://github.com/user-attachments/assets/0e5c91c4-30a2-47ff-bee0-f283474ed176)

something like : 
(
NotificationIconWidth = 16
 NotificationIconHeight = 16
)
will be great

tobitege:
@mohatchaks - this should be fixed in alpha branch now, if you want to give it a try.

PWagner1:
@mohatchaks 

Can this be closed?

cc. @tobitege

mohatchaks:
hi which alpha version !! i cant find a way to change the icon size on 100.25.9.244-alpha
using KryptonBasicToastNotificationData also it act like normal form or it should be movable !! also size is bigger now  

<img width="1920" height="1050" alt="Image" src="https://github.com/user-attachments/assets/1fce7a2b-1d68-4b08-a524-9d0052b66932" />

PR #2126, opened 2025-02-22
* Implemented #2125

PWagner1:
* Implemented #2125 - V95

image


Smurf-IV:
Please place the "UpdateIcon" into a base class, that is lot of repeated code that should not be there.

PWagner1:
> Please place the "UpdateIcon" into a base class, that is lot of repeated code that should not be there.

@Smurf-IV Will do for V100, as some major tweaking will need to be undertaken

Issue #2134, opened 2025-03-01
[Feature Request]: Implement a way of getting a way to show keyboard shortcuts


PWagner1:

PWagner1:
@Smurf-IV & @giduac 

This will make sense soon :)

Smurf-IV:
??

![Image](https://github.com/user-attachments/assets/9dc843ce-2be0-4e21-827a-02461189a9a7)

PWagner1:
> ??
> 
> ![Image](https://github.com/user-attachments/assets/9dc843ce-2be0-4e21-827a-02461189a9a7)

Yes, so the `*.ktheme` has a value bundled in it which deals with this.

Smurf-IV:
> Yes, so the `*.ktheme` has a value bundled in it which deals with this.

I think it's down to windows if they are displayed by default or not, or just when the user presses the `Alt` key.

It's not an app thing !

giduac:
@PWagner1 
Can you please have a look at this.
@kamaniar says that since this change shortcu keys aint working anymore.
https://github.com/Krypton-Suite/Standard-Toolkit/issues/2180#issuecomment-2809296510

PWagner1:
> [@PWagner1](https://github.com/PWagner1) Can you please have a look at this. [@KamaniAR](https://github.com/KamaniAR) says that since this change shortcu keys aint working anymore. [#2180 (comment)](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2180#issuecomment-2809296510)

@giduac 

This should not affect V95

giduac:
> > [@PWagner1](https://github.com/PWagner1) Can you please have a look at this. [@KamaniAR](https://github.com/KamaniAR) says that since this change shortcu keys aint working anymore. [#2180 (comment)](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2180#issuecomment-2809296510)
> 
> [@giduac](https://github.com/giduac)
> 
> This should not affect V95

Yes indeed and solved in:
https://github.com/Krypton-Suite/Standard-Toolkit/pull/2210

giduac:
@PWagner1 

Is this one still on?

PWagner1:
> [@PWagner1](https://github.com/PWagner1)
> 
> Is this one still on?

No, not needed

PR #2143, opened 2025-03-09
* 2134 feature request implement a way of getting a way to show keyboard shortcuts

PWagner1:
* Implemented #2134 + some theming stuff

image


giduac:
@PWagner1 

Can you please provide an overview of how this works on the discord channel where we discuss the theming changes.
Thus we keep an overview what is being implemented.

PWagner1:
> @PWagner1
> 
> Can you please provide an overview of how this works on the discord channel where we discuss the theming changes. Thus we keep an overview what is being implemented.

@giduac 

Just pinged you on Discord with a sample

Source/Krypton Components/Krypton.Toolkit/Utilities/KryptonThemeLoader.cs
```diff
@@ -85,32 +85,41 @@ private static KryptonTheme ParseTheme(XmlDocument doc)
...
+                    {
+                        theme.Images[key!] = Image.FromStream(ms);
+                    }
                 }
             }
 
             // Extract Fonts
-            XmlNodeList fontNodes = root.SelectNodes("Resources/Fonts/Font");
-            foreach (XmlNode fontNode in fontNodes)
+            XmlNodeList? fontNodes = root.SelectNodes("Resources/Fonts/Font");
```
Smurf-IV: Why is this a Base64Coded string. Font has a toString and from string conversion

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Utilities/KryptonThemeLoader.cs
PWagner1: @Smurf-IV & @giduac 

Should **only** the images be encoded as base64?

giduac: @Smurf-IV & @PWagner1 & @Ahmed-Abdelhameed 

With all respect for the work already put in.

Also like this PR and especially these changes should be done through dedicated PRs and not mixed in with other work, so we can keep track of what has happened and be able to do a roll-back if applicable.

Earlier announced tests to see if we can fully export the built-in themes (including images) still do not have been reported on e.g. there is no status on that and no knowledge to take forward in the decision making process. All changes need to be compatible throughout the components affected avoiding patchwork. 

I stay at the point to first complete the discussion, possible testing, and planning before coding starts.

giduac: Also I'm not sure if this is the correct approach.

The earlier understanding or idea was to export the built-in themes to the KCusotmpalettebase format which already can serialize to and from XML. 

Probably this serialization in KCPB needs extending and is still up for discussion.

PWagner1: @Smurf-IV & @giduac 

The `KryptonThemeLoader` will export the built-in themes to `*.ktheme`s in its current state. **But** I can't find a way to bundle `Krypton.Base.Themes.dll` into the NuGet package. If either of you have the time, please could you investigate a way of doing this, while I clean the PR's up?

Note: I'll set this as a draft

PWagner1: > Also I'm not sure if this is the correct approach.
> 
> The earlier understanding or idea was to export the built-in themes to the KCusotmpalettebase format which already can serialize to and from XML.
> 
> Probably this serialization in KCPB needs extending and is still up for discussion.

So, the new `*.ktheme` files are essentially the existing XML files, but with extra metadata to accommodate theme name, images & fonts

giduac: This is what I'm saying earlier that the discussion, etc. is not complete and this is floating in an undetermined direction.

Smurf-IV:
Conflicts
And, Please redo this PR with just the changes for the KB shortcuts

Issue #119, opened 2021-05-30
[Feature Request]: `KryptonCheckedListBox` with `DataSource`, `DisplayMember` and `ValueMember` property


DarthVader85PA:
Is it possible to implement in `KryptonCheckedListBox` a DataSource-Property like `Combobox` and `Listbox`?
And also `DisplayMember` and `ValueMember`?

Greetings from Germany, Augsburg,
Danny

PWagner1:
Hi @DarthVader85PA 

Apologies that this has taken 4 years to implement, but please test using `nightly` build > .100

PR #2175, opened 2025-04-12
* Implement #119

PWagner1:
* Implement #119

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCheckedListBox.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCheckedListBox.cs
@@ -951,6 +951,8 @@ namespace Krypton.Toolkit
 
         #region Instance Fields
 
+        private readonly CheckedListBox _checkedListBox;
+
         private readonly PaletteTripleOverride _overrideNormal;
         private readonly PaletteTripleOverride _overrideTracking;
         private readonly PaletteTripleOverride _overridePressed;
@@ -974,6 +976,10 @@ namespace Krypton.Toolkit
         private bool _alwaysActive;
         private bool _forcedLayout;
         private bool _trackingMouseEnter;
+        private object? _dataSource;
+        private string _displayMember;
+        private string _valueMember;
+
         #endregion
 
         #region Events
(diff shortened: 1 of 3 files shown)


giduac:
@PWagner1 

I gave this a quick test.
When a datasource is connected, be it in the designer properties window or from code no items show up in the list at runtime.

Display- an Valuemember have been set.
The datasource works when connected tot a dataset.

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCheckedListBox.cs
```diff
@@ -2404,6 +2518,53 @@ private void OnListBoxMouseChange(object? sender, EventArgs e)
...
+        {
+            _checkedListBox.BeginUpdate();
+
+            _checkedListBox.Items.Clear();
+
+            if (_dataSource != null)
+            {
+                if (BindingContext != null && BindingContext[_dataSource] is CurrencyManager cm)
+                {
+                    for (int i = 0; i < cm!.Count; i++)
```
Smurf-IV, line 2532: Why is this null thing here, it should not be needed as it will not pass in because of the `is` clause above

Smurf-IV: And should not the `i` be related to the `List size` and not what count of the base is set to ?

Smurf-IV (changes requested):

Smurf-IV:
Please also update the example suite code for this and add the comment about the Datasource / Designer above

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCheckedListBox.cs
PWagner1: The code has been changed since

giduac:
@PWagner1 

When the datasource is connected it is possible to add normal items to the list manually (see image).
Not sure if this should be left this way?

When the datasource is disconnected will the items being cleared from the list?

Smurf-IV:
Conflicts

PWagner1:
> Conflicts

Now resolved

PR #2177, opened 2025-04-12
* Resolved #868 - V100

### LINKED ISSUE #868: [Bug]: `KryptonForm` Does not route `WM_HELP` request
Whilst investigating #867 and performing the actions of https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.messagebox.show?view=windowsdesktop-7.0#system-windows-forms-messagebox-show(system-string-system-string-system-windows-forms-messageboxbuttons-system-windows-forms-messageboxicon-system-windows-forms-messageboxdefaultbutton-system-windows-forms-messageboxoptions-system-boolean)

The callback function for `this.HelpRequested += new System.Windows.Forms.HelpEventHandler (this.Form1_HelpRequested);`
Is **not** routed in a KryptonForm

PWagner1:
* Resolved #868 - V100

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/General/Definitions.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/General/Definitions.cs
@@ -3074,7 +3074,7 @@ namespace Krypton.Toolkit
 
     #endregion
 
-    #region KryptonUseRTLLayout
+    #region Enum KryptonUseRTLLayout
 
     /// <summary>Use RTL to display the controls and UI.</summary>
     public enum KryptonUseRTLLayout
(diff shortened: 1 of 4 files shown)


Source/Krypton Components/Krypton.Toolkit/General/Definitions.cs
```diff
@@ -3087,4 +3087,29 @@ public enum KryptonUseRTLLayout
 
     #endregion
 
+    #region Struct HELPINFO
```
Smurf-IV, line 3090: These should go into the same place as the other `PInvoke structs`

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/General/Definitions.cs
PWagner1: Fixed locations

PR #2198, opened 2025-04-27
 - Fix The border locations when a Form is maximised

### LINKED ISSUE #2166: [Bug]: Form with Krypton Ribbon, when maximized, cuts off the right, left and bottom edges.
**Describe the bug**
Form with Krypton Ribbon, when maximized, cuts off the right, left and bottom edges.

**To Reproduce**
Add a Krypton Ribbon to the form.
When the window is maximized, the right, left and bottom edges will have some pixels cut off.
The same problem can be seen in the Krypton Toolkit Demos, with ribbon.

**Expected behavior**
With the window maximized, the right, left and bottom borders should appear the same as in the normal window.

**Screenshots**

Image

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Version: 19045.5608
 - Framework/.NET Version: 4.8
 - Toolkit Version: 95.24.12.344 and above

**Additional context**
Workaround, change the Padding property in the SizeChanged event. Example:

```
Select Case WindowState
    Case FormWindowState.Maximized
        Padding = New Padding(7, 0, 1, 10)
    Case FormWindowState.Normal
        Padding = New Padding(0, 0, 0, 0)
End Select
```

Smurf-IV:
- Also deal with maximised MDI forms

fixes #2166

image


giduac (approved):

PWagner1 (approved):

giduac:
@Smurf-IV 
Will give this a test as well now

Issue #2251, opened 2025-06-21
[Feature Request]: Create an ActiveFormTracker


giduac:
**Is your feature request related to a problem? Please describe.**
Yes: #2200 

**Describe the solution you'd like**
- Tracks within the application if one of it's forms is active.
- Updates KryptonForm to use the tracker automatically

PR #2252, opened 2025-06-21
2251-V95-implements-an-active-form-tracker

giduac:
[Issue 2251-implements-an-active-form-tracker](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2251)
- Implements the ActiveFormTracker
- And the changelog

compile-results

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Utilities/ActiveFormTracker.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Utilities/ActiveFormTracker.cs
@@ -10,202 +10,201 @@
  */
 #endregion
 
-namespace Krypton.Toolkit
-{
-    /// <summary>
-    /// Static class that tracks the active form within a WinForms application in an event drive way.<br/>
-    /// It removes the need for the use of Form.ActiveForm which relies on GetForeGroundWindow().
-    /// </summary>
-    public static class ActiveFormTracker
-    {
-        #region Private static values
-        private static volatile Form? _activeForm;
-        private static volatile Form? _activeMdiChild;
-        #endregion
+namespace Krypton.Toolkit;
 
-        #region Static Identity
-        static ActiveFormTracker()
+/// <summary>
+/// Static class that tracks the active form within a WinForms application in an event drive way.<br/>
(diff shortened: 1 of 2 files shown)


Source/Krypton Components/Krypton.Toolkit/Utilities/ActiveFormTracker.cs
```diff
@@ -10,202 +10,201 @@
...
-        /// <param name="e">Not used.</param>
-        private static void HandleDestroyedMdiChild(object? sender, EventArgs e)
+    /// <summary>
+    /// Unsubscribes from the child form events.
+    /// </summary>
+    /// <param name="sender">The source of the event.</param>
+    /// <param name="e">Not used.</param>
+    private static void HandleDestroyedMdiChild(object? sender, EventArgs e)
+    {
+        if (sender is Form form)
```
Smurf-IV, line 205: Should this check if this form is equal to the passed in form, what if it is a mdiForm containing a form ?

giduac: > Should this check if this form is equal to the passed in form, what if it is a mdiForm containing a form ?

See the Attach method ..

giduac: > Should this check if this form is equal to the passed in form, what if it is a mdiForm containing a form ?

All forms on being destroyed will also fire the the Deactivated() event before being destroyed, changing the state of their respective _active* forms vars

giduac: @PWagner1, @Smurf-IV & @lesandrog,

Ideas:
- We might add a public counter (getter) to the tracker that is incremented / decremented as forms subscribe resp. unsubscribe. That can be used from within other controls to see if the tracker is active.
- We might add a property to KForm for tracking to be turned on/off. What would be a default for that?

Here's a video from a test setup with some logging added to visualise the behaviour of the tracker.

https://github.com/user-attachments/assets/eaeedfd7-d74b-4750-9b6d-57c8294f1f71

Issue #1263, opened 2024-01-04
[Question]: Is it possible to create custom pre-processor directives


PWagner1:
Say I want to implement a property to test, but want it to be not accessible in the stable version. For example, in the `KryptonThemeComboBox`, I want to implement a `ReportThemeIndex` property, to show the index of a selected value? I would assume it would go something like:

```cs
   #if Nightly || Canary
   // Do something
   #endif
```

PWagner1:
```xml
<PropertyGroup Condition="'$(Configuration)' == 'Nightly'">
  <DefineConstants>$(DefineConstants);NIGHTLY</DefineConstants>
</PropertyGroup>

<PropertyGroup Condition="'$(Configuration)' == 'Canary'">
  <DefineConstants>$(DefineConstants);CANARY</DefineConstants>
</PropertyGroup>
```

PR #2259, opened 2025-06-24
* Implement #1263

PWagner1:
* Implement #1263 

image

@Smurf-IV & @giduac Will update documentation for usage

### DIFF
--- a/Source/Krypton Components/Krypton.Workspace/Krypton.Workspace 2022.csproj
+++ b/Source/Krypton Components/Krypton.Workspace/Krypton.Workspace 2022.csproj
@@ -99,17 +99,4 @@
     <Optimize>True</Optimize>
   </PropertyGroup>
 
-	<!--Custom pre-processor directives-->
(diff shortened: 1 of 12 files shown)


Source/Krypton Components/Krypton.Workspace/Krypton.Workspace 2022.csproj
```diff
@@ -99,4 +99,17 @@
     <Optimize>True</Optimize>
   </PropertyGroup>
 
+	<!--Custom pre-processor directives-->
```
Smurf-IV: Please use a single file, then use a shared reference into each of the other projects

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Workspace/Krypton.Workspace 2022.csproj
PWagner1: Addressed feedback

PR #2288, opened 2025-07-06
(feat) Added PaletteViewer to TestForm's StartScreen

### LINKED ISSUE #2287: [Feature Request]: Add Themes Colour Checker Gui from Tobitege
**Is your feature request related to a problem? Please describe.**
Adds a new form to Testform

tobitege:
Added PaletteViewer form to StartScreen and adjusted layout in TestForm.

Fixed index errror in PaletteMicrosoft365Blue to prevent runtime exception (out of bounds color index). 
Also fixed enum typos in comments, which are relevant for this tool!

Included Newtonsoft.Json package reference in project file.
Added manifest/config for HighDPI display support. 
StartScreen made slightly larger to not cut off buttons.

Fixes #2287 

Usage hints: specify path to Source folder; click "Add ALL" to fill grid with all compatible palettes; scroll down/across to find cells with error flags; "Missing value" means that the row's key is missing in the palette's array (can cause exceptions!).
Grid can be exported in CSV, JSON or XML format.

Development 100% directed by tobitege, 95% implemented by OpenAI's o3 model in close cooperation. 😁

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Palette Builtin/Microsoft 365/Bases/PaletteMicrosoft365Base.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Palette Builtin/Microsoft 365/Bases/PaletteMicrosoft365Base.cs
@@ -27,7 +27,7 @@ public abstract class PaletteMicrosoft365Base : PaletteBase
     private static readonly Padding _contentPaddingHeader2 = new Padding(2, 1, 2, 1);
     private static readonly Padding _contentPaddingDock = new Padding(2, 2, 2, 1);
     private static readonly Padding _contentPaddingCalendar = new Padding(2);
-    //private static readonly Padding _contentPaddingHeaderForm = new Padding(owningForm!.RealWindowBorders.Left, owningForm!.RealWindowBorders.Bottom / 2, 0, 0);         
+    //private static readonly Padding _contentPaddingHeaderForm = new Padding(owningForm!.RealWindowBorders.Left, owningForm!.RealWindowBorders.Bottom / 2, 0, 0);
     private static readonly Padding _contentPaddingLabel = new Padding(3, 1, 3, 1);
     private static readonly Padding _contentPaddingLabel2 = new Padding(8, 2, 8, 2);
     private static readonly Padding _contentPaddingButtonInputControl = new Padding(0);
--- a/Source/Krypton Components/Krypton.Toolkit/Palette Builtin/Microsoft 365/Non Official Themes/PaletteMicrosoft365Blue.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Palette Builtin/Microsoft 365/Non Official Themes/PaletteMicrosoft365Blue.cs
@@ -1,12 +1,12 @@
 ﻿#region BSD License
 /*
- * 
+ *
  * Original BSD 3-Clause License (https://github.com/ComponentFactory/Krypton/blob/master/LICENSE)
  *  © Component Factory Pty Ltd, 2006 - 2016, (Version 4.5.0.0) All rights reserved.
- * 
+ *
  *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
  *  Modifications by Peter Wagner (aka Wagnerp), Simon Coghlan (aka Smurf-IV), Giduac & Ahmed Abdelhameed et al. 2017 - 2025. All rights reserved.
- *  
+ *
  */
 #endregion
 
@@ -195,8 +195,8 @@ public class PaletteMicrosoft365Blue : PaletteMicrosoft365Base
         Color.FromArgb(162, 173, 185), // FormBorderHeaderInactive
         Color.FromArgb(193, 212, 236), // FormBorderHeaderActive1
         Color.FromArgb(187, 206, 230), // FormBorderHeaderActive2
-        Color.FromArgb(223, 235, 247), // FormBorderHeaderInctive1
-        Color.FromArgb(223, 235, 247), // FormBorderHeaderInctive2
+        Color.FromArgb(223, 235, 247), // FormBorderHeaderInactive1
+        Color.FromArgb(223, 235, 247), // FormBorderHeaderInactive2
         Color.FromArgb(30, 57, 91), // FormHeaderShortActive
         Color.FromArgb(106, 128, 168), // FormHeaderShortInactive
         Color.FromArgb(30, 57, 91), // FormHeaderLongActive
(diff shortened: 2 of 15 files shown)


giduac:
@tobitege, @lesandrog, @PWagner1, @Smurf-IV 

First testing the changes to Toolkit.

tobitege:
Hey @giduac 
thanks! I'll put up a small commit in a couple of minutes to include a README and an exception handler to avoid crashes due to theme indices missing.

giduac:
Hi @tobitege 

Just thinking it over if JSon is really needed. That would save pulling in the Newtonsoft nuget into the repo. 
XML and Csv is probably enough for the export....

What's your take on this?

tobitege:
> Hi @tobitege
> 
> Just thinking it over if JSon is really needed. That would save pulling in the Newtonsoft nuget into the repo. XML and Csv is probably enough for the export....
> 
> What's your take on this?

No problem, I'll remove it completely.

giduac:
@tobitege 

Proposed additions please:

Add a KThemeCombo
image

Make KDGV rows autosize (makes it a bit slower but will show all multi line string in the cells.

```cs
        private void BuildBaseRows()
        {
            this.dataGridViewPalette.Columns.Clear();
            this.dataGridViewPalette.Rows.Clear();
            this.dataGridViewPalette.AutoSizeRowsMode = DataGridViewAutoSizeRowsMode.AllCells;
```

tobitege:
Oh, since removing the Json nuget, I have to also remove my config file management, will fix asap

tobitege:
Fixed Json removal and row auto-sizing. Will add theme combo next.

giduac:
@tobitege 

**Little bug in reloading data:**
- start the checker
- press add palette
- press clear
- press add palette

Then api calls column remains empty.


**And another one:**
Can we have a Environment.NewLine after each api call so we'll get them nicely listed one at a time?

Current:
image

Desired layout:
image

And then when all is correct we can set columns auto sizing for api calls col by adding.
```cs
// Column 2: API Call(s)
colApi.AutoSizeMode = DataGridViewAutoSizeColumnMode.AllCells;
```

tobitege:
{35A2F175-C13B-446B-89BD-4CDF8FE17D0A}
commit incoming

giduac:
@tobitege 

We're getting there...

I think this can be disabled. I tried it and when turned off the autosizing works correct.
Give it a try please.

```cs
private void DataGridViewPalette_ColumnAdded(object sender, System.Windows.Forms.DataGridViewColumnEventArgs e)
{
    // auto-size once then freeze
    //int preferred = e.Column.GetPreferredWidth(System.Windows.Forms.DataGridViewAutoSizeColumnMode.AllCells, true);
    //preferred = System.Math.Max(MinColumnWidth, System.Math.Min(preferred, 300));
    //e.Column.Width = preferred;
    //e.Column.AutoSizeMode = System.Windows.Forms.DataGridViewAutoSizeColumnMode.None;
    //e.Column.SortMode = System.Windows.Forms.DataGridViewColumnSortMode.NotSortable;
}

```

tobitege:
Tbh, I'd rather leave that in because it assures to disable sorting and keeps column width between 120 and 300, as sanity checks.

giduac:
@tobitege 

Then please apply only if not col index 2.
```cs
private void DataGridViewPalette_ColumnAdded(object sender, System.Windows.Forms.DataGridViewColumnEventArgs e)
{
    // auto-size once then freeze
    int preferred = e.Column.GetPreferredWidth(System.Windows.Forms.DataGridViewAutoSizeColumnMode.AllCells, true);
    preferred = System.Math.Max( MinColumnWidth, System.Math.Min( preferred, 300 ) );
    e.Column.Width = preferred;
            
    if (e.Column.Index != 2)
    {
        e.Column.AutoSizeMode = System.Windows.Forms.DataGridViewAutoSizeColumnMode.None;
    }

    e.Column.SortMode = System.Windows.Forms.DataGridViewColumnSortMode.NotSortable;
}
```

tobitege:
Applied that column index check. Made theme selector combo wider so the lengthy names don't get truncated.

giduac:
@tobitege 

Shall we set the api calls col to frozen to keep it in view when scrolling horizontally?

https://github.com/user-attachments/assets/b33e9966-ca10-4678-8827-fed2e8ea8d5a

giduac:
@tobitege 

A bit in doubt if we shouldn't do a separate PR for the two palette files to keep changes to TK and testform apart since it flows into the release.

Smurf-IV:
Mt $0.02....
It would be nice if this was in the actual Palette designer repo.
Also, for the missing numbers to have some indication as to which are missing.

But, This is all good work ;-)

tobitege:
- API column is now frozen, which avoids paint issues while scrolling, too
- reverted value addition in PaletteMicrosoft365Blue.cs
- the changes in 2nd theme file should stay, they include enum name typo fixes and consistent formatting (helps with parsing tools like this!)
- fixed Cancel button positioning
This is as much time as I got for this now.

giduac:
> * API column is now frozen, which avoids paint issues while scrolling, too
> * reverted value addition in PaletteMicrosoft365Blue.cs
> * the changes in 2nd theme file should stay, they include enum name typo fixes and consistent formatting (helps with parsing tools like this!)
> * fixed Cancel button positioning
>   This is as much time as I got for this now.

@tobitege 

Nice work!

I'll do some more testing and get the palette files merged in separate PR later on.

Let you know how far I get....

PWagner1:
> Mt $0.02.... It would be nice if this was in the actual Palette designer repo. Also, for the missing numbers to have some indication as to which are missing.
> 
> But, This is all good work ;-)

@Smurf-IV 

I agree.

@tobitege Nice work, would you mind creating a separate PR for the Palette designer repo also? :)

giduac:
@tobitege 

The PR for the palette files is in.
https://github.com/Krypton-Suite/Standard-Toolkit/pull/2289

You can remove the palette file from this PR.

Still need to test the source and export features....

giduac:
@lesandrog, @Smurf-IV & @PWagner1,

This PR cannot be merged now.
We signal when testing and the PR is complete to proceed.
Until then please leave this alone.

Thanks.

tobitege:
> > Mt $0.02.... It would be nice if this was in the actual Palette designer repo. Also, for the missing numbers to have some indication as to which are missing.
> > But, This is all good work ;-)
> 
> @Smurf-IV
> 
> I agree.
> 
> @tobitege Nice work, would you mind creating a separate PR for the Palette designer repo also? :)

Which repo is that exactly? 🤔

giduac:
@tobitege 

If you like you can update this branch with the latest commit from alpha.
Then all should be on par...

tobitege:
Done :)

PWagner1:
> > > Mt $0.02.... It would be nice if this was in the actual Palette designer repo. Also, for the missing numbers to have some indication as to which are missing.
> > > But, This is all good work ;-)
> > 
> > 
> > @Smurf-IV
> > I agree.
> > @tobitege Nice work, would you mind creating a separate PR for the Palette designer repo also? :)
> 
> Which repo is that exactly? 🤔

@tobitege 

It's this one [Krypton-Suite/Theme-Palettes](https://github.com/Krypton-Suite/Theme-Palettes)

giduac:
@tobitege 

Had a first look at the "Browse Source" button. And these are the first things I found
- Once the sourcepath has been selected/set it cannot be cleared. 
- When the tool is run from another repo on the same system and the source path was saved to the settings file earlier it will use this specific path.

Also it is not clear/documented in the readme what this option actually does next to the already present functionality in the GUI...?

tobitege:
The tool only supports one repo at a time, thus the source path is remembered once, hadn't tried with separate repos.

The app does 2 different checks: 
1) based on the available enums via API calls and
2) actually parsing source files for the palette's color arrays' indexes and their comments, which give the best indication WHICH enum they are adressing (thus can find missing enums fairly well).

The source path _could_ point to a different repo, but since the app is compiled against whatever sources you got checked out, I wanted the user to at least be able to point to the source folder in case someday the structure changes.
I could add a simple button to clear the source path, so that it won't scan files, but that would disable a vital check, basically.

giduac:
@tobitege 

So entering the source directory is mandatory for correct functioning and must be set before a palette can be added to the grid?

tobitege:
You can still add palettes (except the unsupported Sparkle ones):

without access to the sources of the palette files, it can only show the values/colors of their public `ColorTable`, but cannot analyze any relationships between index and enum value (it does italic empty or transparent values, though, but that isn't an error necesserily).
The sources all have comments next to their color entry with their enum name (thankfully!).

Only with the latter we can find actual, missing color values and flag them with an error icon to their cell (toolhint). Only the entries at the very end of the table could be checked, if the table has less entries than the enum (232) itself.

After work I'll add some more explanations to the README and add the "Clear path" button, just to be clear.

giduac:
@tobitege 

Should we make the path mandatory and always perform that relationship check ?

tobitege:
> @tobitege
> 
> Should we make the path mandatory and always perform that relationship check ?

Hmm... not sure if that's needed. IF someone compiles against NuGet instead of source, they could still use the minimal display version?

giduac:
> > @tobitege
> > Should we make the path mandatory and always perform that relationship check ?
> 
> Hmm... not sure if that's needed. IF someone compiles against NuGet instead of source, they could still use the minimal display version?

It makes not much sense to run it from nuget.
The whole point of the exercise is to get themes working correctly before a release is made.
And that's our task... imo

tobitege:
Alright, not a problem, was just clarifying.
Will make it mandatory with next commit 😃

tobitege:
{0DBC2994-31C0-4ECD-8512-ABF1C32D1D5F}
@giduac - source path mandatory now with smart folder checking and "Required" hint if missing (and clear button).

tobitege:
Updated the README.md file with source path explanations.

This shall do it from my side for this PR now.

giduac:
@tobitege  

Over and out for today....

tobitege:
Added one more fix for high DPI handling across frameworks to make compiler happy.

giduac:
Fine so far.
Will merge this now and do some updating afterwards.

PR #2293, opened 2025-07-09
2287-V100-add-testform-themes-colour-checker-gui

giduac:
[Issue 2287-add-testform-themes-colour-checker-gui](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2287)
- Small updates to the Palette Viewer
- No changelog entry required

### DIFF
--- a/Source/Krypton Components/TestForm/Classes/WindowStateStore.cs
+++ b/Source/Krypton Components/TestForm/Classes/WindowStateStore.cs
@@ -29,9 +29,13 @@ namespace TestForm
             _filePath = Path.Combine(dir, "windowstate.ini");
         }
 
-        public WindowStateInfo Load()
+        public WindowStateInfo? Load()
         {
-            if (!File.Exists(_filePath)) return null;
+            if (!File.Exists(_filePath))
+            {
+                return null;
+            }
+
             var dict = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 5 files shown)


Source/Krypton Components/TestForm/Classes/WindowStateStore.cs
```diff
@@ -60,9 +64,9 @@ int ParseInt(string key)
                 State = state,
                 SelectedTab = ParseInt("SelectedTab"),
                 ContentSplitterDistance = ParseInt("ContentSplitterDistance"),
-                LastWtsFolder = dict.TryGetValue("LastWtsFolder", out var l1) ? l1 : null,
-                LastTheme = dict.TryGetValue("LastTheme", out var l2) ? l2 : null,
-                SourcePath = dict.TryGetValue("SourcePath", out var l3) ? l3 : null
+                LastWtsFolder = dict.TryGetValue("LastWtsFolder", out var l1) ? l1 : string.Empty,
```
tobitege, line 67: Oh, LastWtsFolder is a fragment from an earlier use of the class and can be removed, I overlooked that.

PWagner1 (approved):

PWagner1:
> Not sure all this "?" syntax is needed, is the ReSharper style?

@tobitege It might be the conditional operator introduced with C# 8 or 9. See https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/conditional-operator

PR #2299, opened 2025-07-11
2104-V100-fix-palette-base-leak

tobitege:
Non-static, instance-wired event handlers for PaletteBase-derived classes do not dispose automatically.

Analysis of the leak:

1. SystemEvents.UserPreferenceChanged is a static event.  
   • If objects subscribe from an instance constructor and do NOT unsubscribe, the static delegate keeps a reference to that instance, preventing it from being garbage-collected and causing a memory leak.

2. In this codebase 36 files perform the subscription.  
   • 7 UI classes (VisualPanel, VisualForm, etc.) already detach in their Dispose logic – they are safe.  
   • Most colour-table classes subscribe in a static constructor; static delegates → static handlers live for the whole process, so they do not cause an expanding leak.  
   • The Palette hierarchy (PaletteBase and every palette theme that derives from it) subscribed in the **instance** constructor but never detached. Each time a new palette object was created it was pinned in memory forever.

Root cause: PaletteBase did not unsubscribe, causing the leak.

Fixes #2104 

Fix applied:

PaletteBase now overrides Dispose and:

• Unhooks SystemEvents.UserPreferenceChanged  
• Releases font resources via existing DisposeFonts()  
• Calls base.Dispose(disposing)

This single change automatically repairs every derived palette/theme class because they all inherit the new behaviour, solving the leak across the entire codebase without touching dozens of theme files.

No other classes required modification, as all remaining subscriptions are either already balanced or static-by-design.

Issue #2104, opened 2025-01-27
[Bug]: Memory leak when disposing Krypton.Toolkit.PaletteBase and derived classes.


MarthinusR:
**Describe the bug**
The class "Krypton.Toolkit.PaletteBase" does not unsubscribe its "OnUserPreferenceChanged" method from the static rooted event "SystemEvents.UserPreferenceChanged", leading to objects persistently maintained in second generation garbage, i.e. memory leaks.

Derived classes also instantiate member controls that do not get disposed and become dangling objects.

**To Reproduce**
1. Start the "TestForm" project.
2. Click on "Buttons" to open the "ButtonsTest" form.
3. Click on the close button.
4. "PaletteBase" objects will not be collected on close.

**Expected behavior**
All "PaletteBase" objects must be garbage collected.

**Screenshots**
![Image](https://github.com/user-attachments/assets/bc2e47d8-0413-42c8-9ed6-805ff6f6b0b2)

**Desktop:**
 - OS: Windows 11
 - Version: 10.0.22621
 - Framework/.NET Version: 4.7.2 
 - Toolkit Version: 90.25.1.27

**Additional context**
Some work has been done ([Krypton_memory_fix.patch](https://github.com/user-attachments/files/18556434/Krypton_memory_fix.patch)) to clean up resources, but due to unfamiliarity with the project it is unclear who the oner of created objects are.  KryptonReadOnlyControls espesially is a point of concern and should be addressed. The provided patch might not be stable.

giduac:
Discussion opened on the KDev server to review this proposal:
https://discord.com/channels/1261930150585569350/1333697703154286652

MarthinusR:
The patch "Krypton_master_abfebc02_memory_fix_20250210_0809.patch" roughly provides a 5 fold reduction in memory usage in our software. Very basic testing was done and it seems to be stable. For the full effect to take place, one must ensure that all controls that are created in the designer gets disposed when a form closes (take note that some controls created in the InitializeComponent(), might not be attached to the form's "Controls", and these might have to be disposed manually).

This patch is intended to be used on the master branch at commit id abfebc02. A corresponding patch for the alpha branch is provided at commit id 2803aec6, though it was not tested, as this revision cannot run on our software at this time.

Take note that the intent of this patch is to get a reasonable amount of memory to be deallocated, and there is place for improvement as it was done without finesse to get at a reasonable position.

[...]

PR #2301, opened 2025-07-11
2104-V85-LTS-fix-palette-base-leak 

tobitege:
Non-static, instance-wired event handlers for PaletteBase-derived classes do not dispose automatically.

Analysis of the leak:

1. SystemEvents.UserPreferenceChanged is a static event.  
   • If objects subscribe from an instance constructor and do NOT unsubscribe, the static delegate keeps a reference to that instance, preventing it from being garbage-collected and causing a memory leak.

2. In this codebase several files perform the subscription.  
   • The Palette hierarchy (PaletteBase and every palette theme that derives from it) subscribed in the **instance** constructor but never detached. Each time a new palette object was created it was pinned in memory forever.

Root cause: PaletteBase did not unsubscribe, causing the leak.

Relates to #2104 #2299 #2300 

Fix applied:

PaletteBase now overrides Dispose and:

• Unhooks SystemEvents.UserPreferenceChanged  
• Releases font resources via existing DisposeFonts()  
• Calls base.Dispose(disposing)

This single change automatically repairs every derived palette/theme class because they all inherit the new behaviour, solving the leak across the entire codebase without touching dozens of theme files.

No other classes required modification, as all remaining subscriptions are either already balanced or static-by-design.

### DIFF
--- a/Documents/Help/Changelog.md
+++ b/Documents/Help/Changelog.md
@@ -3,6 +3,7 @@
 =======
 
 # 2025-08-25 - Build 2508 (Patch 8) - August 2025
+* Resolved [#2301](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2301), Fix memory leak in PaletteBase
 * Resolved [#2235](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2235), `OSUtilities` Adds OsVersionInfo to the properties.
 * Resolved [#2264](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2264), Implementation of style: `PaletteBackStyle.Control`, `PaletteContentStyle.LabelAlternateControl` and `PaletteContentStyle.LabelAlternatePanel` in all themes.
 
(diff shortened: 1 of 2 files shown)


giduac:
@tobitege 

This is a structural problem throughout the TK as 2104 shows.
It needs a structural approach but will time consuming. 
Probably some more discussion and planning needed to get this on the road. But can be done on a per item basis. 
Opened a topic on KDServer

tobitege:
Running out of time doesn't mean there aren't users still using it. ;)

giduac:
Comment updated ! 👍

giduac (approved):

PWagner1:
@giduac & @tobitege 

Planning to push a canary build out for **both** V85 & V100 next weekend

giduac:
> @giduac & @tobitege
> 
> Planning to push a canary build out for **both** V85 & V100 next weekend

@PWagner1 
Can you do a nightly for V95 as well somewhere next week.

PWagner1:
> > @giduac & @tobitege
> > Planning to push a canary build out for **both** V85 & V100 next weekend
> 
> @PWagner1 Can you do a nightly for V95 as well somewhere next week.

@giduac 

Unfortunately not, V100 only (as the NuGet feeds and IDs will get even more complex) :)

PR #2311, opened 2025-07-17
2312-V100-fix-GDI-resource-handling

### LINKED ISSUE #2312: [Bug]: GDI resource issues (V100)
Numerous controls in the suite used a custom double-buffering routine that:

Created a compatible bitmap.
Selected that bitmap into a long-lived cached memory DC.
Drew the control’s visuals into that DC.
Immediately destroyed the bitmap – while it was still selected into the DC.
Deleting a GDI object that remains selected corrupts the device context.
Depending on timing and palette colours the next paint cycle can fill the client area with solid black, display completely wrong background colours, or show random artefacts.

tobitege:
## Summary  

Numerous controls in the suite used a custom double-buffering routine that:

1. Created a compatible bitmap.  
2. Selected that bitmap into a long-lived cached memory DC.  
3. Drew the control’s visuals into that DC.  
4. Immediately destroyed the bitmap – **while it was still selected** into the DC.

Deleting a GDI object that remains selected corrupts the device context.  
Depending on timing and palette colours the next paint cycle can fill the client area with solid black, display completely wrong background colours, or show random artefacts.  

Fixes #2312 

### What was changed (High-level)  

* Audited every `PI.SelectObject`/`PI.DeleteObject` pair in the toolkit.  
* Recorded the return value of `SelectObject` (the previously selected GDI object).  
* Restored the original object to the DC **before** disposing of the temporary bitmap, always in a `try/​finally` block to guarantee clean-up on exceptions.  

## Fixed components

- KryptonCheckedListBox.cs
- KryptonComboBox.cs
- KryptonDomainUpDown.cs
- KryptonListBox.cs
- KryptonListView.cs
- KryptonNumericUpDown.cs
- KryptonPropertyGrid.cs
- KryptonTreeView.cs
- VisualForm.cs

---

## Concrete example – `KryptonPropertyGrid`  

Before:  

```csharp
IntPtr hBitmap = PI.CreateCompatibleBitmap(hdc, width, height);
PI.SelectObject(_screenDC, hBitmap);
// render…
PI.DeleteObject(hBitmap);   // <- deletes while still selected
```

After:  

```csharp
IntPtr hBitmap  = PI.CreateCompatibleBitmap(hdc, width, height);
IntPtr oldBmp   = PI.SelectObject(_screenDC, hBitmap);
try
{
    // render…
}
finally
{
    PI.SelectObject(_screenDC, oldBmp);   // restore first
    PI.DeleteObject(hBitmap);             // safe to delete now
}
```

The same pattern has been applied to all painting helpers that relied on off-screen bitmaps, eliminating intermittent black or wrongly coloured backgrounds across the toolkit.

---

## TestForm `PropertyGrid` - “Run Stress Test” button

1. Baseline GDI-handle count  
    • `GetGuiResources(ProcessHandle, 0)` returns the number of GDI objects (bitmaps, pens, brushes, …) currently owned by the process.  
    • We store that value in `gdiStart`.

2. Prepare a reusable bitmap buffer  
    • A single `Bitmap` the size of the `KryptonPropertyGrid` is created once (ARGB 32-bit).  
    • Re-using the same bitmap avoids creating 500 extra GDI handles inside the test itself.

3. Stress loop (500 iterations)  
    For each iteration `i`:
    a. `kpgExample.Invalidate()` – marks the control dirty.  
    b. `kpgExample.Update()` – forces a synchronous paint, so `InternalPropertyGrid.WmPaint` is executed immediately.  
    c. `DrawToBitmap` copies the freshly painted control into the bitmap buffer.  
    d. `IsBitmapSolid` scans the bitmap:  
       - If every pixel has the same colour the control probably painted a solid rectangle (classic symptom of the DC-corruption bug).  
       - If that happens an `InvalidOperationException` is thrown with the failing iteration number.

4. Final GDI-handle verification  
    a. We read the handle count (`gdiEnd`) once, then run `GC.Collect(); GC.WaitForPendingFinalizers(); GC.Collect();` to ensure any disposable bitmaps created by WinForms are released.  
    b. We re-read the count to include the effect of garbage collection.  
    c. Small fluctuations (theme pens, cached brushes, etc.) are normal, so we allow a difference of up to 3 handles (`AllowedLeak`).  
       - If `gdiEnd - gdiStart > AllowedLeak`, an exception is thrown reporting the suspected leak.

5. Outcome dialog  
    • If no exception was thrown the test shows:  
      “Stress test completed successfully.”  
    • Any exception shows “KryptonPropertyGrid Test Failed” plus the exception message.

In short:
• The test pounds the property grid with 500 consecutive paints.  
• It fails if a paint produces an all-solid bitmap (black or any other colour) – a proxy for the original DC-corruption issue.  
• It also flags up any substantial GDI handle growth that would indicate unreleased bitmaps or other leaks.

### DIFF
--- a/Source/Krypton Components/TestForm/PropertyGridTest.Designer.cs
+++ b/Source/Krypton Components/TestForm/PropertyGridTest.Designer.cs
@@ -31,6 +31,7 @@
             this.kryptonPanel1 = new Krypton.Toolkit.KryptonPanel();
             this.kpgExample = new Krypton.Toolkit.KryptonPropertyGrid();
             this.kryptonThemeComboBox1 = new Krypton.Toolkit.KryptonThemeComboBox();
+            this.kbtnStressTest = new Krypton.Toolkit.KryptonButton();
             ((System.ComponentModel.ISupportInitialize)(this.kryptonPanel1)).BeginInit();
             this.kryptonPanel1.SuspendLayout();
             ((System.ComponentModel.ISupportInitialize)(this.kryptonThemeComboBox1)).BeginInit();
@@ -38,6 +39,7 @@
             // 
             // kryptonPanel1
             // 
+            this.kryptonPanel1.Controls.Add(this.kbtnStressTest);
             this.kryptonPanel1.Controls.Add(this.kryptonThemeComboBox1);
             this.kryptonPanel1.Controls.Add(this.kpgExample);
             this.kryptonPanel1.Dock = System.Windows.Forms.DockStyle.Fill;
@@ -48,11 +50,11 @@
             // 
             // kpgExample
             // 
-            this.kpgExample.Location = new System.Drawing.Point(13, 41);
+            this.kpgExample.Location = new System.Drawing.Point(13, 69);
             this.kpgExample.Name = "kpgExample";
             this.kpgExample.Padding = new System.Windows.Forms.Padding(1);
             this.kpgExample.SelectedObject = this.kpgExample;
-            this.kpgExample.Size = new System.Drawing.Size(362, 485);
+            this.kpgExample.Size = new System.Drawing.Size(362, 457);
             this.kpgExample.TabIndex = 0;
             this.kpgExample.Text = "kryptonPropertyGrid1";
             // 
@@ -66,6 +68,15 @@
             this.kryptonThemeComboBox1.Size = new System.Drawing.Size(362, 22);
             this.kryptonThemeComboBox1.StateCommon.ComboBox.Content.TextH = Krypton.Toolkit.PaletteRelativeAlign.Near;
             this.kryptonThemeComboBox1.TabIndex = 1;
+            // 
+            // kbtnStressTest
+            // 
+            this.kbtnStressTest.Location = new System.Drawing.Point(13, 41);
+            this.kbtnStressTest.Name = "kbtnStressTest";
+            this.kbtnStressTest.Size = new System.Drawing.Size(362, 22);
+            this.kbtnStressTest.TabIndex = 2;
+            this.kbtnStressTest.Values.Text = "Run Stress Test";
+            this.kbtnStressTest.Click += new System.EventHandler(this.kbtnStressTest_Click);
             // 
             // PropertyGridTest
             // 
@@ -89,5 +100,6 @@
         private KryptonPanel kryptonPanel1;
         private KryptonPropertyGrid kpgExample;
         private KryptonThemeComboBox kryptonThemeComboBox1;
+        private Krypton.Toolkit.KryptonButton kbtnStressTest;
     }
 }
--- a/Source/Krypton Components/TestForm/PropertyGridTest.cs
+++ b/Source/Krypton Components/TestForm/PropertyGridTest.cs
@@ -7,6 +7,10 @@
  */
 #endregion
 
+using System.Diagnostics;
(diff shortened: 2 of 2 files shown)


tobitege:
{AD788182-2580-4656-AC01-4B4CC5A953AD}

tobitege:
Investigating deeper, underlying paint issues currently. There's something else wrong with the DC handling.

tobitege:
Just by chance I started looking into the RTL toggle issue and after 2 hours I'm throwing the towel.
This PR deals with a different issue afterall.

Here's a summary of my coding buddy about the RTL topic itself:
I’ve now burned through every safe Win-Forms option (handle recreation, DC rebuild, forced redraw, simulated input, delayed invalidation, etc.).  
The remaining facts are:

1. `PropertyGrid` + custom buffered painting was never designed to be flipped to `RightToLeft = Yes` at runtime.  
   • Microsoft’s own documentation warns that any control derived from `ScrollableControl` “acts improperly” under dynamic RTL toggles.  
   • Internally `PropertyGrid` re-lays out only part of its tree; the next focus/scroll navigation invalidates a different clip-rect, so rows are painted over with whatever colour GDI last used (black in Dark mode).

2. The very first focus event after RTL toggle is what repaints a single line and hides the symptom—that’s why clicking the value cell or tabbing in/out appears to “fix” it for a moment. As soon as the underlying scrollable host re-scrolls or re-focuses, the artefact returns.

3. Any attempt to force repaint externally (timer, `RedrawWindow`, `SendKeys`, fake mouse, size-jiggle) either:
   • happens too early (before the grid’s internal mirror pass finishes) or  
   • re-triggers the same broken paint path and the bars come back.

Because the defect lives inside native `PropertyGrid` code and GDI mirroring, no amount of managed-side repainting will solve it reliably.

What you **can** do:

A. Design-time RTL only  
   If you must show RTL, create a second form or user-control built with `RightToLeft = Yes` from the start and load that version instead of toggling on the fly.

B. Full re-instantiate the form  
   Close the current test window and open a new one where the grid is created with the desired RTL value. (That’s what Visual Studio’s designer “Reset” effectively does.)

C. Use a third-party or WPF grid that supports dynamic RTL  
   WPF’s `PropertyGrid` (or popular MVVM-friendly grids) handle mirroring at runtime. Embedding one in an ElementHost avoids the Win-Forms limitation entirely.

D. Keep RTL toggle hidden in production  
   If end-users never need to flip RTL at runtime, expose the setting only in config or via a command-line switch and construct the grid once at startup.

Anything short of those work-arounds will keep running into the same GDI-level bug.  
So at this stage the practical path forward is to remove the runtime toggle from the demo and document the limitation rather than continue chasing a fix that the framework itself cannot deliver.

Smurf-IV:
My $0.02
This would be better as a `using` disposable pattern, as then all the `try - finally` would also disappear.

tobitege:
> My $0.02 This would be better as a `using` disposable pattern, as then all the `try - finally` would also disappear.

But that wouldn't work here as the using only takes care of the "g" (Graphics) object for disposal,
and not execute what is part of the `finally` block, which could differ from component to component?

Edit: ohh... wrapping all objects into helper classes etc... hmm... will look into it.

Looked into it: that is another PR for itself, tbh. imho it wouldn't improve readability or make targeted fixes easer to apply or debug.

PR #2314, opened 2025-07-18
2178-V95-Fix-KryptonPropertyGrid-Reset-menu-item

### LINKED ISSUE #2178: [Bug]: `KryptonPropertyGrid` 'Reset' option
The `KryptonPropertyGrid` 'Reset' option is not 'greyed' out, even if the value of the property is the default value. Related to #2108.

https://github.com/user-attachments/assets/b889877c-87b7-4bae-b6b1-543c74939098

tobitege:
### Problem  
In `KryptonPropertyGrid` the “Reset” item in the custom context-menu was:  
1. Always shown as enabled because the enable logic relied only on the presence of a `DefaultValueAttribute`.  
2. Updated inside a `PropertyGrid_MouseDown` handler that never fires when the user right-clicks a property row (the underlying `PropertyGrid` sends `WM_CONTEXTMENU` directly).  
As a result the item was never greyed out for properties already at their default value, and any code inside `PropertyGrid_MouseDown` was effectively dead.

### Fix  
1. **Reliable trigger:**  
   • Moved the enable/disable logic to the `WM_CONTEXTMENU` branch of `WndProc`, which **always executes** for right-clicks on property rows.  
2. **Accurate enable check:**  
   • New helper `CanResetCurrentProperty()`  
     – Iterates through all selected objects.  
     – Uses `PropertyDescriptor.CanResetValue(obj)`; if that fails, falls back to comparing the current value with `DefaultValueAttribute`.  
3. **Dead code removal:**  
   • `PropertyGrid_MouseDown` is now deleted along with its subscription.

### Outcome  
“Reset” is only enabled when the selected property differs from its default.  
No unnecessary event handler remains.

Fixes #2178 

{F392BFAD-6401-4553-A261-99757BF8583A}

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonPropertyGrid.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonPropertyGrid.cs
@@ -1,7 +1,7 @@
 ﻿#region BSD License
 /*
  *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
- *  Modifications by Peter Wagner(aka Wagnerp) & Simon Coghlan(aka Smurf-IV), et al. 2021 - 2024. All rights reserved.
+ *  Modifications by Peter Wagner(aka Wagnerp) & Simon Coghlan(aka Smurf-IV), tobitege et al. 2021 - 2025. All rights reserved.
  */
 #endregion
 
@@ -57,7 +57,7 @@ namespace Krypton.Toolkit
             }
 
             /// <summary>
-            /// Releases all resources used by the Control. 
+            /// Releases all resources used by the Control.
             /// </summary>
             /// <param name="disposing">true to release both managed and unmanaged resources; false to release only unmanaged resources.</param>
             protected override void Dispose(bool disposing)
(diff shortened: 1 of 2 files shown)


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonPropertyGrid.cs
```diff
@@ -968,6 +944,52 @@ private void OnResetClick(object? sender, EventArgs e)
             }
         }
 
+        /// <summary>
+        /// Determine if currently selected property can be reset.
+        /// </summary>
+        /// <returns>True when reset is possible.</returns>
+        private bool CanResetCurrentProperty()
+        {
+            if (_propertyGrid.SelectedGridItem is not GridItem selectedGridItem || selectedGridItem.PropertyDescriptor is null)
```
Smurf-IV, line 953: Just a minor thing.. I prefer the combining logic to go onto a separate line, i.e. starting `||`
This allows more complex things to be easier read.

Smurf-IV (approved):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonPropertyGrid.cs
tobitege: Will remember that for next time.

PR #2316, opened 2025-07-18
2312-KPropGridTest-enhancements-Followup-V100

tobitege:
Enhance PropertyGridTest with 2nd stress testing functionality

Followup to #2312  [Bug]: GDI resource issues (V100)

- Introduced a new button for running a stress test demonstrating the bug as if the fix wasn't applied (general GDI issue).
- Updated stress test to track paint failures and provide detailed results.

### DIFF
--- a/Source/Krypton Components/TestForm/PropertyGridTest.cs
+++ b/Source/Krypton Components/TestForm/PropertyGridTest.cs
@@ -2,7 +2,7 @@
 /*
  *
  *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
- *  Modifications by Peter Wagner(aka Wagnerp) & Simon Coghlan(aka Smurf-IV), et al. 2024 - 2025. All rights reserved.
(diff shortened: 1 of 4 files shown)


PWagner1 (approved):

Source/Krypton Components/TestForm/PropertyGridTest.cs
```diff
@@ -38,41 +39,195 @@ private void kbtnStressTest_Click(object sender, EventArgs e)
...
+                         $"Bug Reproduced (Success): {successes}\n" +
+                         $"Bug Not Reproduced (Fail): {reproductionFailures}";
+
+        KryptonMessageBox.Show(message, title);
+    }
+
     [DllImport("user32.dll")]
     private static extern int GetGuiResources(IntPtr hProcess, int uiFlags);
 
+    [DllImport("gdi32.dll", SetLastError = true)]
```
Smurf-IV, line 175: Note:
Most of these are in the PI class, And if not then please add them

tobitege: Will address it with issue #2324  👍

PR #2319, opened 2025-07-19
2196-V100-KCommand-KContextMenuItem-enhancements

### LINKED ISSUE #2196: [Question]: KryptonContextMenu
### Discussed in https://github.com/Krypton-Suite/Standard-Toolkit/discussions/2195

<div type='discussions-op-text'>

<sup>Originally posted by **yanchenghan** April 21, 2025</sup>
It a nice Job.  But Here One kryptonContextMenu's submenu item(button) only can call a kryptonCommand*.   When there are many menus, a lot of kryptonCommand must are needed.This will be very chaotic. Two or more  submenu item(button) can call one kryptonCommand where Use if as needed Elseif... to judge, to call different function blocks</div>

tobitege:
Enhance KryptonContextMenuItem and KryptonCommand functionality

- Added `CommandParameter` property to `KryptonContextMenuItem` for passing parameters to commands.
- Introduced a new `PerformExecute` method in `KryptonCommand` to allow sender specification.
- Updated `ButtonsTest` in `TestForm` app to demonstrate command execution with parameters.
- Added custom `ReferenceConverter` to fix `Property value is not valid` design time error when trying to change command assignments.

Fixes #2196 
---
This introduces several enhancements to command handling for `KryptonContextMenuItem`.

### 1. Command `Execute` Sender Identity

-   **Previously**: When multiple menu items shared the same `KryptonCommand`, the `Execute` event handler would always receive the command object itself as the `sender`. This made it impossible to determine which specific menu item was clicked without creating separate command instances or using individual `Click` events.
-   **Now**: The `KryptonContextMenuItem` instance that initiates the action is passed as the `sender` to the command's `Execute` event. This allows the event handler to cast the sender and identify the exact originator of the command.

    A single `KryptonCommand` can be assigned to multiple items, and the handler can differentiate them based on their properties (`Text`, `Tag`, etc.).

    ```csharp
    void Command_Execute(object sender, EventArgs e)
    {
        var item = (KryptonContextMenuItem)sender;   // <-- The exact originating item
        // ...
    }
    ```

### 2. Design-Time `CommandParameter`

A `CommandParameter` property has been added to `KryptonContextMenuItem`. This provides a formal mechanism for distinguishing items at design-time without needing to parse the `Text` property.

```csharp
void Command_Execute(object sender, EventArgs e)
{
    var param = ((KryptonContextMenuItem)sender).CommandParameter;
    switch (param)
    {
        case "OpenFile": /* ... */ break;
        case "SaveFile": /* ... */ break;
        case "CloseApp": /* ... */ break;
    }
}
```

### 3. Improved Designer Experience

A custom `ReferenceConverter` has been implemented for the `KryptonCommand` property on `KryptonContextMenuItem`. This resolves a design-time issue where a "Property value is not valid" error would occur if the referenced `KryptonCommand` was removed from the designer.

---

### Implementation Rationale

The separation of concerns between the command and the menu item is key to this design.

-   **`KryptonCommand`**: A shared object that can be attached to many UI elements. It exposes state common to all of them (e.g., `Enabled`, `Checked`, `Text`).
-   **`KryptonContextMenuItem`**: The actual control the user clicks. While multiple items can share a command, each must retain its own identity.

The `CommandParameter` parameter allows each to carry a unique value while sharing the command's logic and state.

```txt
+-------------   one object   -------------+
¦              KryptonCommand              ¦
¦   Execute event fires once per click     ¦
+-------------------------------------------+
       ?           ?              ?
       ¦ shared    ¦ shared       ¦ shared
+----------+   +--------+     +--------+
¦ Item A   ¦   ¦ Item B ¦     ¦ Item C ¦   <-- separate controls
¦ Param=1  ¦   ¦ Param=2¦     ¦ Param=3¦
+----------+   +--------+     +--------+
```

Inside a single `Execute` handler, the implementation could be as follows:

```csharp
void Command_Execute(object sender, EventArgs e)
{
    var item = (KryptonContextMenuItem)sender;
    var param = item.CommandParameter as string;
    switch (item.CommandParameter)
    {
        case "1": // ... action for Item A
            break;
        case "2": // ... action for Item B
            break;
        case "3": // ... action for Item C
            break;
    }
}
```

This architecture allows a single `KryptonCommand` to synchronize state across different controls (buttons, ribbon items, etc.) while providing a clear way to identify which specific menu item triggered the action.

---

This is a demo setup within the `ButtonsTest` form:

1st and 2nd context menu items share the same event handler, but have different text and `CommandParameter` (the command's `Text` property is left empty so the context menu items' `Text` is used):
image
image

Clicks on the context menu items now allow to distinguish between the calling sender in the event handler (CommandParameter), so the same action may be used on several menu items:
image
image
image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/ContextMenu/KryptonContextMenuItem.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/ContextMenu/KryptonContextMenuItem.cs
@@ -1,12 +1,12 @@
 ﻿#region BSD License
 /*
- * 
+ *
  * Original BSD 3-Clause License (https://github.com/ComponentFactory/Krypton/blob/master/LICENSE)
  *  © Component Factory Pty Ltd, 2006 - 2016, (Version 4.5.0.0) All rights reserved.
- * 
+ *
  *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
- *  Modifications by Peter Wagner (aka Wagnerp), Simon Coghlan (aka Smurf-IV), Giduac & Ahmed Abdelhameed et al. 2017 - 2025. All rights reserved.
- *  
+ *  Modifications by Peter Wagner (aka Wagnerp), Simon Coghlan (aka Smurf-IV), Giduac, Ahmed Abdelhameed, tobitege et al. 2017 - 2025. All rights reserved.
+ *
  */
 #endregion
 
@@ -24,6 +24,14 @@ namespace Krypton.Toolkit;
 [DefaultEvent(nameof(Click))]
 public class KryptonContextMenuItem : KryptonContextMenuItemBase
 {
+    #region Nested Classes
+    // Provides proper design-time reference conversion for the KryptonCommand property even when the item is not sited.
+    private sealed class KryptonCommandReferenceConverter : ReferenceConverter
+    {
+        public KryptonCommandReferenceConverter() : base(typeof(KryptonCommand)) { }
+    }
+    #endregion
+
     #region Instance Fields
     private bool _enabled;
     private bool _splitSubMenu;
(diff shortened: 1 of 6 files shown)


PWagner1 (approved):

tobitege:
Changelog incoming...

giduac (approved):

tobitege:
@giduac would like to get a 2nd approval :)

Issue #2220, opened 2025-06-08
[Bug/Feature Request]: `KryptonLabel` does not support surrogates


neumatho:
As the topic says, KryptonLabel does not support unicode surrogates when rendering. Take this example:

Hello 👾

In the screenshot below, the first label is a normal WinForm label control. The second is a KryptonLabel. Both uses Segoe UI font size 18.

![Image](https://github.com/user-attachments/assets/4c12c7f8-67e1-4f1a-88fa-7e2bd5a5b257)

Added sample code surrogates (courtesy of our good friend Claude):

```cs
// 1. Emoji: Smiling Face with Heart-Eyes
string heartEyes = "\uD83D\uDE0D";

// 2. Emoji: Rocket
string rocket = "\uD83D\uDE80";

// 3. Flag of Japan
string japanFlag = "\uD83C\uDDEF\uD83C\uDDF5";

// 4. Musical Note
string musicalNote = "\uD83C\uDFB5";

// 5. Snowman with Snow
string snowman = "\u26C4";

// 6. Sparkles
string sparkles = "\u2728";

// 7. Grinning Face
string grinningFace = "\uD83D\uDE00";

// 8. Thumbs Up
string thumbsUp = "\uD83D\uDC4D";

// 9. Rainbow
string rainbow = "\uD83C\uDF08";

// 10. Alien Monster
string alienMonster = "\uD83D\uDC7E";

// Example of printing and using these surrogates
Console.WriteLine(heartEyes);     // Prints 😍
Console.WriteLine(rocket);         // Prints 🚀
Console.WriteLine(japanFlag);      // Prints 🇯🇵
```

giduac:
Hi @neumatho,

Thanks for posting.

If you switch `UseCompatibleTextRendering` on the WinForms label you can see that when set to `true` the label does not interpret the unicode character correctly.

The Kryptonlabel only uses` Graphics.DrawString()`, which is needed to display the label vertically (`TextRenderer.DrawText()` cannot rotate text) . 

It might well be that Segoe does not have your character onboard. If it does it is a WinForms problem.

In this context we cannot not fix this given the vertical display functionality.

![Image](https://github.com/user-attachments/assets/cabc0dbc-b30e-49f0-96e3-e9e6493c0043)

PWagner1:
Hi @neumatho & @giduac 

This one sounds like a feature request for EXT to me

cc. @Smurf-IV

neumatho:
[...]

PR #2327, opened 2025-07-20
2220-V100-KryptonLabel-does-not-support-surrogates

giduac:
[Issue 2220-KryptonLabel-does-not-support-surrogates](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2220)
- Enables unicode surrogates in horizontal mode
- And the changelog

compile-results

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonLabel.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonLabel.cs
@@ -92,6 +92,37 @@ public class KryptonLabel : VisualSimpleBase, IContentValues
 
     #region Public
     /// <summary>
+    /// Enables the display of Unicode Surrogates, but only when the Orientation is set to Top.
+    /// </summary>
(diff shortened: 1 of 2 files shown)


giduac:
Going in a different direction afterwards.

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonLabel.cs
```diff
@@ -493,6 +524,39 @@ protected override bool EvalTransparentPaint() =>
         // Always need to draw the background because always transparent
         true;
 
+    /// <inheritdoc/>
+    protected override void OnPaint(PaintEventArgs? e)
```
Smurf-IV, line 528: Why could this not be done in the place that already renders the text for labels?
Becaise now if there are a lot of labels on a form, then they will be drawn twice.

giduac: PR is closed and will not be applied.

Issue #2328, opened 2025-07-20
[Other Issues]:  Set the baseline support to .NET Framework 4.7.2


PWagner1:
In version 100, the baseline support should be set to .NET Framework 4.7.2. The rationale behind this is due to the complexity of implementation of newer features in older framework versions.

PR #2332, opened 2025-07-21
* Implement #2328

PWagner1:
* Implement #2328 

image


Smurf-IV:
The `Readme.md` has not been updated with the breaking change.
Also the code for detecting lower than 4.7 needs to be removed, as part of this PR please (Or at least reference the Work Item number that will!)

PWagner1:
> Also the code for detecting lower than 4.7 needs to be removed

@Smurf-IV 

Just done a scan for this, there is no code that uses `#if NETFRAMEWORK471_OR_LOWER`, but there are several instances of `#if NETCOREAPP3_0_OR_GREATER` & ` #if NET6_0_OR_GREATER`. Should I move these to use `#if NET8_0_OR_GREATER`?

Smurf-IV:
`NETFRAMEWORK471_OR_LOWER` will not be used.
But there is defo code for detecting TFM4.6.x as I added it.
As for the Netcore6 or greater, then that should have been done when that TFM was "Dropped"
So all of these need to be in a separate WI and then PR's

PWagner1:
> `NETFRAMEWORK471_OR_LOWER` will not be used. But there is defo code for detecting TFM4.6.x as I added it. As for the Netcore6 or greater, then that should have been done when that TFM was "Dropped" So all of these need to be in a separate WI and then PR's

@Smurf-IV 

Got it! (I forgot what the directive was) I'll update the others in a separate PR.

PR #2340, opened 2025-07-24
2341-V100-fix-prevent-exception-in-ContentFontForButtonForm

tobitege:
`RenderStandard` class, exception in method `ContentFontForButtonForm`.
Not easily reproducable, could be timing issue during a form close.

Actual fix is in lines 5723-5724

Is this worthwhile applying?

Fixes https://github.com/Krypton-Suite/Standard-Toolkit/issues/2341

---

image

---

* A.I. explanation

What happens:

1. You are in `Krypton.Toolkit.Rendering.RenderStandard.ContentFontForButtonForm(...)`.
2. The method decides the normal header font is too tall for the caption bar and wants to create a smaller one:  

   ```c#
   var point = 72 / context.Graphics.DpiY * (fontSpace / 1.333f);
   font = new Font(font.FontFamily, point, font.Style);  // ← throws
   ```
3. `new Font(…)` throws `System.ArgumentException ("Ungültiger Parameter.")`.  
   GDI+ raises that exception when the requested `emSize` is invalid:
   • ≤ 0,  
   • `NaN` / `Infinity`,  
   • or larger than the GDI+ limit (≈ 2048 pt).

Why it fails here:

`point` is computed with a division by `context.Graphics.DpiY`.  
If `DpiY` is 0 (or extremely small) the result becomes `Infinity` (or > 2048) and the
constructor fails.  
`DpiY == 0` can occur when the `Graphics` object you get from
`ViewLayoutContext.Graphics` has already been disposed or was never initialised
properly.

How to confirm:

• While stopped in the debugger, add a watch on  
  `context.Graphics.DpiY` and on the calculated `point`.  
  You will see it is 0 → `Infinity` (or some enormous value).

Suggested fix:

Defensively clamp the value before creating the font:

```c#
float dpiY = context.Graphics.DpiY;
if (dpiY <= 0f) dpiY = 96f;                     // fallback to a sane default
float point = 72f / dpiY * (fontSpace / 1.333f);

if (!float.IsNaN(point) &&
    !float.IsInfinity(point) &&
    point > 3f && point <= 2048f)
{
    font = new Font(font.FontFamily, point, font.Style);
}
```

(or simply short-circuit when `dpiY <= 0`).

That prevents the invalid‐parameter condition while still shrinking the font when feasible.

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,7 @@
 ====
 
 ## 2025-11-xx - Build 2511 (V10 - alpha) - November 2025
+* Resolved [#2341](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2341), Fix exception in `RenderStandard.ContentFontForButtonForm` during teardown
 * Implemented [#2328](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2328), Set the baseline support to .NET Framework 4.7.2
 	- **Note:** This is a breaking change, as the minimum supported version of .NET Framework has been raised from 4.6.2 to 4.7.2.
 * Resolved [#2329](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2329), `AccurateText.StringFormatToFlags()` performs incorrect conversion to TextFormatFlags.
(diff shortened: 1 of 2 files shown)


giduac:
@tobitege 

Changelog...
Is this also applicable to v85/95?

tobitege:
> @tobitege
> 
> Changelog... Is this also applicable to v85/95?

Both V85 and V95 have the same identical method, so technically, yes.

PWagner1 (approved):

Smurf-IV:
> > @tobitege
> > Changelog... Is this also applicable to v85/95?
> 
> Both V85 and V95 have the same identical method, so technically, yes.

If not already, then create PR's for them as well please ?

PR #2370, opened 2025-07-29
66-V100-Fix-KryptonRibbonGroupContainerCollectionEditor-for.NET5-7

tobitege:
.NET 5 to 7 intermittently used `NewItemTypes` in the CollectionEditor instead of `CreateNewItemTypes`,
which caused it at design time to reach down to the abstract class.

Needs closer testing!

### DIFF
--- a/Source/Krypton Components/Krypton.Ribbon/Designers/Collection Editors/KryptonRibbonGroupContainerCollectionEditor.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/Designers/Collection Editors/KryptonRibbonGroupContainerCollectionEditor.cs
@@ -1,12 +1,12 @@
-﻿#region BSD License
+#region BSD License
 /*
- * 
+ *
  * Original BSD 3-Clause License (https://github.com/ComponentFactory/Krypton/blob/master/LICENSE)
  *  © Component Factory Pty Ltd, 2006 - 2016, All rights reserved.
- * 
+ *
  *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
- *  Modifications by Peter Wagner (aka Wagnerp), Simon Coghlan (aka Smurf-IV), Giduac & Ahmed Abdelhameed et al. 2017 - 2025. All rights reserved.
- *  
+ *  Modifications by Peter Wagner (aka Wagnerp), Simon Coghlan (aka Smurf-IV), Giduac, Ahmed Abdelhameed, tobitege et al. 2017 - 2025. All rights reserved.
+ *
  */
 #endregion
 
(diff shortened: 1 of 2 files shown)


tobitege:
Set to draft as I need to test it more (and resolve conflicts)

Smurf-IV:
> Set to draft as I need to test it more (and resolve conflicts)

And also test with e a.Net4.x **ONLY** project please, and this needs to be retrofitted down to 85.LTS

tobitege:
I've tried adding new ribbon groups at designtime with _all available versions_ and cannot reproduce the error anymore:
{6C42D8B0-6C50-4A48-8BCD-E37981CA0D93}

giduac:
@tobitege 

Gave this a test and changed TFMs in testform.csproj to:
```xml
<TargetFrameworks>net8.0-windows;net9.0-windows;net10.0-windows</TargetFrameworks>
```
And rebuild TestForm
Added a ribbon and went through the collections to add a Triple.

Is this the correct way to do it or are NetFramework TFMs needed?

image

tobitege:
Thanks @giduac - I'll research this further!

giduac:
> Thanks @giduac - I'll research this further!

This has to do with the Winforms Out Of Process designer.
Our long standing history on this:

https://github.com/Krypton-Suite/Standard-Toolkit/discussions/684

tobitege:
hmm... worked for me... did you restart Visual Studio, too?
{DF6526FB-757A-4B1A-9803-748F32427086}

Source/Krypton Components/Krypton.Ribbon/Designers/Collection Editors/KryptonRibbonGroupContainerCollectionEditor.cs
```diff
@@ -1,37 +1,46 @@
...
+    #endregion
+
     /// <summary>
     /// Initialize a new instance of the KryptonRibbonGroupTopCollectionEditor class.
     /// </summary>
     public KryptonRibbonGroupContainerCollectionEditor()
         : base(typeof(KryptonRibbonGroupContainerCollection))
     {
+#if NET5_0_OR_GREATER && !NET8_0_OR_GREATER
+        // .NET 5–7 designer relies on NewItemTypes; property has a setter in those versions.
```
PWagner1: .NET 5 - 7 is no longer supported in V100

tobitege: Yes, but if/when this is patched down, who will remember this constellation?
The whole condition construct explains itself and does not break anything.

giduac:
> hmm... worked for me... did you restart Visual Studio, too? {DF6526FB-757A-4B1A-9803-748F32427086}

Edit the TFM line in csproj.
Switch to another repo and back.
And build

tobitege:
Yeh, was able to reproduce the error now, working on it again.

tobitege:
The OOP is making it impossible to avoid such exception as in the issue.
The only way seems to be to enforce the legacy WinForm's designer, which works fine with those collections.
I'll leave it to the other maintainers, if this change to the `Directory.Build.props` file (which does not get shipped with NuGet builds) is worth keeping.
The `KryptonRibbonGroupContainerCollectionEditor.cs` is reverted back to its original code.

**Further info from A.I. that might be of interest:**

Your `Directory.Build.props` lives in *your* repo and only affects MSBuild when it’s sitting next to the projects you’re building. When you pack your DLLs and push them to NuGet, that props file doesn’t go along for the ride—NuGet packages don’t automatically carry your root `Directory.Build.props` into consuming solutions.  

So:

- **In-repo**: Every project under your repo root (Ribbon, Toolkit, TestForm, etc.) will see the `<UseWindowsFormsDesigner>false/>` switch when you build locally.  
- **NuGet consumers**: Their own apps (which only install your compiled DLLs) will *not* import your props file unless you explicitly ship a `.props` file in the package’s `build\` folder.  

If you *do* need consumers of your NuGet packages to force the legacy designer, you have two paths:

1.  Embed a `.props` in your NuGet (e.g. `build\Krypton.Ribbon.props`) that sets `<UseWindowsFormsDesigner>false</UseWindowsFormsDesigner>` under the same `Condition="'$(UseWindowsForms)'=='true'"`.  MSBuild will auto-import that when they install your package.  
2.  Document in your README that any WinForms app using your toolkit should add  
    ```xml
    <PropertyGroup>
      <UseWindowsFormsDesigner>false</UseWindowsFormsDesigner>
    </PropertyGroup>
    ```  
    to their own project (or root `Directory.Build.props`).  

But as it stands today, your props file only governs *your* builds—not end-user apps.

giduac:
@tobitege 

What is the exact effect of this:
```xml
<PropertyGroup>
  <UseWindowsFormsDesigner>false</UseWindowsFormsDesigner>
</PropertyGroup>
```

tobitege:
Explanation I got below
---
In VS 2022’s .NET 8+ Preview designer your custom `CreateCollectionForm` (and all other UITypeEditor overrides) never actually run.  Under the covers the out-of-proc DesignToolsServer spins up its own built-in `DefaultCollectionForm` and simply ignores any `CreateCollectionForm` or toolbar plucking you do in your editor subclass.  

That is why you still see the same dialog (and still get the abstract-class exception) even though your code is in place.

Unfortunately there is no API hook in the new designer to inject per-type buttons (or swap out the form).  Your options today are:

  * Turn off the preview (out-of-proc) WinForms designer and go back to the legacy in-process designer.  In VS 2022 you can toggle this under  
    **Tools → Options → Windows Forms Designer → Use legacy designer for .NET Core and later projects**  
    (or set `<UseWindowsFormsDesigner>false</UseWindowsFormsDesigner>` in your project file).  The legacy designer will honor your `CreateCollectionForm` override exactly as you’ve written it.

  * Build a full VSIX extension against the new designer framework to provide your own collection-editor UI.  (This is a non-trivial undertaking.)

  * File a feature request/bug with the WinForms team asking for extension points in the new designer to allow custom collection-editor forms or per-type toolbar buttons.

For the moment the quickest way to get your multi-button UI is to fall back to the legacy designer.

tobitege:
Closing. Not a solution in combination with the out of process designer.

PR #2371, opened 2025-07-29
* Resolves #2103

### LINKED ISSUE #2103: [Bug/Feature Request]: Ensure that `KryptonForm` properly supports RTL/LTR
Since it's inception, the toolkit has never properly supported RTL. That means that some developers are at a disadvantage.

Have you tried toggling the RightToLeftLayout property though?

Winforms behaviour is as follows:
RTL on / RTL Layout off : Title remains the same
RTL on / RTL Layout on: Title bar and buttons are switched.
RTL off / RTL LAyout on: nothing happens / everything remains the same

This property isn't visible anymore in the editor nor properties window / designer.
And not serialised to to the designer source.
And no default value

Possible change...
```cs
    /// <summary>
    /// Gets and sets the RightToLeft property.
    /// </summary>
    [Browsable(true)]
    [DefaultValue(RightToLeft.No)]
    [EditorBrowsable(EditorBrowsableState.Always)]
    [DesignerSerializationVisibility(DesignerSerializationVisibility.Visible)]
    public override RightToLeft RightToLeft
    {
```

PWagner1:
* Resolves #2103 (V100 only)

image

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,7 @@
 ====
 
 ## 2025-11-xx - Build 2511 (V10 - alpha) - November 2025
+* Resolved [#2103](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2103), Ensure that `KryptonForm` properly supports RTL/LTR
 * Implemented [#2365](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2365), Fix exception in `KryptonCustomPaletteBase.GetPaletteBackGridHeaderColumn` for `BoldedOverride` state.
 * Implemented [#2349](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2349), Fix `KryptonListBox` shifting on visible item selection.
 * Implemented [#2354](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2354), `KryptonDataGridView.DoubleBuffered` property added.
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
@@ -2243,7 +2243,7 @@ public class KryptonForm : VisualForm,
             Rectangle availableArea = context.DisplayRectangle;
             
             // Check if RTL layout is enabled
-            bool isRTL = context.Control?.RightToLeftLayout == true;
+            bool isRTL = (context.Control as Form)?.RightToLeftLayout == true;
             
             // Get the title alignment from the form
             var kryptonForm = context.Control as KryptonForm;
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
@@ -161,7 +161,10 @@ public class KryptonForm : VisualForm,
 
         // Content draws the text and icon inside the title bar
         _drawContent = new ViewDrawContent(StateActive.Header.Content, this, VisualOrientation.Top);
-        _drawHeading.Add(_drawContent, ViewDockStyle.Fill);
+        
+        // Create RTL-aware layout for title bar content
+        var titleBarLayout = new RTLTitleBarLayout(_drawContent);
+        _drawHeading.Add(titleBarLayout, ViewDockStyle.Fill);
 
         // Create a decorator so that the heading has a fixed sized and not based on content
         _headingFixedSize = new ViewDecoratorFixedSize(_drawHeading, Size.Empty);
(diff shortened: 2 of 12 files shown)


tobitege:
Great job! I'll take a look after work soon-ish.

PWagner1:
> Great job! I'll take a look after work soon-ish.

It was irritating me for years, will do the ribbon next, followed by other controls in separate PRs/issues.

tobitege:
_Just for fun, I had to ask o3 about RTL a little, didn't know anything about it:_ 🤣 

The short answer is: most of the time you do **not** have to decide at all—Windows’ bidirectional (-BiDi) engine will lay out mixed Arabic + Latin text correctly inside an ordinary LTR control.  
Only when you want the *editing experience* (caret movement, selection, arrow keys) or alignment of the whole string to feel natural to an Arabic/Hebrew user do you need to switch `RightToLeft` to `Yes`.  
If you want to do that automatically you can use a simple “first-strong-character” heuristic.

────────────────────────────────────────────────────────
1. What happens if you do nothing?
────────────────────────────────────────────────────────
`TextRenderer`/GDI+ (and all modern text stacks) implement the Unicode BiDi algorithm.  
Given a string like:

    "محمد Ali <ali@example.com>"

they will:

• Detect the first strong directional character (`م`, an RTL letter)  
• Treat the subsequent Arabic run as RTL, then switch back for “Ali ”, etc.  

So the user sees naturally ordered glyphs even though the control’s `RightToLeft` property is `No`.  
Most mail clients work exactly this way: the *message list* stays left-to-right, but individual sender names render in RTL where needed.

────────────────────────────────────────────────────────
2. When would you flip `RightToLeft` automatically?
────────────────────────────────────────────────────────
• Text boxes where the user is expected to *edit* Arabic/Hebrew.  
• Labels that should be right-aligned when the text is Arabic but left-aligned otherwise.  
• Search boxes, address bars, or chat bubbles that feel awkward if caret direction is opposite to the script in use.

────────────────────────────────────────────────────────
3. Simple heuristic: “first strong character”
────────────────────────────────────────────────────────
Unicode designates every code point as L, R, AL, EN, etc. (Left-to-Right, Right-to-Left, Arabic-Letter, European Number…).  
A common rule—also used by browsers—is:

    If the first *strong* character in the string is RTL → treat the whole
    paragraph as RTL; otherwise treat it as LTR.

In .NET you can implement that with `CharUnicodeInfo`:

```csharp
using System.Globalization;
using System.Text;

public static class BidiHelper
{
    public static bool IsRtl(string s)
    {
        if (string.IsNullOrEmpty(s))
            return false;

        foreach (var ch in StringInfo.GetTextElementEnumerator(s).Cast<string>().Select(e => e[0]))
        {
            var dir = CharUnicodeInfo.GetBidiCategory(ch);
            if (dir is BidiCategory.LeftToRight or BidiCategory.LeftToRightEmbedding or BidiCategory.LeftToRightOverride)
                return false;

            if (dir is BidiCategory.RightToLeft or BidiCategory.RightToLeftArabic
                       or BidiCategory.RightToLeftEmbedding or BidiCategory.RightToLeftOverride)
                return true;
        }
        return false;
    }
}
```

Usage for a label or textbox:

```csharp
var rtl = BidiHelper.IsRtl(text);
myLabel.RightToLeft = rtl ? RightToLeft.Yes : RightToLeft.No;
myLabel.TextAlign    = rtl ? ContentAlignment.MiddleRight : ContentAlignment.MiddleLeft;
```

Alternatively call the Win32 API `GetStringTypeW` with `CT_CTYPE2` and test `C2_RIGHTTOLEFT`.

────────────────────────────────────────────────────────
4. What about partial strings or mixed content?
────────────────────────────────────────────────────────
If you encounter corner cases such as:

• “Re: محمد Ali – Meeting notes”  
• Mixed phone numbers, punctuation, etc.

let Windows handle it: keep `RightToLeft = No` (so layout stays LTR) but, if needed, insert the Unicode formatting marks `RLM`, `LRM`, `RLE/PDF` to nudge the algorithm. That is what Outlook and Gmail do internally.

────────────────────────────────────────────────────────
Summary
────────────────────────────────────────────────────────
• For a mail viewer that occasionally shows Arabic names, leave the control/layout LTR; the BiDi algorithm renders the text correctly.  
• Switch `RightToLeft = Yes` only for controls where *user interaction* should follow RTL conventions, or when almost all content is RTL.  
• An easy auto-toggle is to inspect the first strong character and set `RightToLeft` accordingly; users from RTL cultures generally expect that behaviour.

tobitege:
The form itself seems to do what it is expected of, nice!

@PWagner1 you should add that toggle to the `Controls Test` dialog, maybe?

giduac:
The Kform title bar behaves differently from the standard winforms form when only RTL is on.
On the Winforms form when RTL is on and RTL layout is off the title bar does not change.

https://github.com/user-attachments/assets/9941ae08-8dc7-4609-9328-45cd04dae67c

PWagner1:
> The Kform title bar behaves differently from the standard winforms form when only RTL is on. On the Winforms form when RTL is on and RTL layout is off the title bar does not change.
> 
>  devenv_6Bexi2iagq.mp4

Have you tried toggling the `RightToLeftLayout` property though?

Winforms behaviour is as follows:
RTL on / RTL Layout off : Title remains the same 
RTL on / RTL Layout on: Title bar and buttons are switched.
RTL off / RTL LAyout on: nothing happens / everything remains the same

PWagner1:
Going to close and reopen a new PR... messed up somewhere @giduac please can you copy your findings to #2103?

PR #2387, opened 2025-08-04
Fix exceptions in KryptonTextBox and KryptonMaskedTextBox when disabl…

### LINKED ISSUE #2386: [Bug]: KTextBox paint exceptions (V100)
**Describe the bug**
`KryptonMaskedTextBox` and `KryptonTextBox` each cause an internal paint exception during them toggling for disabled components.

**To Reproduce**
1. Open up `TestForm`
2. Click on 'Controls Test'
3. With VS Studio stop-on-exception enabled toggle between MS365 themes
4. See below error

**Expected behavior**
No (internal) exceptions for disabled components.

Fix is rather simple: the `catch` already does the right thing to prevent the `Font` being invalid.
It is by order of magnitudes faster to just use the inline font check than try/catch and do a 2nd paint on every toggle (already tested with incoming PR).

PR incoming.

**Screenshots**

Image

**Desktop (please complete the following information):**
 - OS: Windows 11 Pro x64, current updates
 - Version: most current
 - Framework/.NET Version: at least 4.7.2
 - Toolkit Version: current alpha branch

tobitege:
Fixes #2386 

We **can** eliminate the first-chance exception completely with a small, targeted change.

Why the exception happens  
• Immediately after a theme swap, the internal `Font` still points to the *old* palette font object, which has just been disposed.  
• The first call to `Graphics.DrawString` therefore throws.  
• A single `OnNeedPaint` later the control updates its `Font` to the fresh palette font, so everything continues to work – but the debugger has already stopped on that first-chance exception.

A surgical fix  
Instead of asking the component for its `Font` during the disabled-paint path, ask the **palette** directly; it will hand back the new font object even during the transition frame.

```csharp
// Define the font to use for disabled painting – always query the palette first.
var disabledFont = _kryptonMaskedTextBox
    .GetTripleState()
    .PaletteContent?
    .GetContentShortTextFont(PaletteState.Disabled)
    ?? Font;  // Fallback: current Font if palette returns null

// Draw using a solid brush
using var foreBrush = new SolidBrush(ForeColor);
g.DrawString(drawText, disabledFont,
             new RectangleF(rect.left, rect.top,
                            rect.right - rect.left, rect.bottom - rect.top),
             stringFormat);
```

What changes:

1. We **remove the try/catch** block altogether – nothing throws because `disabledFont`
   is always a valid, live object.
2. No performance hit: calling the palette getter is fast (it just returns a cached `Font`).
3. No more debugger noise when switching Microsoft 365 themes – or any future palette
   that swaps fonts at runtime.

Example exception (if stop on error for that type is enabled):
{13B9543D-2AFC-49C9-B2E2-22B7F4742772}

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,7 @@
 ====
 
 ## 2025-11-xx - Build 2511 (V10 - alpha) - November 2025
+* Implemented [#2384](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2384), Fix exceptions in `KryptonTextBox`, `KryptonMaskedTextBox` when disabled
 * Implemented [#2368](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2368), Fix exception in `PaletteRedirectGrids.GetInheritBack` due to missing `BoldedOverride`.
 * Implemented [#2365](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2365), Fix exception in `KryptonCustomPaletteBase.GetPaletteBackGridHeaderColumn` for `BoldedOverride` state.
 * Implemented [#2349](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2349), Fix `KryptonListBox` shifting on visible item selection.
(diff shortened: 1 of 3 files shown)


giduac (approved):

giduac:
@tobitege 

Conflicts....

tobitege:
Will take care of them asap :)

Issue #2397, opened 2025-08-09
[Bug]: KForm Parent control is null


KamaniAR:
KForm Line 1962 -> control = control.Parent ?? throw new InvalidOperationException("Parent control is null.");

Error when press Ctrl + F1 on Form.
Checked in test Project.

V100 / V95
V85 not affected.

tobitege:
Which branch? Alpha?

KamaniAR:
Yes

giduac:
Checked and confirmed.

PR #2398, opened 2025-08-09
2397-V100-Kform-Parent-control-is-null

giduac:
- Issue #2397 
- Return statement adjusted
- And the changelog

compile-results


tobitege (approved):
LGTM

PWagner1 (approved):

PR #2399, opened 2025-08-09
2397-V95-Kform-Parent-control-is-null

giduac:
- Issue #2397
- Return statement adjusted
- Warning removed: Krypton.Toolkit\Controls Visuals\VisualForm.cs(732,50): warning CS0168: The variable 'ioEx' is declared but never used
- And the changelog

compile-results


tobitege (approved):
LGTM

Issue #662, opened 2022-03-01
[Bug]: Can not "Control / Set Disabled" Background Colour Of Comboboxes, in `DropDown` mode


MattH-Work:
![image](https://user-images.githubusercontent.com/84494183/156185082-7d45f7a8-40ba-4d24-9594-e383bec3969d.png)

All of the comboxes in the attached image are using the DropDownStyle "DropDownList" and behave as expected
I am overriding the Back colour for both StateCommon & StateDisabled to be LightYellow 

The Description combobox is using DropDownStyle "DropDown" and as you can see has the LightYellow backround (again I am overriding  both StateCommon & StateDisabled)  

BUT...

Where has the Grey box come from?
Is this a bug, or do I need to override soemthing else?

I find the same issue with any combobox using DropDownStyle "DropDown"

Note, the Groupbox that hosts all the controls is currently disabled (Enabled = False), when enabled ALL comboxes display as expected
![image](https://user-images.githubusercontent.com/84494183/156186184-bb6215bc-c8f6-4056-8d02-fe66b347fffc.png)

Cheers

MattH

Smurf-IV:
Trying to remember of the top of my head, but when in "DropDown" then there is an edit component, therefore it will also need to be overridden.
If you can wait, I can check at the weekend.

MattH-Work:
I can wait

Thanks for the help

Smurf-IV:
Grey box is also being applied to the KRTF in the disabled state:
![image](https://user-images.githubusercontent.com/2418812/156877122-44037198-ec30-47ff-b3ed-f9d2895ece1d.png)

Smurf-IV:
Hmm. I wonder..
https://stackoverflow.com/questions/1946139/how-to-change-the-background-color-of-a-rich-text-box-when-it-is-disabled

Smurf-IV:
It gets worse..
![image](https://user-images.githubusercontent.com/2418812/156916180-3d406396-18c9-4da2-bda8-bc7889475643.png)

Wagnerp:
@Smurf-IV This also applies to the `KryptonRichTextBox`

![image](https://user-images.githubusercontent.com/949607/161314140-b7c5f742-ed7e-4d21-9fcf-865f9e31b551.png)

Smurf-IV:
> @Smurf-IV This also applies to the `KryptonRichTextBox`

Already displayed in the "green circles"

Smurf-IV:
[...]

PR #2403, opened 2025-08-10
2048 V100 Fix KryptonComboBox event handling

### LINKED ISSUE #2048: `KryptonStatusStrip` Issues
image
on .net9 even the StatusStrip turns blue, and one can not simply solve this by setting the BackColor, what's wrong？
In before only menustrip and toolstrip got blue, but now even worse?

_Originally posted by @xiaoyuvax in https://github.com/Krypton-Suite/Standard-Toolkit/issues/1979#issuecomment-2579805873_

tobitege:
Fixes #2048 

- Fix for `KryptonComboBox` event `OnEnabledChanged`:  call the base method only when the control is enabled.
- `TestForm`: enhanced `GroupBoxTest` (button "GroupBox") to include new combo box controls and a button to toggle their enabled state.

### Root cause

- Suppressing `InternalComboBox.OnEnabledChanged` likely prevents the underlying WinForms `ComboBox` from refreshing its internal enabled state and input routing when transitioning from disabled (especially when disabled at design-time prior to handle creation) to enabled at runtime.  
- The outer control repaints correctly, but the hosted native combo’s internal logic never runs its enable/disable transition code because `base.OnEnabledChanged(e)` is not called in `InternalComboBox`.

### Solutions
  - Only forward when transitioning to enabled:
  
    ```csharp
    protected override void OnEnabledChanged(EventArgs e)
    {
        if (Enabled)
        {
            base.OnEnabledChanged(e);
        }
    }
    ```
  - This preserves the original intent of not forwarding on disable (for #662), while still letting the native `ComboBox` re-arm itself when re-enabled. In practice, the simple “always call base” above is sufficient.

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,7 @@
 ====
 
 ## 2025-11-xx - Build 2511 (V10 - alpha) - November 2025
+* Resolved [#2401](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2401), Fix `KryptonComboBox` event `OnEnabledChanged`.
 * Resolved [#2397](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2397), `KryptonForm` throws and exception on CTRL+F1.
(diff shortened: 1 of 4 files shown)


giduac (approved):
Tests positive.

PR #2413, opened 2025-08-12
2406 V100 KryptonProgressBar enhancements

tobitege:
### PR description
- Fix KryptonProgressBar value fill rendering to match palette 3D style
  - Progress value now uses `PaletteColorStyle.GlassNormalFull` by default (instead of flat fill), driven by renderer and theme.
  - Added `ValueBackColorStyle` property to allow choosing any `PaletteColorStyle` for the value area.

- Correct and unify style rendering
  - Reworked painting into a single `switch (Style)` with clear branches for `Marquee`, `Blocks`, and `Continuous`.
  - Fixed `Marquee` computation to use pixel-per-unit mapping, ensuring correct band size and position in all orientations and ranges.
  - Implemented proper `Blocks` style:
    - Draws discrete segments with theme 3D fill.
    - Supports RTL and vertical orientations.
    - Renders a partial block for fractional progress (no “jump” between blocks).

- New customization options
  - `BlockCount` property: exact number of blocks; when `0` (default), block count derives from the range as one block per 10 units (e.g., 10 blocks for 0–100) for intuitive percentage reading.
  - Text readability:
    - `ShowTextShadow` and `TextShadowColor` (default shadow uses palette text color darkened; subtle and tight to glyphs).
    - `ShowTextBackdrop` and `TextBackdropColor` (rounded semi-transparent backdrop behind text, centered and sized to measured content).

- Misc
  - Kept palette-driven colors; rendering paths leverage `RenderStandardBack`/`RenderStandardContent`.
  - Maintained clipping for rounded borders and respecced RTL/orientation behavior.
  - Removed nullability warning by asserting non-null `PaletteContent` in the paint path.

- `TestForm` -> `ProgressBar` demo shows all new options

image

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,7 +3,7 @@
 ====
 
 ## 2025-11-xx - Build 2511 (V10 - alpha) - November 2025
-* Resolved [#2125](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2125), Refactored `GetToastNotificationIconType`.
+* Resolved [#2406](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2406), Enhance `KryptonProgressBar` options and painting.
 * Implemented [#2048](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2048), `KryptonStatusStrip` control theming.
 * Implemented [#2291](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2291), Add tooltips to `KryptonPictureBox`.
 * Resolved [#2401](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2401), Fix `KryptonComboBox` event `OnEnabledChanged`.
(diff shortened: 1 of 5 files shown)


tobitege:
{7AE333E0-68B2-47CF-85D9-3F986B86F4D8}

tobitege:
Demo page expanded with vertical progress bars:

image

tobitege:
This is actually easy to test, demo page shows all the new features...

Smurf-IV:
Why does it look like the text is drawn twice with a pixel offset ?

tobitege:
> Why does it look like the text is drawn twice with a pixel offset ?

I see that it is more prominent in the vertical bars. It is meant for better readability because the "normal" paint has the text pretty much unreadable, tbh. I can fix the extra offset in the vertical bars, which seems to leave a 1 px "gap".

tobitege:
> Why does it look like the text is drawn twice with a pixel offset ?

I've tried now several attempts to get rid of that tiny offset in the vertical text, but nothing worked.
Since the backdrop and shadow text are optional, I made those default to false now and not put any more time into it now.

Issue #2329, opened 2025-07-20
[Bug]: AccurateText.StringFormatToFlags handles conversion incorrect


giduac:
**Describe the bug**
AccurateText.StringFormatToFlags handles conversion incorrect.
It uses bitwise AND instead of OR

**Desktop (please complete the following information):**
V100 for now since nobody ever complained ...

Smurf-IV:
@giduac If this exists in V85 - Can this also be fixed in onwards ?

giduac:
> [@giduac](https://github.com/giduac) If this exists in V85 - Can this also be fixed in onwards ?

@Smurf-IV 
It seems that nobody ever used it.
They should've found out since it malfunctions....

But I'll add to 85/95 as well.

PR #2425, opened 2025-08-16
2424 v100 NoClipping regression fix

### LINKED ISSUE #2424: [Bug]: V100 RenderStandardContent ignores displayRect; text overflows due to forced NoClipping in AccurateText
Regression since #2329 

**Describe the bug**
Text drawn by Krypton renderer overflowed outside the requested displayRect.
Root cause was `AccurateText.DrawString` forcing `TextFormatFlags.NoClipping`, which bypassed the renderer’s clipping intent.

Impact: Headers, DataGridView headers/cells, and custom controls could not constrain text.
Resizing or constrained regions showed obvious bleed.

Fix:
Stop forcing `NoClipping` in `AccurateText.DrawString` for horizontal text.
Keep `DrawContent` using its intersected displayRect; no extra Graphics clipping required after fix.

tobitege:
Fixes #2424 (regression since #2329 )

Fix renderer clipping: respect displayRect by removing NoClipping in AccurateText.DrawString

Details:
AccurateText: remove TextFormatFlags.NoClipping and add a comment preventing reintroduction.

RenderStandard: retain intersected rectangles; remove extra Graphics.SetClip code.

Rationale: 
The renderer’s API contract expects draw calls to be clipped to displayRect. Forcing NoClipping at the text layer violated this and caused visible overflow.

Secondary change:
Made the ellipsis conditional - it is not right to either remove it completely, nor apply it always. 
It should be applied when actual clipping would happen.

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/AccurateText/AccurateText.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/AccurateText/AccurateText.cs
@@ -304,7 +304,9 @@ public class AccurateText : GlobalId
                         var tff = StringFormatToFlags(memento.Format);
 
                         // Conditional ellipsis: only allow if text would overflow the target rect
-                        var ellipsisFlags = TextFormatFlags.EndEllipsis | TextFormatFlags.WordEllipsis | TextFormatFlags.PathEllipsis;
+                        const TextFormatFlags ellipsisFlags = TextFormatFlags.EndEllipsis |
+                                                              TextFormatFlags.WordEllipsis |
+                                                              TextFormatFlags.PathEllipsis;
                         if (((tff & ellipsisFlags) != 0) && (memento.Size.Width <= rect.Width))
                         {
                             tff &= ~ellipsisFlags;


giduac (approved):

tobitege:
@PWagner1 @Smurf-IV @giduac - what do you guys think about the change for the Ellipsis?
As an example, take our TestForm's start screen - If ellipsis flag was always set, all buttons would get their captions with an ellipsis regardless if it was actually necessary.

Now, it only applies it, if the text actually gets clipped, see bottom button:

{F8A38813-912B-4E07-8357-34EEFB044A86}

That to me looks more "natural". But not sure, if this change in behavior is wanted or needs further conditions?

PWagner1 (approved):

PWagner1:
> @PWagner1 @Smurf-IV @giduac - what do you guys think about the change for the Ellipsis? As an example, take our TestForm's start screen - If ellipsis flag was always set, all buttons would get their captions with an ellipsis regardless if it was actually necessary.
> 
> Now, it only applies it, if the text actually gets clipped, see bottom button:
> 
> {F8A38813-912B-4E07-8357-34EEFB044A86}
> That to me looks more "natural". But not sure, if this change in behavior is wanted or needs further conditions?

I like it

Issue #2433, opened 2025-08-17
[Other Issues]: KMB Native Window, replace constants with PI values


giduac:
Replace the defined constants with those from PInvoke.

PR #2434, opened 2025-08-17
2433-V100-replace-constants

giduac:
- Issue: #2433 
- Remove constants.
- Internal change, so no change log.

compile-results


tobitege (approved):
LGTM

Issue #984, opened 2023-03-25
[Bug]: `SizeGripStyle` is only ever assigned to in the Krypton Toolkit


Smurf-IV:
**Describe the bug**
i.e the value that is assigned iis never checked or used in any code in the Krypton Toolkit.

**To Reproduce**
Perform a search on the `Krypton Toolkit` source for `SizeGripStyle` and see if it shows up in any code to perform a draw modification !

**Reason**
I need it to be shown in the File "Open / Save" Dialogs

Wagnerp:
I think something is in the `StatusStrip` renderer. I'd suggest having a look there...

Smurf-IV:
It is _also_ supposed to show up when there is **no** status strip....

giduac:
@PWagner1 & @Smurf-IV 

Which is the status if this ticket?

PWagner1:
@giduac 

Needs to be fixed

tobitege:
### Plan: add a real “grippie” that honors Form.SizeGripStyle

- **Goal**
  - Respect `Form.SizeGripStyle` (Auto/Show/Hide) for `KryptonForm` when using custom chrome.
  - Draw a classic sizing grip glyph in the bottom corner and make it resize-capable via hit-testing.
  - Keep changes surgical to `KryptonForm` only; do not affect StatusStrip’s own sizing grip.

- **Where to implement**
  - File: `Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs`
  - Touch points:
    - `WindowChromePaint(...)` → draw the grip overlay.
    - `WindowChromeHitTest(...)` → map grip hit area to resize hit-tests.
    - Add small helpers inside `KryptonForm`.
    - Optionally override `OnSizeGripStyleChanged` to trigger repaint.

- **When to show the grip**
  - Implement a helper `ShouldShowSizingGrip()`:
    - Return false if:
      - `SizeGripStyle == Hide`
      - Window is not resizable: `FormBorderStyle` not in `Sizable`/`SizableToolWindow`
      - Window state is `Maximized` or `Minimized`
      - A bottom `StatusStrip` is present/visible/merged and its own `SizingGrip` is enabled (to avoid double grippies)
    - Return true if:
      - `SizeGripStyle == Show` and resizable and not maximized/minimized
      - `SizeGripStyle == Auto` and resizable and not maximized/minimized
[...]

PR #2436, opened 2025-08-18
(feat) KryptonForm grippie implementation. Fix memory leak in Admin C…

tobitege:
Implements #984 

### Add real, theme-aware sizing grip to `KryptonForm`

- Implement form-level sizing grip that honors `Form.SizeGripStyle` and resizability
  - New helpers in `KryptonForm`:
    - `ShouldShowSizingGrip()` (Hide/Auto/Show; resizable only; not min/max; suppressed when merged `StatusStrip` with `SizingGrip=true`)
    - `(Rectangle gripRect, bool isRtl) GetGripRectAndRtl()`
    - `DrawSizingGrip(Graphics g, Rectangle gripRect)`
    - `TryDrawResourceGrip(Graphics g, Rectangle dest)` (resource-first grip)
- Paint integration
  - Grip is drawn in `OnNonClientPaint(...)` after base NC paint to overlay cleanly on client content (no dependency on internal panel paints).
  - `WindowChromePaint(...)` remains the main chrome paint; grip rendering is separate and minimal.
- Hit-testing
  - `WindowChromeHitTest(...)`: before border checks, map grip rect to `HT.BOTTOMRIGHT` or `HT.BOTTOMLEFT` (RTL), preserving resize affordance with flat/hidden borders.
- Visuals and theming
  - Resource-first rendering: choose themed BMP from `ResourceFiles.SizeGripStyles.SizeGripStyleResources` (Office 2003/2007/2010 variants) based on theme name or `PaletteMode`.
  - Fix for old BMP transparency: color-key transparency via top-left pixel; crisp rendering using NearestNeighbor for tiny glyphs; light DPI upscaling.
  - Fallback when no resource applies: draw larger, DPI-scaled 2x2 diagonal dots using theme-derived colors from the current palette to ensure contrast in light/dark themes; RTL-aware anchoring.
- RTL support
  - Grip anchor and hit-test return `HT.BOTTOMLEFT` when `RightToLeftLayout` is true.

### Notes and cleanup
- See `FormBorderTest` form (TestForm -> `Form Border`) for change events, that _could_ be necessary for runtime toggling of values, but should be optional.

### TODO Testing
- Verify with `FormBorderStyle = Sizable/SizableToolWindow`, `SizeGripStyle = Show/Auto/Hide`.
- Toggle maximize/minimize; ensure grip hides appropriately.
- With a bottom `StatusStrip` merged into chrome and `SizingGrip=true`, confirm only the strip’s grip is visible.
- Check RTL: grip appears bottom-left and hit-test returns `HT.BOTTOMLEFT`.

---

Examples

{DF84B3E9-7BFB-44AB-A162-8C3C1708CBD1}

{DEA89DC7-9ECD-4C8D-846E-821C63A22FFE}

---

{0D67E86E-3436-42A8-8031-FE352477F5E9}

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,10 @@ Follow the links to see the different objects and layouts that this framework al
 
 There are list of changes that have occurred during the development of the V100.00 version
 
+* Implemented [#984](https://github.com/Krypton-Suite/Standard-Toolkit/issues/984), `KryptonForm` using `SizeGripStyle` for grippie.
+  - **Note:** This contains a breaking change for **custom** themes as it introduces a new abstract method:
+    `public abstract Image? GetSizeGripImage(RightToLeft isRtl);`
+    so that themes can return a size grip image (or null).
 - Implemented [#2328](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2328), Set the baseline support to .NET Framework 4.7.2
   - **Note:** This is a breaking change, as the minimum supported version of .NET Framework has been raised from 4.6.2 to 4.7.2.
 - Resolved [#240](https://github.com/Krypton-Suite/Standard-Toolkit/issues/240), **[Breaking Change]** `KryptonRichTextBox` Why is it not possible to have the `ButtonSpecs` aligned to the top of a control


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
```diff
@@ -228,6 +229,252 @@ public KryptonForm()
 
     private float GetDpiFactor() => DeviceDpi / 96F;
 
+    /// <summary>
+    /// Determines whether the form-level sizing grip should be shown.
+    /// </summary>
+    private bool ShouldShowSizingGrip()
```
Smurf-IV, line 240: Please add a reference to the issue of your excellent explanation of this log

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
```diff
@@ -228,6 +229,252 @@ public KryptonForm()
...
+            else if (themeName.IndexOf("Blue", StringComparison.OrdinalIgnoreCase) >= 0)
+                bmp = ResourceFiles.SizeGripStyles.SizeGripStyleResources.Office2010BlueGripStyle;
+            else if (themeName.IndexOf("Silver", StringComparison.OrdinalIgnoreCase) >= 0)
+                bmp = ResourceFiles.SizeGripStyles.SizeGripStyleResources.Office2010SilverGripStyle;
+        }
+        else
+        {
+            switch (PaletteMode)
+            {
+                case PaletteMode.ProfessionalOffice2003:
```
Smurf-IV: This should be in the resource class and not in here,
i.e. `themeClass.GetGrippieImage`

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
```diff
@@ -228,6 +229,252 @@ public KryptonForm()
...
+        Rectangle target = new Rectangle(dest.Right - w, dest.Bottom - h, w, h);
+
+        // Use top-left pixel as color-key transparency (legacy resource convention)
+        Color key;
+        try { key = bmp.GetPixel(0, 0); } catch { key = Color.Magenta; }
+        using var ia = new System.Drawing.Imaging.ImageAttributes();
+        ia.SetColorKey(key, key);
+
+        var state = g.Save();
+        g.InterpolationMode = System.Drawing.Drawing2D.InterpolationMode.NearestNeighbor;
```
Smurf-IV: There is already a using graphics class that can set and restore these things (Or extend so that it does please?)

tobitege:
@Smurf-IV @PWagner1 @giduac 
In order to move the SizeGrip images into the themes, I've added a new abstract method to `PaletteBase`:
```
public abstract Image? GetSizeGripImage(bool isRtl);
```
The question is:
- where do I mark it in the docs as Breaking Change?
- OR should I just make it `virtual`?

The existing size grip images for some of the older themes will already pull the images from the resources.
Most other palettes will just return `null` and a size grip will be painted instead.

tobitege:
@Smurf-IV I think I got your 3 points all covered now, thanks for the feedback!

giduac:
> @Smurf-IV @PWagner1 @giduac In order to move the SizeGrip images into the themes, I've added a new abstract method to `PaletteBase`:
> 
> ```
> public abstract Image? GetSizeGripImage(bool isRtl);
> ```
> 
> The question is:
> 
> * where do I mark it in the docs as Breaking Change?
> * OR should I just make it `virtual`?
> 
> The existing size grip images for some of the older themes will already pull the images from the resources. Most other palettes will just return `null` and a size grip will be painted instead.

The `RightToLeft` property is an enum called RightToLeft.
Might use that and pass the property down as param in to the theme GetGrippie

tobitege:
> RightToLeft

Done that locally, now still tracking down some timing issues during painting :(

tobitege:
Here's a summary of what we fought with for couple of hours, for anyone interested:

### What was happening
- The form’s paint pipeline renders in this order: non-client (frame) → client (`PerformViewPaint`) → screen. When we drew the grippie during non-client paint, the later client paint erased it. When we tried painting into the window DC without strict clipping, it could overwrite child controls and leave trails during rapid resizes.

### Key fixes and why they work
- Draw after client paint
  - We hook `WM_PAINT` in `WndProc` and call `DrawSizingGripOverlayIfNeeded` after `base.WndProc`. This makes the grippie the last thing painted each frame, so `PerformViewPaint` cannot erase it.

- Draw only inside the grip bounds
  - In `DrawSizingGripOverlayIfNeeded` we obtain a window DC via `GetWindowDC(Handle)` and then limit drawing to the exact grip rectangle using `IntersectClipRect(hDC, newRect.Left, newRect.Top, newRect.Right, newRect.Bottom)`. This guarantees nothing outside the small corner rect can be touched, so no dots ever appear over other controls and no diagonal trails are produced.

- Clear the previous overlay region deterministically
  - We track the last grip area in client coords and invalidate that small region with `RedrawWindow` when the grip moves or is hidden. That lets the normal pipeline repaint it cleanly and removes stale pixels.

- Stop double-drawing and async churn
  - We no longer draw the grippie in `OnNonClientPaint`, and we removed deferred `BeginInvoke`-based non-client recalcs around palette changes for this case. This prevents reordering and flicker introduced by queued invalidations during fast resizing.

- Use correct geometry at all times
  - The grip rectangle is computed from real window metrics, not `ClientSize`, via `RealWindowBorders` and `RealWindowRectangle`. This makes the math stable while resizing and across themes, RTL, and DPI:
    - Size: scaled from `GetDpiFactor`.
    - Position: bottom-right or bottom-left with `RightToLeftLayout`.
    - Visibility: gated by `ShouldShowSizingGrip` (resizable border, not minimized/maximized, not drawing if a merged `StatusStrip` provides its own).

- Theme-correct rendering
  - `DrawSizingGrip` uses a palette-derived color with a contrast check and falls back to a themed resource if available, so the glyph remains visible regardless of theme while still respecting palette visuals.

### Why this survives fast resizing
- Every WM_PAINT completes the full client draw; immediately afterward, we clip to the grip rect and draw the glyph. If the window size changes again, we first invalidate the old small client rect so it gets repainted by the normal pipeline, then we draw the grippie in the new rect, again clipped. There’s no chance for later paints in that frame to overwrite it, and clipping ensures no stray pixels.

- Because we draw using the window DC and a tiny intersected clip, the operation is cheap and doesn’t cause invalidation storms or flicker, even with many WM_SIZE/WM_PAINT messages during a drag.

- Not relying on the `InternalPanel` avoids the known constraint where its events don’t reliably fire in this form’s structure.

- **Result**: The grippie is always painted last, only inside its rect, and the previous area is proactively cleared, eliminating the artifacts.

Smurf-IV (approved):

PR #2432, opened 2025-08-17
1022 V100 Enhance PaletteMaterialBase (`Material - Dark`)

tobitege:
Followup to PR #2414 (Issue #1022 )

Enhance theme with improved border visibility and content text colors. Updated border return values for FormMain and HeaderForm styles to ensure a visible outline. Added new methods for content short text colors to align with scheme-defined header text for various grid styles.

Issues addressed:
- form top border becomes visually "invisible" against dark backgrounds, re-enable 1px border
- `KryptonDataGridView` header cells (and other theme content areas) now have light font, not black, on dark background

### Edits applied
- `Source/Krypton Components/Krypton.Toolkit/Palette Builtin/Material/Bases/PaletteMaterialBase.cs`
  - Form border:
    - Changed `GetBorderWidth` for `FormMain`/`HeaderForm` to return 1 instead of 0, restoring a visible 1px outline.
    - Switched border colors for `FormMain`/`HeaderForm` to use base palette’s scheme (`base.GetBorderColor1/2`) instead of mirroring header background, so it picks up `FormBorderActive/Inactive` values from `PaletteMaterialDark_BaseScheme`.
  - DataGridView header text:
    - Overrode `GetContentShortTextColor1/2` for `PaletteContentStyle` grid header styles to return `BaseColors.HeaderText`. In Material Dark, this is white, so column/row header text turns white.

- Impact:
  - The top/left window/title bar now has a clear outline in Material Dark (1px using scheme `FormBorderActive/Inactive` colors).
  - `KryptonDataGridView` header row text is white under Material Dark, matching `HeaderText` in `PaletteMaterialDark_BaseScheme`.

---

**FIXED DISPLAY WITH THIS PR:**

image

PR #2440, opened 2025-08-19
2439 V100 KDataGridViewDateTimePickerCell/Column sizing/format

### LINKED ISSUE #2439: [Bug]: V100 KDataGridViewDateTimePickerCell autosizing cuts off text
**Describe the bug**
Cell type cuts off text of displayed text due to mismatch of internal formats.
If cell text is `19.08.2025 00:00:00`, setting `AutoSizeColumnMode` to `AllCells`, will cut off the text roughly after the first timestamp's "0".

Wrong:

Image

Correct:

Image

tobitege:
### KryptonDataGridViewDateTimePicker: correct preferred size and formatting for autosize; keep designer-safe defaults

Fixes #2439 - Fixes KDGV-related regressions introduced with #2432

### Summary
- Fix incorrect measurement for DateTime cells (long date/time) so AutoSizeColumnsMode=AllCells computes accurate widths, and None mode can adopt a correct minimum without exploding width.
- Ensure displayed text formatting and preferred size calculations use a full date+time pattern when no explicit cell style format is provided.
- Avoid forcing designer-serialized defaults; sizing is corrected at runtime without introducing breaking visual defaults.

### Details
- KryptonDataGridViewDateTimePickerCell
  - GetFormattedValue: When the cell value is DateTime and no cell-style format is provided, fall back to:
    - column `Format` mapping (Long/Short/Time/Custom → culture patterns; Custom via `CommonHelper.MakeCustomDateFormat`), and finally
    - “G” general date/time format
    - Result is culture-aware via `cellStyle.FormatProvider` or `CurrentCulture`.
  - Paint: When `formattedValue` is a string that parses to DateTime and cell style format is empty, apply the same fallback format logic as above before drawing text. Keeps text rendering consistent with formatted value and width measurement.
  - GetPreferredSize:
    - For headers (rowIndex < 0): use base size plus indicator width/margin.
    - For data rows: compute the text to display using the same formatting fallback as above; measure with `TextRenderer.MeasureText` honoring alignment/wrap; return width = measured text + padding + indicator width/gap (+1), height = max of measured and base height.
    - This ensures autosize widths include the indicator and the true date+time text length.

- KryptonDataGridViewDateTimePickerColumn
  - Kept DefaultCellStyle untouched to avoid forcing “G” into designer serialization. A `ShouldSerializeDefaultCellStyle()` helper is provided so only non-default style changes are serialized.

### Rationale
- Base DataGridView sizing treats raw DateTime without the correct display format, often leading to date-only measurement and truncated times. By formatting consistently (or defaulting to “G”), we measure exactly what is displayed.
- When AutoSize is None, the grid will not resize automatically; ensuring a minimum width equal to preferred prevents visible truncation without changing application-set widths when already sufficient.
- Avoiding a hard default cell-style format prevents unintended designer changes while still fixing runtime sizing/painting.

### Impact
- Visual: DateTime picker cells size correctly to include full timestamp and indicator; no truncation in AllCells, and reduced/avoided truncation in None unless app intentionally narrows columns.
- Behavior: No breaking change to default cell styles serialized by the designer; existing explicit formats/custom formats are honored.
- Performance: Minimal; text measurement is localized to preferred-size calculation.

---

DateTime column editor now correctly allows to save `DefaultCellStyle` with empty `Format` ("No formatting"):
Before this was always rewritten to "G". Now the `CustomFormat` is finally used:

{2A55C788-3227-4A96-9A3B-F792C725780B}

---

{11A66B95-1B50-441A-B10E-A8023AF6B29B}

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridViewComboBoxCell.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridViewComboBoxCell.cs
@@ -382,7 +382,7 @@ public class KryptonDataGridViewComboBoxCell : DataGridViewTextBoxCell
         if (DataGridView?.EditingControl is not null)
         {
             bool rtl = DataGridView.RightToLeft == RightToLeft.Yes;
-            int locX = rtl ? editingControlBounds.X + IndicatorGap : editingControlBounds.X;
+            int locX = rtl ? editingControlBounds.X : editingControlBounds.X + IndicatorGap;
             int width = editingControlBounds.Width - IndicatorGap;
             DataGridView.EditingControl.Location = new Point(locX, editingControlBounds.Y);
             DataGridView.EditingControl.Size = new Size(width, IndicatorSize - 2);


giduac:
Testing...

giduac:
@tobitege 

```cs
    private void EnsureColumnWidthForContent()
    {
        if (DataGridView == null)
        {
            return;
        }

        int preferred = GetPreferredWidth(DataGridViewAutoSizeColumnMode.AllCells, true);
        if (DataGridView.AutoSizeColumnsMode == DataGridViewAutoSizeColumnsMode.None)
        {
            if (Width < preferred)
            {
                Width = preferred;
            }
        }
        else if (DataGridView.AutoSizeColumnsMode == DataGridViewAutoSizeColumnsMode.Fill)
        {
            // Nudge FillWeight so this column receives at least its proportional share
            // matching the ratio between preferred and current width
            if (Width > 0)
            {
                double ratio = Math.Max(1.0, (double)preferred / Width);
                FillWeight = (float)Math.Min(65535.0, FillWeight * ratio);
            }
        }
    }
```
AllCells mode is a performance killer. Specially on larger sets because also the data that is not in view will be checked for it's size. On a table with 10,000 rows....all rows will fetched in the background. Most operational grids will use fixed width since that gives better performance.

```cs 
if this.InheritedAutoSizeMode == DataGridViewAutoSizeColumnMode.None
// The column width should not be touched. It's the choice of the user.

if this.InheritedAutoSizeMode == DataGridViewAutoSizeColumnMode.NotSet
// The mode is inherited from the parent
```
Please read up on these two about inheritance.
These are the properties to query not the one on the DataGridView / component level.
https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.datagridviewcolumn.inheritedautosizemode?view=windowsdesktop-9.0
https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.datagridviewcell.inheritedstyle?view=windowsdesktop-9.0

tobitege:
Ok, removed the whole `OnDataGridViewChanged` code block now.

Smurf-IV:
@tobitege  Conflicts :-(.. Stoopid Github !

tobitege:
> @tobitege Conflicts :-(.. Stoopid Github !

Fixed

giduac:
> Ok, removed the whole `OnDataGridViewChanged` code block now.

Will test this later on....

giduac:
@tobitege 
First thing I came across.
The indicator image on the DateTimePicker column is different from the rest.
All should use the same image.

Not part of this ticket but a problem anyway.
Looks like your AI buddy pushed it a little too far on the cell indicator image.
The fixed size was specifically chosen back then.

https://github.com/user-attachments/assets/6933af3f-2b3d-4458-82cf-c3ea7be12d7c

Testing the rest of the change coming...

tobitege:
Yeh, will look into the scaling, that shouldn't happen.
But as to the glyph, they look the same to me:

{DB3FBA84-6F41-4764-8BFB-D51A86367835}

But I'll have a look, still.

giduac:
> Yeh, will look into the scaling, that shouldn't happen. But as to the glyph, they look the same to me:
> 
> {DB3FBA84-6F41-4764-8BFB-D51A86367835}
> But I'll have a look, still.

Strange not here, using the same branch...

image

giduac (approved):

tobitege:
Thanks @giduac for the time to test!
Don't merge yet, I'll take a look into the arrows tomorrow.

giduac:
@tobitege 

Pleasure...
Testing pays off and makes our things better!

tobitege:
I'm stuck with this one, weird issue *only* with the datetime picker column (see DataGrid demo).
Been fighting for 2 days to get this seemingly easy issue resolved that in edit mode (when the edit control _does_ get sized to full width, which it isn't now on purpose) the dropdown glyph gets overpainted with background color again. 😞

giduac:
> I'm stuck with this one, weird issue _only_ with the datetime picker column (see DataGrid demo). Been fighting for 2 days to get this seemingly easy issue resolved that in edit mode (when the edit control _does_ get sized to full width, which it isn't now on purpose) the dropdown glyph gets overpainted with background color again. 😞

@tobitege 

Will test the latest changes, but it might take a (busy atm) couple of days.

tobitege:
> > I'm stuck with this one, weird issue _only_ with the datetime picker column (see DataGrid demo). Been fighting for 2 days to get this seemingly easy issue resolved that in edit mode (when the edit control _does_ get sized to full width, which it isn't now on purpose) the dropdown glyph gets overpainted with background color again. 😞
> 
> @tobitege
> 
> Will test the latest changes, but it might take a (busy atm) couple of days.

Thanks, no hurry, I'll try my best to get ahead of this issue asap... just a tad frustrating, that an issue appears with a single cell type only.

tobitege:
Finally got it fixed... apparently it was a small thing... 🙈

giduac:
> Finally got it fixed... apparently it was a small thing... 🙈

Testing during the weekend....

tobitege:
@giduac - wait with testing for now. There's still something off with the dropdown glyph sizing and I'll try to get this back to a consistent appearance. Back to draft.

giduac:
> @giduac - wait with testing for now. There's still something off with the dropdown glyph sizing and I'll try to get this back to a consistent appearance. Back to draft.

@tobitege 
Testing ATM.

I don't see any reason to make the indicator resizable and rather see this taken out.
But replace the old image with the new one at the size we were using already.

tobitege:
The indicator is not resizable (not the intention), it wrongly appears that way currently because of the issue(s) I'm looking into.

giduac:
> The indicator is not resizable (not the intention), it wrongly appears that way currently because of the issue(s) I'm looking into.

@tobitege 
Just roll the indicator class back to it's original state and replace the image with the new one.

giduac:
@tobitege 

If formats are put in place by us these can and mostly will interfere with formatting the user applies at higher levels.
Any customisation on cell formatting should be removed.
This is up to the user not us or rather on the application-level.

Some more documentation on cell style inheritance.
https://learn.microsoft.com/en-us/dotnet/desktop/winforms/controls/cell-styles-in-the-windows-forms-datagridview-control

giduac:
@tobitege 

Testing...

The handling of the kdgv mouse down can be done in the same manner for all of these cell types:
image

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridViewDomainUpDownCell.cs
```diff
@@ -41,6 +41,23 @@ public override string ToString() =>
...
+    }
+
+    [Browsable(false)]
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+    public new bool MultilineStringEditor
+    {
+        get => base.MultilineStringEditor;
+        set => base.MultilineStringEditor = value;
+    }
+
```
giduac, line 60: Do these properties get serialised into the designer source?

```cs
    [Browsable(false)]
    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
    public new bool Multiline
    {
        get => base.Multiline;
        set => base.Multiline = value;
    }

    [Browsable(false)]
    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
    public new bool MultilineStringEditor
    {
        get => base.MultilineStringEditor;
        set => base.MultilineStringEditor = value;
    }
```

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridViewNumericUpDownCell.cs
```diff
@@ -59,6 +59,23 @@ public override string ToString() =>
...
+    }
+
+    [Browsable(false)]
+    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
+    public new bool MultilineStringEditor
+    {
+        get => base.MultilineStringEditor;
+        set => base.MultilineStringEditor = value;
+    }
+
```
giduac, line 78: Do these properties get serialised into the designer source?
```cs
    [Browsable(false)]
    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
    public new bool Multiline
    {
        get => base.Multiline;
        set => base.Multiline = value;
    }

    [Browsable(false)]
    [DesignerSerializationVisibility(DesignerSerializationVisibility.Hidden)]
    public new bool MultilineStringEditor
    {
        get => base.MultilineStringEditor;
        set => base.MultilineStringEditor = value;
    }
```

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridViewTextBoxCell.cs
```diff
@@ -190,20 +194,26 @@ public override void PositionEditingControl(bool setLocation,
...
             singleVerticalBorderAdded, singleHorizontalBorderAdded,
             isFirstDisplayedColumn, isFirstDisplayedRow);
 
-        editingControlBounds = GetAdjustedEditingControlBounds(editingControlBounds, cellStyle);
-        DataGridView!.EditingControl!.Location = new Point(editingControlBounds.X, editingControlBounds.Y);
-        DataGridView.EditingControl.Size = new Size(editingControlBounds.Width, editingControlBounds.Height);
+        editingControlBounds = GetAdjustedEditingControlBounds(editingControlBounds, padded);
+        editingControlBounds.X += IndicatorGap + 2;
+        editingControlBounds.Width -= IndicatorGap + 2;
+
```
giduac, line 218: Textbox cells don't use indicator images.

```cs
        editingControlBounds = GetAdjustedEditingControlBounds(editingControlBounds, padded);
        editingControlBounds.X += IndicatorGap + 2;
        editingControlBounds.Width -= IndicatorGap + 2;
```

tobitege:
So the issues actually started with my Material theme PR #2432 
With many(!) hours dabbling with this, I *think* I got it into a nicer state than it even was before - taste might differ: 😏 
- the old grid's dropdown arrows were ugly and small and just didn't fit together with the glyphs when editing a cell
- all dropdown cell types should now show almost pixel-perfect a glyph for editing and non-editing mode
- I even added more support for RTL than I thought was needed, but the DGV demo has an option to toggle RTL now
- the same demo now allows to add/edit columns in the designer as usual without breaking the code that fills the grid with test data

As to the earlier comment about formats: the PR doesn't **change** any formats: it aims at much better date conversion handling across different formats to NOT run into crash-to-desktop exceptions for the end user.

@PWagner1 @Smurf-IV I'd really would appreciate your guys' feedback on this, too.

If need absolutely be, I can put the old arrow display back, but it'll still look crap, imho.

tobitege:
{5FE7B284-3925-4D6D-80E3-BA1C4F6C82C5}

{0F31F7B6-58F3-4422-9137-954A7511D496}

tobitege:
{8C2947B5-C74B-47F7-8848-F0577E4A9F12}

(No, I did not look into spec buttons, something for another PR)

giduac:
@tobitege 

Naah, the arrow is cool. Couldn't find a nicer one when I added the default drop-down image to the columns.
Looks good so far.

Still a couple of pending questions open though...

tobitege:
> Still a couple of pending questions open though...

Which ones?

giduac:
image

tobitege:
where did you post those? I don't see them 🤔

giduac:
below this one
https://github.com/Krypton-Suite/Standard-Toolkit/pull/2440#issuecomment-3218008117

tobitege:
> below this one [#2440 (comment)](https://github.com/Krypton-Suite/Standard-Toolkit/pull/2440#issuecomment-3218008117)

Oh, I still don't see them - but the answer is: no, they're hidden/non-browsable and won't be serialized.

giduac (approved):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridViewDomainUpDownCell.cs
tobitege: No

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridViewNumericUpDownCell.cs
tobitege: No

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridViewTextBoxCell.cs
tobitege: The method does not have any glyph-related impact on text cells and can remain as is.

tobitege:
> Naah, the arrow is cool. Couldn't find a nicer one when I added the default drop-down image to the columns. Looks good so far.

Btw, my critique wasn't about the choice of glyph, that's fine with me, too. 😃 
It's more how small they were displayed and disjunct from the editing glyphs.

giduac (approved):

Issue #2448, opened 2025-08-23
[Bug]: KryptonForm does not display components without designer source


giduac:
**Describe the bug**
When a KForm is instatianted manually, added components are not displayed.

**To Reproduce**
```cs
KryptonForm form = new():
form.Controls.Add(new KryptonButton())):
form.Show();
```
**Expected behavior**
N/A

**Screenshots**
N/A

**Desktop (please complete the following information):**
V100

**Additional context**
PRs in the making

PR #2449, opened 2025-08-23
2448-V100-KryptonForm-does-not-display-components

giduac:
- Issue: #2448
- Resolves the problem
- And the changelog

compile-results


giduac:
> Will this be needed in V85 as well ?

@Smurf-IV 
Naah V100 only...

PR #2450, opened 2025-08-24
2448-V100-KryptonForm-does-not-display-components

giduac:
- Issue: #2448
- Small update to 2448 & PR #2449
- Adds a condition so that the code block is excuted only at runtime.


tobitege (approved):

PWagner1:
@giduac 

Conflicts :)

Issue #892, opened 2022-12-26
[Feature Request]: `KryptonButton` UAC Shield icons need to match the OS style


PWagner1:
`KryptonButton` UAC Shield icons need to match the OS style for supported sizes

Wagnerp:
Small (16 x 16), Medium (64 x 64) & Large (128 x 128)

Smurf-IV:
And 4K / 8K i.e. 256*256

Wagnerp:
@Smurf-IV How does the `Storage` API work, as I've been trying to combine all these options under one property with no success?

PWagner1:
@Smurf-IV 

Next on my todo list ;)

PR #2477, opened 2025-09-05
* OS & Themed UAC Shield images 

PWagner1:
* Implements #892


giduac (changes requested):
This PR mixes UAC shields with ThemedMenu & modern build changes.
We still have designer issues and another PR with themed menu and modern build changes.

Best to take the things which don't belong here out and put it on hold until the designer issues have been remediated.

Issue #648, opened 2022-02-13
[Feature Request]: SystemMenu to be theme related


Smurf-IV:
Ad part of the work on #646 there is a handler for the system menu now.

This could be replaced by a KryptonContextMenu with custom controls:
See the following locations for some guidance:

https://www.actiprosoftware.com/blog/post/25194/wpf-windowchrome-system-menu-customization
https://stackoverflow.com/questions/22852461/win32-api-getmenuiteminfo-returns-only-the-first-character-of-the-item-text
https://www.codeguru.com/csharp/manipulating-the-system-menu-using-c/

Some indicators of what to look for in the "Actions" / fields / bitmaps / enablement of menus here:
https://stackoverflow.com/questions/49056495/how-to-show-menu-item-with-icon-and-text-in-winforms-system-menu/49065061?noredirect=1#comment85223866_49065061

PWagner1:
https://github.com/user-attachments/assets/1ddfb617-4faf-4543-9e0c-2d912db04c5e

giduac:
@PWagner1 

V85 cannot build testform.
V100 cannot use the designer.
Both after the merge of 648.

As discussed in
https://discord.com/channels/1261930150585569350/1406345107816579203

giduac:
@PWagner1 

In V85 and V95 these debug message flood the output window. `Image Info - Type: Restore`
Can these please be taken out?

PWagner1:
> [@PWagner1](https://github.com/PWagner1)
> 
> In V85 and V95 these debug message flood the output window. `Image Info - Type: Restore` Can these please be taken out?

Will do

PWagner1:
Reopening for re-investigation in V110

PWagner1:
@giduac 

Can this be closed, or is there still more work to be done?

giduac:
> [@giduac](https://github.com/giduac)
> 
> Can this be closed, or is there still more work to be done?

Nah, there's another bug in it which I made fix for but will be posted after v100 is online, next week somewhere..

PR #2479, opened 2025-09-06
* Fix bug as noted in #648 - V100

PWagner1:
* Fix bug as noted in #648 - V100
* Renamed `KryptonThemedSystemMenu` has been renamed to `KryptonSystemMenu` and its associated APIs. Updated README to reflect these changes.

image


Smurf-IV (approved):

giduac (changes requested):
In my opinion this should be put on hold until the designer issues have been resolved.

Next to that the changes to /Scripts/ModernBuild have been mixed into this PR. This needs to be done in a separate PR to avoid problems when doing a rollback/revert.

Issue #2508, opened 2025-09-18
[Feature Request]: Add an internal InDesignMode extension method to Control and Components


giduac:
**Is your feature request related to a problem? Please describe.**
Some objects/classes have the `DesignMode` property, but not all.
This will add an extension method to both types that can be used to query if the object is in design mode.

**Describe the solution you'd like**
N/A

**Describe alternatives you've considered**
N/A

**Additional context**
PR underway...

giduac:
Remove the extension on Control since Control is derived from Component.

PR #2509, opened 2025-09-18
2508-V100-Add-internal-InDesignMode-extension-methods-to-Control-and-Component

giduac:
- Issue: #2508
- Add the extension methods
- And the changelog

Warnings are from an earlier PR, to be remediated.

compile-results

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/General/Extensions.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/General/Extensions.cs
@@ -1,9 +1,6 @@
 ﻿#region BSD License
 /*
  *
- * Original BSD 3-Clause License (https://github.com/ComponentFactory/Krypton/blob/master/LICENSE)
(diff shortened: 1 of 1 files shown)


Source/Krypton Components/Krypton.Toolkit/General/Extensions.cs
```diff
@@ -0,0 +1,37 @@
+﻿#region BSD License
+/*
+ *
+ * Original BSD 3-Clause License (https://github.com/ComponentFactory/Krypton/blob/master/LICENSE)
```
Smurf-IV: New files does not need the "Old" license mention from Component
And I think Amed can be removed

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/General/Extensions.cs
giduac: fixed

Issue #952, opened 2023-02-19
[Feature Request]: Place built NuGet packages into separate directory 


PWagner1:
Place the built NuGet packages into separate directory instead of the `bin` folder, for easier management

PR #2517, opened 2025-09-22
* Place built NuGet packages into separate directory

PWagner1:
* Implemented #952 
* Fixed `run.cmd`

image

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,7 @@
 ====
 
 ## 2025-11-xx - Build 2511 (V10 - alpha) - November 2025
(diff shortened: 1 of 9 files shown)


giduac:
@PWagner1 

Can you please test if this is correct.

image

Not sure if others use it but I do use local builds now and then.

PWagner1:
> @PWagner1
> 
> Can you please test if this is correct.
> 
> image
> Not sure if others use it but I do use local builds now and then.

@giduac 

It won't remove local builds, just place all the NuGet packages into a separate 'NuGet' directory under `Bin`

Smurf-IV (approved):

giduac (approved):

Issue #2490, opened 2025-09-13
[Bug]: KryptonComboBox uses incorrect item editor string


giduac:
**Describe the bug**
- Item editor string inhibits the editor from correct functioning at runtime from a propertygrid.
- Also `OnPaletteNeedPaint` override calls the wrong base.

**To Reproduce**
N/A

**Expected behavior**
N/A

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - Toolkit Version: V100

**Additional context**
PR underway...

PR #2529, opened 2025-09-28
2490-V100-KryptonComboBox-uses-incorrect-item-editor-string

giduac:
- Issue: #2490
- Correct Items property Editor string
- And the changelog

Warnings don't stem from this PR.

compile-results


PWagner1 (approved):

Smurf-IV:
Can this also be merged into v85 please ?

giduac:
> Can this also be merged into v85 please ?

It only occurs from a propertygrid at runtime.
We had no complains about it until I found it testing with KTD.

Better to let V85 and V95 fizzle out ...., imo...

PR #2601, opened 2025-10-28
* Remove retention days

PWagner1:
* Remove retention days
* Future-proof workflows
* Remove V95 branch


.github/workflows/nightly.yml
```diff
@@ -25,12 +25,19 @@ jobs:
         with:
           dotnet-version: 9.0.x
 
-      # .NET 10 (Preview)
-      - name: Setup .NET 10 (preview)
+      # .NET 10 (GA)
```
Smurf-IV, line 105: You have missed a few of these in the next change

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ on:
 
 jobs:
   build:
-    runs-on: windows-2022
+    runs-on: windows-2025
```
Smurf-IV: Would be easier to change this to `windows-latest`

Smurf-IV (changes requested):

.github/workflows/nightly.yml
PWagner1: Done

.github/workflows/build.yml
PWagner1: Done

PWagner1:
@Smurf-IV & @giduac 

Ok, so it seems that `nightly.yml` won't trigger **unless** it's on `master`. Seems that the default branch has some sort of special authority.

giduac:
> @Smurf-IV & @giduac
> 
> Ok, so it seems that `nightly.yml` won't trigger **unless** it's on `master`. Seems that the default branch has some sort of special authority.

Consult the documentation about how you need to set this up....
And if this is possible in the way you are currently doing this...

PWagner1:
> > @Smurf-IV & @giduac
> > Ok, so it seems that `nightly.yml` won't trigger **unless** it's on `master`. Seems that the default branch has some sort of special authority.
> 
> Consult the documentation about how you need to set this up.... And if this is possible in the way you are currently doing this...

It does work, however because it's using `- cron`, it means that it cannot run automatically, unless it's on the default branch. It seems that `master` has the 'final say'.

PWagner1:
@Smurf-IV & @giduac 

Please also evaluate PR https://github.com/Krypton-Suite/Standard-Toolkit-Online-Help/pull/57, as it contains documentation about how the workflows work, critical constraints etc.

giduac (changes requested):
This can wait imo, as further work on this will be done on the test repo...

PWagner1:
> This can wait imo, as further work on this will be done on the test repo...

It is working as expected, just need to make a few minor adjustments...

giduac:
> > This can wait imo, as further work on this will be done on the test repo...
> 
> It is working as expected, just need to make a few minor adjustments...

You did one test one time....for a process that needs to run solidly...

giduac:
Turning this into draft until testing is complete and avoid an accidental merge.

PWagner1:
Ok so in testing, it will run every night, release a new package if any changes are detected and won't if there's no changes. Equally there's a kill switch to pause the workflow from running. 

As for .NET 11, it's using the .NET 10 logic, while using .NET 10 as a fallback. It means that although 11 does not exist yet, it will not interfere with the build or publish process.

PR #2633, opened 2025-11-24
* Fix build script (Canary)

PWagner1:
* Fix build script - removes archive releases
* Keeps branch always 1 commit ahead of `master`


PWagner1:
Will redo once #2636 is merged

PR #2640, opened 2025-11-24
* Adds missing support for automatic V105 releases

PWagner1:
* Adds missing support for automatic V105 releases - these will follow the normal stable releases, until V85 is EOL.


PWagner1:
Not needed, addressed in PR #2652

Issue #2641, opened 2025-11-25
[Bug]: MDI Child Windows do not react when minimized.


giduac:
**Describe the bug**
When an mdi child is minimized the windows does no more react to mouse clicks.

**To Reproduce**
N/A

**Expected behavior**
The opposite

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - Toolkit Version: V105 / V110

**Additional context**
PRs are in the works...

PR #2642, opened 2025-11-25
2641-V110-MDI-Child-Windows-do-not-react-when-minimized

giduac:
- Issue: #2641
- Corrects the MDI child window behaviour
- Adds an MDI Window to TestForm
- And the changelog

image

PR #2643, opened 2025-11-25
2641-V105-MDI-Child-Windows-do-not-react-when-minimized

giduac:
- Issue: #2641
- Corrects the MDI child window behaviour
- Adds an MDI Window to TestForm
- And the changelog

image

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -2,6 +2,11 @@
 
 ====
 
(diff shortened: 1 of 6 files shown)


PWagner1:
@giduac 

Is it ok to hold till I put in the kill switch stuff? PR #2642 is ok to merge, sorted out all the branch parity stuff. We're back to normal.

Issue #2631, opened 2025-11-23
[Bug]: KryptonForm FormBorderStyle.None bug


mohatchaks:
**Describe the bug**
When set KryptonForm FormBorderStyle to none and windowstate to maximize it will cut Form ( Left + Top )

**Screenshots**

<img width="1400" height="1026" alt="Image" src="https://github.com/user-attachments/assets/9f33be6f-3666-498e-8501-fdbc84f01475" />

<img width="1923" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ae05a23d-c3ad-42b0-8715-44dfc68842b6" />

 - OS: Windows 10
 - Framework/.NET Version: 4.8
 - Toolkit Version: Krypton.Navigator.Canary 100.25.11.321-beta

mohatchaks:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/e5408586-a4b8-47ef-9b61-807fa11b54f9" /> here is a clear image

AngeloCresta:
Also PaletteBorder.cs now has errors:

Before:

<img width="444" height="277" alt="Image" src="https://github.com/user-attachments/assets/90bd10ef-e498-4aa8-a142-b4e6d63f18a4" />

After:

<img width="424" height="222" alt="Image" src="https://github.com/user-attachments/assets/04aae71e-4c63-4b1e-bc5f-67728fcc1593" />


Rever this part back to have it solved (PaletteBorder.cs):

```

    /// <summary>
    /// Gets the border width.
    /// </summary>
    /// <param name="state">Palette value should be applicable to this state.</param>
    /// <returns>Border width.</returns>
    public int GetBorderWidth(PaletteState state) => Width != -1 ? Width : _inherit.GetBorderWidth(state);
    #endregion

    #region Rounding
    /// <summary>
    /// Gets and sets the border rounding.
    /// </summary>
    [KryptonPersist(false)]
    [Category(@"Visuals")]
    [Description(@"How much to round the border corners.")]
    [DefaultValue(GlobalStaticValues.DEFAULT_PRIMARY_CORNER_ROUNDING_VALUE)]
    [RefreshProperties(RefreshProperties.All)]
    public float Rounding
    {
        get => _storage?.BorderRounding ?? GlobalStaticValues.DEFAULT_PRIMARY_CORNER_ROUNDING_VALUE;

        set
        {
            if (_storage != null)
            {
[...]

PR #2675, opened 2025-11-28
2631-V110-KryptonForm-FormBorderStyle-None-produces-incorrect-results

giduac:
- Issue: #2631
- Override the PaletteFormBorder draw property to return false when FormBorderStyle is none
- And the change log

image


PWagner1:
@giduac 

Conflicts

Issue #2681, opened 2025-11-28
[Bug]: KryptonDataGridView ( Problem in ColumnHeaders when scrolling  )


mohatchaks:
### Summary

Problem in ColumnHeaders Painting when scrolling ( using scrollbar )

Edit by @giduac (2025-11-29)
- This is probably connected to the OnCellPaint update form (#2422).
- Changing from DrawText to DrawString when painting the header text does not resolve the problem.
- It only happens when the scroll is performed through the mouse and the scrollbar is dragged around (fast).
- An override will be added for OnScroll that will force a header cell repaint when a horizontal scroll by the mouse happens.
- V110 & V105 affected (85 works fine)

### Description

There is Problem when scrolling using scrollbar 
here is a gif

![Image](https://github.com/user-attachments/assets/1177875c-a232-4dd9-b249-46b880dd03ef)

### Steps to Reproduce

just kryptondatagridview with soma columns (more then 10+ columns i was testing )

### Expected Behavior

.

### Actual Behavior

ColumnHeaders Painting problem when scrolling

### Operating System

Windows 10

### Framework/.NET Version

4.8 - 4.6.2

### Toolkit Version

110.25.11.332-alpha nigthly -  90.24.11.317

### Additional Information

in version 90.24.12.345 there is no problem

### Areas Affected

None

PR #2686, opened 2025-11-29
2681-V110-KryptonDataGridView-column-headers-do-not-repaint-on-scroll

giduac:
- Issue: #2681
- Adds an OnScroll override that selectively repaints the column headers on horizontal mouse scroll
- And the change log

image


PWagner1:
@giduac 

Conflicts

giduac:
Reverting for a better solution.

Issue #2658, opened 2025-11-26
[Feature Request]: Achieve 100% parity with WinForms controls


PWagner1:
Achieve 100% parity with WinForms controls, at the moment we're 85% there. See [audit](https://krypton-suite.github.io/Standard-Toolkit-Online-Help/articles/Contributing/Auditing/KryptonToolkitAudit.html) for further details.

- [ ] `KryptonRating`
- [ ] `KryptonSplitButton` - KButton does have drop-down functionality
- [ ] `KryptonTagInput`

PWagner1:
@Krypton-Suite/reviewers 

See https://github.com/Krypton-Suite/TEST-Standard-Toolkit for completed/WIP controls.

Issue #1297, opened 2024-01-20
[Bug]: Context, Menu, Status and Tool Strip fonts don't follow Krypton's


PWagner1:
Whilst testing, I encountered a bug whereby the Context, Menu, Status and Tool Strip fonts all don't follow Krypton's base font.

![Font Bug](https://github.com/Krypton-Suite/Standard-Toolkit/assets/949607/89020eee-6d2e-456b-8758-0bc4d1617bbb)

Expectation, for these controls to do the same as other Krypton controls.

Smurf-IV:
Part of this is a duplicate (Menu's and Toolstrips)
But it is expected because they are "Native" WinForm components.

So the fix (If it can be) would be the same as the "Integrate StatusStrip" designer option, so that it does not break backwards compatibility.

Wagnerp:
@Smurf-IV & @giduac

Append this to backlog?

Smurf-IV:
It's "An existing" winform thing, So I expect it to be in V90

Wagnerp:
@Smurf-IV Is it possible to override the font to use KManager's `BaseFont`?

Smurf-IV:
> @Smurf-IV Is it possible to override the font to use KManager's `BaseFont`?

Might / Should be.
Or whatever the theme is currently using for KContextMenu's at least.

Wagnerp:
Hi @Smurf-IV & @giduac

What's `Segoe UI, 12world`?

![image](https://github.com/user-attachments/assets/66a21d6e-1c8e-40f2-8985-6a9f4d7a9548)

Smurf-IV:
Where did that image come from? It's bit cryptic without more info ?
This is what I get for the BaseFont...
![image](https://github.com/user-attachments/assets/b0019f29-597a-491d-8222-7bac5493140f)

Wagnerp:
> Where did that image come from? It's bit cryptic without more info ?
This was from the `MenuStrip`

Smurf-IV:
??
![image](https://github.com/user-attachments/assets/0ce0ad54-e31d-4b02-8b67-1f23a8210f90)

Wagnerp:
[...]

PR #2690, opened 2025-11-29
* `KryptonMenuStrip` - Part of #2658

PWagner1:
* Implements #2689
  - **Note:** Still needs to update with font changes.

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -3,6 +3,8 @@
 ====
 
 ## 2026-11-xx - Build 2611 (V110 Nightly) - November 2026
+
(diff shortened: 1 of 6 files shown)


PWagner1:
@Smurf-IV 

> Is there going to be an Example app modified for this as well please ?

There is one in the `TestForm`, but #1297 needs looking into

Smurf-IV:
> There is one in the `TestForm`, but #1297 needs looking into

Yeah, Test Form is not seen by developers who download the nugets etc.

PWagner1:
> > There is one in the `TestForm`, but #1297 needs looking into
> 
> Yeah, Test Form is not seen by developers who download the nugets etc.

Will, introduce it alongside `KryptonToolStripContainer`

Issue #2230, opened 2025-06-15
[Feature Request]: A `KryptonSearchBox` control


PWagner1:
Can we have a dedicated `KryptonSearchBox` control with autocomplete etc?

mohatchaks:
can you add something like this : https://docs.devexpress.com/WindowsForms/DevExpress.XtraEditors.GridLookUpEdit
or 
search with multi column example ( i use krypton toolkit ) : 
![Image](https://github.com/user-attachments/assets/1e71b050-f3ea-47eb-9709-4edc68f9f9c7)

PWagner1:
@mohatchaks 

Can do as a future feature, currently looks like this...

![Image](https://github.com/user-attachments/assets/9b0d69cf-9b18-4a40-b489-3aa3abaaf3df)

PR #2714, opened 2025-12-02
* A `KryptonSearchBox` control

PWagner1:
* Implemented #2230 - Documentation & demos to follow
* Rest of filescoped namespaces


giduac:
@Krypton-Suite/reviewers 

- Mixing all kind of files in one PR has been discussed a time or two... iirc
- What I've read about the IDE generated files and making those use file scoped namespaces might causes a problem or two. My advice would be to leave formatting of those (system generated files) to the system.

The SearchBox 
- Is not really a control but an implementation. 
- It is not part of Winforms and should not be in here.

I think I mentioned this earlier but these kind of tools or utilities could be hosted in a lib called "Krypton Utilities"
That will keep controls and implementations separated. I would also draw this line for Extended.

PWagner1:
> @Krypton-Suite/reviewers
> 
> * Mixing all kind of files in one PR has been discussed a time or two... iirc
> * What I've read about the IDE generated files and making those use file scoped namespaces might causes a problem or two. My advice would be to leave formatting of those (system generated files) to the system.
> 
> The SearchBox
> 
> * Is not really a control but an implementation.
> * It is not part of Winforms and should not be in here.
> 
> I think I mentioned this earlier but these kind of tools or utilities could be hosted in a lib called "Krypton Utilities" That will keep controls and implementations separated. I would also draw this line for Extended.

@giduac 

One of the intended applications for this is the `KryptonExceptionDialog` I can make a internal version of it that is not visible to the public. I like the idea of a new `Krypton.Utilities` library, as it can be exclusive to the AIO package and not have to manage another NuGet package, so I'll put the public version in there. Please keep this PR open so I can see what files I need to modify etc.

PR #2744, opened 2025-12-08
* Fix VisualMultilineStringEditorForm only saving last line

PWagner1:
# Fix VisualMultilineStringEditorForm only saving last line

## Summary
Fixes a critical bug in `VisualMultilineStringEditorForm` where clicking OK would only save the last line of text instead of all lines. The issue affected both RichTextBox and TextBox modes of the multiline string editor.

## Changes Made
- **Fixed `kbtnOk_Click` method**: Replaced incorrect foreach loop that overwrote `_contents` with a single-element array on each iteration
- **Properly copy all lines**: Changed to retrieve the `Lines` array from the control and copy all elements into `_contents` using a for loop
- **Removed TODO comments**: Removed the TODO comments that documented the bug (lines 176 and 187)
- **Applied fix to both code paths**: Fixed both RichTextBox (`krtbContents.Lines`) and TextBox (`ktxtStringCollection.Lines`) implementations

## Technical Details
The bug was caused by iterating over `Lines` with a foreach loop where `line` is a single `string`. On each iteration, a new array containing only that single line was created and assigned to `_contents`, overwriting the previous value. This resulted in only the last line being preserved.

The fix:
1. Retrieves the complete `Lines` array from the control
2. Creates a new array of the correct size (`lines.Length`)
3. Copies each line element into the new array

## Impact
- **Bug fix**: Multiline string editor now correctly saves all lines instead of just the last one
- **No breaking changes**: Public API remains unchanged
- **No UI changes**: Visual behavior is identical; only the data persistence logic is corrected
- **Affects both modes**: Fix applies to both RichTextBox and standard TextBox multiline editing modes

## Related
Resolves TODO comments in `VisualMultilineStringEditorForm.cs` lines 176 and 187

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -5,6 +5,7 @@
 ## 2026-11-xx - Build 2611 (V110 Nightly) - November 2026
 
 * Resolved [#2745](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2745b), `VisualMultilineStringEditorForm` only saving last line
+* Implemented [#2728](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2728), Ability for developers to set the highlight colour
 * Implemented [#2628](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2628), Improve build system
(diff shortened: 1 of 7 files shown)


PWagner1:
@Smurf-IV & @giduac 

Is this ok to merge?

PR #2759, opened 2025-12-16
* `SaveConfigToArray` and `LoadConfigFromArray` not working correctly

### LINKED ISSUE #2516: [Bug]: `SaveConfigToArray` and `LoadConfigFromArray` not working correctly
_From @ckak07_

Hello,

I am using the Krytpon-Docking-Library in the nuget package version 95.25.8.235. I am implementing an application that looks like this:

Image

If I create a new KryptonPage and place this dockable window in the right edge of the logging window (The bottom window) I get the following result:

Image

If I save the configuration and load it when starting the application I get the following result:

Image

You see that the new added KryptonPage called DeviceLog is missing. But If change the order how I place the device log page on my screen the result changes. If I restart my application and then follow the procedure below, the DeviceLog window appears correctly the next time I start the application:

- Create DeviceLog window
- Dock the DeviceLog window to the chart window
- Then dock the DeviceLog window to the Logging window

Following this procedure the result of my application after starting it looks like this:

Image

What I did next is compare the configurations I get for both processes. Here are the files:

[wrongConfiguration.txt](https://github.com/user-attachments/files/22452666/wrongConfiguration.txt)

[correctConfiguration.txt](https://github.com/user-attachments/files/22452669/correctConfiguration.txt)

Comparing the configurations show that the correct configuration contains another section under the chart workspace.

Image

Please help me to solve the problem.

Best regards

_[Original post](https://github.com/ComponentFactory/Krypton/issues/270)_

PWagner1:
Fixes #2516

### Problem

When docking a `KryptonPage` directly to a workspace edge (left, right, top, or bottom), the page would not be saved correctly when using `SaveConfigToArray()` and `LoadConfigFromArray()`. The page would appear visually but would be missing after reloading the configuration.

The issue occurred specifically when:
- A new page is created and docked directly to the edge of a workspace window
- The configuration is saved immediately after docking

However, if the page was first docked to another window and then moved to the target location, it would save correctly because the workspace structure was already fully established.

### Root Cause

The `SaveElementToXml` method in `KryptonDockingSpace` was saving the workspace layout without ensuring that structural changes (such as new cells/sequences added when docking to edges) were committed to the `Root.Children` collection. Without calling `PerformLayout()`, the workspace might not have fully processed these structural changes before serialization, resulting in an incomplete configuration being saved.

### Solution

Added a `PerformLayout()` call in `KryptonDockingSpace.SaveElementToXml()` before saving the layout. This ensures all structural changes are committed to the workspace's internal structure before serialization, so that pages docked directly to edges are properly saved.

### Changes

- **Modified**: `Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingSpace.cs`
  - Added `SpaceControl.PerformLayout()` call before `SaveLayoutToXml()` to ensure structural changes are committed

### Testing

To verify the fix:

1. Create a new `KryptonPage` (e.g., "DeviceLog")
2. Dock the page directly to the right edge of a workspace window
3. Save the configuration using `SaveConfigToArray()`
4. Close and restart the application
5. Load the configuration using `LoadConfigFromArray()`
6. **Expected**: The docked page should now be present and correctly positioned after reload

A comprehensive test form has been added in `Source/Krypton Components/TestForm/DockingConfigSaveLoadTest.cs` that demonstrates and tests this functionality.

### Breaking Changes

None. This is a bug fix that maintains backward compatibility.

### DIFF
--- a/Source/Krypton Components/TestForm/DockingConfigSaveLoadTest.Designer.cs
+++ b/Source/Krypton Components/TestForm/DockingConfigSaveLoadTest.Designer.cs
@@ -50,7 +50,6 @@
             this.btnCreateChartPage = new Krypton.Toolkit.KryptonButton();
             this.kryptonGroupBoxStatus = new Krypton.Toolkit.KryptonGroupBox();
             this.kryptonTextBoxStatus = new Krypton.Toolkit.KryptonTextBox();
-            this.kryptonLabel1 = new Krypton.Toolkit.KryptonLabel();
             ((System.ComponentModel.ISupportInitialize)(this.kryptonPanel1)).BeginInit();
             ((System.ComponentModel.ISupportInitialize)(this.kryptonDockableWorkspace1)).BeginInit();
             ((System.ComponentModel.ISupportInitialize)(this.panelControls)).BeginInit();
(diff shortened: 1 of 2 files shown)


PWagner1:
@Krypton-Suite/reviewers 

Is this ok to merge?

Issue #2570, opened 2025-10-11
[Bug]: KryptonRibbon tab not scroll bar


gokemg:
1.KryptonRibbon tab not scroll bar 
2.tab text can not show long text

![Image](https://github.com/user-attachments/assets/e618a99a-d19b-4281-a8db-98110624255e)

<img width="817" height="174" alt="Image" src="https://github.com/user-attachments/assets/cf3d71ca-6c8e-4469-800d-d173e1420ac7" />

gokemg:
Toolkit Version: [100.25.10.279]

lesandrog:
@gokemg 
I tested version 100.25.10.290-alpha.
1. Scrolling works, as long as the window is active and the mouse is over the tabs.

https://github.com/user-attachments/assets/cd7e67a5-a461-4b5e-be5f-00f6e0b6c892


2. Tab text is cut off when the form doesn't have enough space to display all the text. The MS Office ribbon behaves the same way.

https://github.com/user-attachments/assets/61177ce1-c100-49bb-bbf9-8d7b87d24723

lesandrog:
@giduac 
What's your opinion on whether scrolling the mouse over the ribbon's `GroupArea` also changes tabs, like in MS Office?
And with an option to enable and disable it.

https://github.com/user-attachments/assets/d6d20949-bc58-490d-9603-fc6d7af6dd00

Issue #331, opened 2021-09-08
[Feature Request]: Make the "No Tab in a ribbon Solution" An Actual Designer Tool


Smurf-IV:
https://github.com/ComponentFactory/Krypton/issues/99#issuecomment-914968988

Wagnerp:
Yes it should

Smurf-IV:
This can be done in 2 ways (Please vote)
1) A better solution that catches all edge cases to the proposed solution in the OP
2) Create a new Class that removes everything and reimplements the Ribbon control only a single ribbon instance (i.e no possibility of programatic tab selection)

Both have pro's and con's and both can be done easily

Or option 3) do both !

Wagnerp:
> This can be done in 2 ways (Please vote)
> 
> 1. A better solution that catches all edge cases to the proposed solution in the OP
> 2. Create a new Class that removes everything and reimplements the Ribbon control only a single ribbon instance (i.e no possibility of programatic tab selection)
> 
> Both have pro's and con's and both can be done easily
> 
> Or option 3) do both !

I would say option 3, as you would get the benefits of both solutions.

Smurf-IV:
Better workaround code:
``` c#
/// <summary>
    /// Ribbon control `Without` the Tabs
    /// </summary>
    public class NoTabRibbon : KryptonRibbon
    {
        public NoTabRibbon()
        {
            var designMode = (LicenseManager.UsageMode == LicenseUsageMode.Designtime);
            if (designMode || DesignMode)
                return;
            // Use reflection to get at 
            //  `private ViewLayoutDocker _ribbonDocker;`
            Type t = typeof(KryptonRibbon);
            var field = t.GetField(@"_ribbonDocker", System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
            var /*ViewLayoutDocker*/ ribbonDocker = field.GetValue(this) as ViewLayoutDocker;

            //this._ribbonDocker.Add((ViewBase) this.GroupsArea, ViewDockStyle.Fill);
            //this._ribbonDocker.Add((ViewBase) this._minimizeBar, ViewDockStyle.Bottom);
            //this._ribbonDocker.Add((ViewBase) this._qatBelowRibbon, ViewDockStyle.Bottom);
            //this._ribbonDocker.Add((ViewBase) this.TabsArea, ViewDockStyle.Top);
[...]

PR #2767, opened 2025-12-17
2570-V110-Added-Tab-scrolling-with-mouse-over-Ribbon's-GroupsArea

lesandrog:
Add in [[Bug]: KryptonRibbon tab not scroll bar #2570](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2570)

Added Tab scrolling with mouse over Ribbon's GroupsArea.

The ScrollTabGroupArea property has also been added, so that when ["No Tab" #331](https://github.com/Krypton-Suite/Standard-Toolkit/issues/331) is deployed, setting it to false disables scrolling.

image

### DIFF
--- a/Source/Krypton Components/Krypton.Ribbon/Controls Ribbon/KryptonRibbon.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/Controls Ribbon/KryptonRibbon.cs
@@ -1753,16 +1753,10 @@ public class KryptonRibbon : VisualSimple,
     #region WIN32 Calls
     public static class MouseControlFinder
     {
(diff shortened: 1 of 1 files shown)


giduac (changes requested):
```cs
        [DllImport(Libraries.User32, CharSet = CharSet.Unicode)]
        private static extern IntPtr WindowFromPoint(Point point);

        [DllImport(Libraries.User32, CharSet = CharSet.Unicode)]
        private static extern IntPtr GetParent(IntPtr hWnd);
```
These imports are already available in PI as:
```cs
PI.WindowFromPoint();
PI.GetParent();
```

giduac (approved):

Issue #2762, opened 2025-12-17
[Feature Request]: Move the public facing version of `KryptonExceptionDialog` to `Krypton.Utilities`


PWagner1:
### Feature Title

Move the public facing version to `Krypton.Utilities`

### Feature Description

Move the public facing version of `KryptonExceptionDialog` to `Krypton.Utilities`, while retaining a 'internal' version as to not break anything.

PR #2785, opened 2025-12-20
* Move the public facing version of KryptonExceptionDialog to Krypton…

PWagner1:
….Utilities

* Implemented #2762


PWagner1:
> 'KryptonExceptionDialog' is inaccessible due to its protection level

@Smurf-IV 

Fixed missing import

Issue #2156, opened 2025-03-23
[Feature Request]: Native support for Font Awesome


PWagner1:
Native support for Font Awesome https://fontawesome.com/

giduac:
> Native support for Font Awesome https://fontawesome.com/

This is no free lunch -> https://fontawesome.com/plans

Smurf-IV:
> This is no free lunch -> https://fontawesome.com/plans

The earlier versions are still free.

Smurf-IV:
@PWagner1 
And doesn't this already exist as a request ?

giduac:
> > This is no free lunch -> https://fontawesome.com/plans
> 
> The earlier versions are still free.

But then again that's where it stops, so further new releases are to pay for?

PWagner1:
> [@PWagner1](https://github.com/PWagner1) And doesn't this already exist as a request ?

@Smurf-IV 

I don't think so... but it could be combined with #2220

PWagner1:
@Krypton-Suite/reviewers 

I think that this one can go into `Krypton.Utilities`.

PR #2805, opened 2025-12-27
* `KryptonTextBox` Validating event fires twice (V105)

### LINKED ISSUE #2801: [Bug]: `KryptonTextBox` Validating event fires twice
### Summary

The Validating event is raised twice for a single focus change.

### Description

The Validating event is raised twice for a single focus change

### Steps to Reproduce

Create a WinForms form.
Add two KryptonTextBox controls.
Handle their Validating event.
Put a simple MessageBox.Show("Test") inside the handler.
Move focus between the textboxes (Tab or mouse).

### Expected Behavior

The Validating event should be raised once per control when it loses focus, consistent with standard WinForms TextBox behavior.

### Actual Behavior

Each textbox raises Validating two times, so with two controls the message box appears four times instead of two.

### Operating System

win 10 pro

### OS Version

_No response_

### Framework/.NET Version

 .NET Framework 4.8.1

### Toolkit Version

100.25.11.328

### Additional Information

VB.NET, using Visual Studio 2019.

### Areas Affected

None

PWagner1:
* Closes #2801

---

> [!NOTE]
> Addresses validation event duplication (issue #2801).
> 
> - Introduces `ForwardValidating(...)`/`ForwardValidated(...)` and suppression logic in `VisualControlBase` (flag `_isForwardingValidationFromChild`, overrides `OnValidating`/`OnValidated`) to avoid double-firing when `ControlStyles.ContainerControl` is set
> - Updates controls to forward child events instead of calling `OnValidating`/`OnValidated` directly: `KryptonTextBox`, `KryptonMaskedTextBox`, `KryptonComboBox`, `KryptonListBox`, `KryptonCheckedListBox`, `KryptonDomainUpDown`, `KryptonNumericUpDown`, `KryptonRichTextBox`, `KryptonTreeView`, `KryptonCalcInput`
> - Adds `KryptonTextBoxValidatingTest` form and StartScreen entry for manual verification
> - Updates `Changelog.md` (build/date and notes for #2801)
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f2614195d9ba2077a4b8390d8b7b583f8895a3a3. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCalcInput.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCalcInput.cs
@@ -1315,8 +1315,10 @@ public class KryptonCalcInput : VisualControlBase, IContainedInputControl
 
     private void OnTextBoxPreviewKeyDown(object? sender, PreviewKeyDownEventArgs e) => OnPreviewKeyDown(e);
 
+    // TODO: Workaround for issue where ContainerControl style causes duplicate validation events. See issue https://github.com/Krypton-Suite/Standard-Toolkit/issues/2801 for details.
     private void OnTextBoxValidating(object? sender, CancelEventArgs e) => ForwardValidating(e);
 
+    // TODO: Workaround for issue where ContainerControl style causes duplicate validation events. See issue https://github.com/Krypton-Suite/Standard-Toolkit/issues/2801 for details.
     private void OnTextBoxValidated(object? sender, EventArgs e) => ForwardValidated(e);
 
     private void OnTextBoxMouseEnter(object? sender, EventArgs e)
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCheckedListBox.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCheckedListBox.cs
@@ -2496,8 +2496,10 @@ public class KryptonCheckedListBox : VisualControlBase,
 
     private void OnListBoxKeyDown(object? sender, KeyEventArgs e) => OnKeyDown(e);
 
+    // TODO: Workaround for issue where ContainerControl style causes duplicate validation events. See issue https://github.com/Krypton-Suite/Standard-Toolkit/issues/2801 for details.
     private void OnListBoxValidated(object? sender, EventArgs e) => ForwardValidated(e);
 
+    // TODO: Workaround for issue where ContainerControl style causes duplicate validation events. See issue https://github.com/Krypton-Suite/Standard-Toolkit/issues/2801 for details.
     private void OnListBoxValidating(object? sender, CancelEventArgs e) => ForwardValidating(e);
 
     private void OnListBoxPreviewKeyDown(object? sender, PreviewKeyDownEventArgs e) => OnPreviewKeyDown(e);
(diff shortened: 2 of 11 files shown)


KamaniAR:
I think this is not enough, need to work more on both Validating and Validated

This all tested on both Validating and Validated
KryptonTextBox - fires twice
KryptonCalcInput - fires **thrice**
KryptonComboBox - fires twice
KryptonMaskedTextBox - fires twice
KryptonNumericUpDown - fires twice
KryptonRichTextBox - fires twice
kryptonListBox - fires twice

Also check all input type controls if i forgot in this test.

giduac:
@PWagner1 

I had a look at this to, and its not sure that this only is present in `KTextBox.`
Since `KTextBox` derives from `VisualControlBase` it might be worth a good look to see if other controls that also have same base class exhibit this behaviour.

@KamaniAR is correct that `Validating` and `Validated` events are both fired twice.

Also this is a symptomatic fix (work-around) and does not remove the underlying cause.

Maybe @antosupersayan-vb is willing to do some more testing on other controls that use the same base as well...?

My £ 0.02

giduac:
I had a look at this to, and also found that this could be present in more controls as @KamaniAR already pointed out.
@KamaniAR is correct that `Validating` and `Validated` events are both fired twice.

Since `KTextBox` derives from `VisualControlBase` it might be worth having a good look at this and see if the other controls also derive from `VisualControlBase`.

Also this is a symptomatic fix (work-around) and does not remove the underlying cause.

My £ 0.02

PWagner1:
@KamaniAR & @giduac 

Good catch, working on it right now

PWagner1:
@KamaniAR & @giduac 

`VisualControlBase` is now fixed, please can you re-test? Thanks

giduac:
The error is still in there somewhere with this work-around.
I ran it through the debugger yesterday and in the path when the focus is lost this gets triggered twice somewhere.
When you design a new control based in this base class then it will need this work-around....

For me it's worth the effort to find the underlying cause and correct it. 
And be done with it for once and all...

PWagner1:
> The error is still in there somewhere with this work-around. I ran it through the debugger yesterday and in the path when the focus is lost this gets triggered twice somewhere. When you design a new control based in this base class then it will need this work-around....
> 
> For me it's worth the effort to find the underlying cause and correct it. And be done with it for once and all...

@giduac 

Hold on, may have something... Let me just finish off #2156 (I've got a lot of stuff in the TEST repo that I've been playing with over the last few weeks).

giduac:
@PWagner1
cc: @KamaniAR

I've had another look at this.
And this is bigger than it seems at first.
I will dig into this later on when there's more time.

For now let's keep your fix to VisualControlBase.
Please add a comment tag including the ticket number to the changes so we can easily find them.

PWagner1:
> @PWagner1 cc: @KamaniAR
> 
> I've had another look at this. And this is bigger than it seems at first. I will dig into this later on when there's more time.
> 
> For now let's keep your fix to VisualControlBase. Please add a comment tag including the ticket number to the changes so we can easily find them.

@giduac 

 Will do, I'll also add @Krypton-Suite/reviewers as a assignee to the bug.

PWagner1:
@giduac & @KamaniAR 

I've now put in `TODO` comments to the relevant parts of the code.

giduac:
@KamaniAR 
If you're OK with this then please approve the PRS for 2801 ....
Thanks

PR #2871, opened 2026-01-11
* `SaveConfigToArray` and `LoadConfigFromArray` not working correctly

### LINKED ISSUE #2516: [Bug]: `SaveConfigToArray` and `LoadConfigFromArray` not working correctly

PWagner1:
* Resolved #2516 for V105 LTS

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -5,6 +5,7 @@
 # 2026-01-19 - Build 2501 (Version 105-LTS - Patch 1) - January 2026
 
 * Resolved [#2516](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2516), `SaveConfigToArray` and `LoadConfigFromArray` not working correctly
+* Implemented [#762](https://github.com/Krypton-Suite/Standard-Toolkit/issues/762), Does Krypton provide the correct UIA providers for the controls it "Hides"
 * Resolved [#948](https://github.com/Krypton-Suite/Standard-Toolkit/issues/948), Why does the KButton change size/padding between using .NET Framework & .NET
 * Resolved [#1809](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1809), Dragging a dock cell appears to lag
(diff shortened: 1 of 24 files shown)


Source/Krypton Components/TestForm/DockingConfigSaveLoadTest.cs
```diff
@@ -0,0 +1,316 @@
+﻿#region BSD License
+/*
+ * 
+ *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
+ *  Modifications by Peter Wagner(aka Wagnerp) & Simon Coghlan(aka Smurf-IV), et al. 2024 - 2026. All rights reserved.
```
Smurf-IV: Wrong start year

PWagner1: Updated year

Issue #2882, opened 2026-01-11
[Feature Request]: `KryptonPrintPreviewDialog` - Part of #2658


PWagner1:
### Feature Title

`KryptonPrintPreviewDialog` - Part of #2658

### Feature Description

To implement a `KryptonPrintPreviewDialog`

PR #2883, opened 2026-01-12
* `KryptonPrintPreviewDialog` - Part of #2658

PWagner1:
* Implements #2882


PWagner1:
> This needs to have an Example created in the "Examples repo" please

Will do :), did a lot on that repo yesterday as it was a rainy day...

PR #2884, opened 2026-01-12
* Implements 'Capsule' shape

PWagner1:
* Implements 'Capsule' shape

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Values/Badge/BadgeColorValues.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Values/Badge/BadgeColorValues.cs
@@ -54,7 +54,7 @@ public class BadgeColorValues : Storage
     [Category(@"Visuals")]
     [Description(@"The background color of the badge.")]
     [RefreshProperties(RefreshProperties.All)]
-    [KryptonDefaultColor]
+    [DefaultValue(typeof(Color), "Red")]
     public Color BadgeColor
     {
         get => _badgeColor;
@@ -76,7 +76,7 @@ public class BadgeColorValues : Storage
     [Category(@"Visuals")]
     [Description(@"The text color of the badge.")]
     [RefreshProperties(RefreshProperties.All)]
-    [KryptonDefaultColor]
+    [DefaultValue(typeof(Color), "White")]
     public Color TextColor
     {
         get => _textColor;


lesandrog:
This is almost good, the capsule would need a minimum width of 22px, so it doesn't look strange when it only has one character.
image

lesandrog:
When the border value is higher, along with BorderBevel Raised and Inset, it's being shifted towards the center. Is that how it's supposed to be?
image

PWagner1:
> When the border value is higher, along with BorderBevel Raised and Inset, it's being shifted towards the center. Is that how it's supposed to be?

Try now...

lesandrog:
With a 6px border, Raised and Inset are ending up with double the size at 12px.
image

lesandrog:
In RoundedRectangle and Capsule, the size is not increased with BadgeDiameter. (80)
image

giduac:
Maybe I'm incorrect but as far as I can see the padding is applied to the outside of the image.
In that case padding should be a margin

For reference Padding versus Margin:
https://learn.microsoft.com/en-us/dotnet/desktop/winforms/controls/windows-forms-controls-padding-autosize

lesandrog:
I did some testing and here are some suggestions for improvement:

- Adjusted the minimum size of the circle and square; when a diameter smaller than the text size was defined, the text was being cut off.
- Adjusted the rounded rectangle and capsule; apply the larger diameter when defined.
- Borders drawn outwards.
- BorderBevel none, now using the same drawing method to maintain the same appearance and measurements.
- Adjusted the border width (it's 99% the correct size. To be 100%, the Pen should be 1, but this generates a color jaggedness on the curves, so it's set to 2).
- Added 1px to some border lines to avoid empty spaces in wider borders.

[ViewDrawBadge.txt](https://github.com/user-attachments/files/24715091/ViewDrawBadge.txt)

mcpbcs:
Fantastic feature!

PWagner1:
@lesandrog & @giduac 

Please could you retry with the latest changes?

lesandrog:
Sometimes, when I copy the button with the badge, the background, border, and text colors become empty in the copy.
I haven't been able to determine exactly when this happens.

*Edit:
It seems to me that the default three color setting has an incomplete definition, but I could be wrong.

The rest seems fine to me!

PWagner1:
> Sometimes, when I copy the button with the badge, the background, border, and text colors become empty in the copy. I haven't been able to determine exactly when this happens.
> 
> *Edit: It seems to me that the default three color setting has an incomplete definition, but I could be wrong.

Hi @lesandrog 

It should be fixed now, the problem was that it was using `KryptonDefaultColor` rather than using actual colours as the default colour.

lesandrog:
For me, it's fully functional.

PWagner1:
@giduac & @Smurf-IV 

Ready to go?

Issue #2886, opened 2026-01-13
[Feature Request]: Jump list support


PWagner1:
### Feature Title

Jump list support

### Feature Description

Jump lists were introduced in Windows 7. It would be ideal to have native support, so that developers don't have to jump through hoops. The 'bare bones' was already introduced with PR #2835, so it's just completing the feature.

### Implementation Ideas (Optional)

For V110+

PR #2889, opened 2026-01-14
* Jump list support

PWagner1:
* Implements #2886  (Please have a play around with this)


Smurf-IV:
Don;t forget to add a task for the Demo Repo

Issue #187, opened 2021-07-07
[Feature Request]: Can the Extended Kit Scrollbars be placed into standard and used by all controls that have scrollability ?


Smurf-IV:
**Is your feature request related to a problem? Please describe.**
Scroll bars need to be themed to. and they should have been done from day one.
The Extended kit has them, so why not move into standard and use by default for all controls that have scrollbars ?

Wagnerp:
I can move them, since they don't need to be in their own module, but I don't think WinForms supports scrollbar overriding yet.

Smurf-IV:
> 
> 
> I can move them, since they don't need to be in their own module, 

Ok

>but I don't think WinForms supports scrollbar overriding yet.

Can you give more information?
Does this mean that the standard scrollbars used by current Winform Components cannot be coloured ??
Hows does windows do it? 
and does WPF allow it?

Wagnerp:
I have posted a query about this in the WinForms repo back in October, as I had the same idea. You can read it [here](https://github.com/dotnet/winforms/issues/2054). The team said that they're looking into it.

PWagner1:
@Smurf-IV 

PR #2899 should allow for this, you couldn't do it with the 'old' `KryptonScrollBar`.

PR #2917, opened 2026-01-23
* Can the Extended Kit Scrollbars be placed into standard and used by…

PWagner1:
… all controls that have scrollability?

* Implemented #187

image


PWagner1:
> How do these deal with RTL Layout ?

Needs testing, they're turned on/off globally via KManager rather than per control basis.

Smurf-IV:
> > How do these deal with RTL Layout ?
> 
> Needs testing, they're turned on/off globally via KManager rather than per control basis.

Yes, but they are now "On" by default for all new implementations ....

Issue #2935, opened 2026-01-28
[Bug]: Maximized MDI window form border is drawn on the second monitor.


mcpbcs:
### Summary

There is an invalid border drawn in the secundary monitor

### Description

![Image](https://github.com/user-attachments/assets/92efb9b7-c599-43c5-8ad8-ead0e569107e)

<img width="1356" height="512" alt="Image" src="https://github.com/user-attachments/assets/eb576cc8-f829-428a-b338-f5caf66145be" />

[WinAppBorderWinDrawInMonitor1.zip](https://github.com/user-attachments/files/24922670/WinAppBorderWinDrawInMonitor1.zip)

In the photo, the monitor on the left is my primary display, and the monitor on the right is my secondary display. When I maximize my MDI form on the second monitor, the form itself appears correctly on the right monitor, but its border/frame is incorrectly drawn on the left (primary) monitor.

### Steps to Reproduce

Sample project added.

### Expected Behavior

No border should be drawn

### Actual Behavior

A border is drawn in monitor 1 (not primary monitor)

### Operating System

Windows

### OS Version

10

### Framework/.NET Version

4.8

### Toolkit Version

110.26.1.28-alpha

### Areas Affected

Toolkit

PWagner1:
Hi @mcpbcs 

Is this working?

mcpbcs:
Hi @PWagner1 

The issue is still present in version  package id="Krypton.Toolkit.Nightly" version="110.26.2.40-alpha" targetFramework="net48"

PWagner1:
See #3013

PR #2957, opened 2026-01-31
* Maximized MDI window form border is drawn on the second monitor

PWagner1:
# Fix maximized MDI window form border drawn on wrong monitor (#2935)

## Summary

Resolves [#2935](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2935): when an MDI child (KryptonForm) is maximized on a secondary monitor, the form border/frame was incorrectly drawn on the primary (left) monitor instead of on the same monitor as the form content.

### 1. **DropSolidWindow** (Krypton.Navigator)

- **Cause:** In the `SolidRect` setter, position was computed as `value.Location - (Size)area.Location` with `area = Screen.GetWorkingArea(this)`. When the drop-indicator window was already on a secondary monitor, `area` was that monitor’s working area, so the result was relative to that screen. That rectangle was then assigned to `DesktopBounds`, which is in **screen** coordinates, so the window was drawn on the primary monitor.
- **Fix:** Set `DesktopBounds = value` directly. `SolidRect` / `DrawRect` are already in screen coordinates, so no conversion is needed.

**File:** `Source/Krypton Components/Krypton.Navigator/Dragging/DropSolidWindow.cs`

### 2. **VisualForm non-client (border) painting** (Krypton.Toolkit)

- **Cause:** Non-client (border) painting used a single cached DC created with `CreateCompatibleDC(IntPtr.Zero)`, i.e. compatible with the primary monitor. For a maximized MDI child on a secondary monitor, drawing through that DC could tie the border to the primary monitor.
- **Fix:** In `OnNonClientPaint`, create a compatible DC from the window’s DC: `memDC = CreateCompatibleDC(hDC)` (where `hDC` is from `GetWindowDC(Handle)`). Use `memDC` for the offscreen bitmap and `WindowChromePaint`, then BitBlt from `memDC` to `hDC`. In `finally`, restore the previous bitmap, delete the bitmap, and call `DeleteDC(memDC)` when non-zero. If `CreateCompatibleDC(hDC)` fails, fall back to the existing cached `_screenDC` behaviour.

**File:** `Source/Krypton Components/Krypton.Toolkit/Controls Visuals/VisualForm.cs`

### 3. **TestForm demo** (Bug 2935 MDI multi-monitor)

- Added a dedicated demo to reproduce and verify the fix:
  - **Bug2935MdiMultiMonitorDemo** – MDI parent with instructions, “New MDI Child”, “Move to secondary monitor” (when multiple screens), and Tile/Cascade/Arrange icons.
  - Instructions explain: move window to second monitor → open MDI child → maximize → confirm border is on the same monitor as the form.
- Registered in StartScreen as **“Bug 2935 MDI multi-monitor”**.

**Files:**
`Source/Krypton Components/TestForm/Bug2935MdiMultiMonitorDemo.cs` `Source/Krypton Components/TestForm/Bug2935MdiMultiMonitorDemo.Designer.cs` `Source/Krypton Components/TestForm/Bug2935MdiMultiMonitorDemo.resx` `Source/Krypton Components/TestForm/StartScreen.cs`

### 4. **Changelog**

- Added a “Resolved” entry for [#2935](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2935) in the 2026-11-xx (V110 Nightly) section.

**File:** `Documents/Changelog/Changelog.md`

## How to verify

1. Build and run TestForm (e.g. from the solution or `dotnet run` for TestForm).
2. On the start screen, open **“Bug 2935 MDI multi-monitor”**.
3. **With two monitors:**
   - Click **“Move to secondary monitor”** (or drag the window to the second display).
   - Click **“New MDI Child”**.
   - Maximize the MDI child (maximize button or double-click title bar).
   - **Expected:** The form border is drawn on the **same** monitor as the form content (secondary). No border appears on the primary monitor.
4. **Optional:** In an app that uses drag-drop with a solid drop indicator (e.g. Navigator), drag on the secondary monitor and confirm the drop-indicator window appears on that monitor, not on the primary.

## Checklist

- [x] Fixes [#2935](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2935)
- [x] No new compiler warnings from these changes
- [x] Changelog updated
- [x] Demo added in TestForm for manual verification


Source/Krypton Components/TestForm/Bug2935MdiMultiMonitorDemo.cs
```diff
@@ -0,0 +1,99 @@
+#region BSD License
+/*
+ *
+ *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
+ *  Modifications by Peter Wagner (aka Wagnerp), Simon Coghlan (aka Smurf-IV), et al. 2017 - 2026. All rights reserved.
```
Smurf-IV: wrong start year

Smurf-IV (changes requested):
Does this need to go into the other repos (i.e. 85 upwards ?)

Source/Krypton Components/TestForm/Bug2935MdiMultiMonitorDemo.cs
PWagner1: Fixed

PWagner1:
> Does this need to go into the other repos (i.e. 85 upwards ?)

Will do.

PR #3027, opened 2026-02-11
* Add support for .NET 11

PWagner1:
* Add support for .NET 11


Smurf-IV (approved):
Will the "Developer MC requirements" doc also be updated
And the supported .Net table ?

PWagner1:
> Will the "Developer MC requirements" doc also be updated And the supported .Net table ?

Yes, in the middle of updating them now, the workflows will also need to be modified.

PR #3096, opened 2026-02-21
* Tooltips with extended/infinite timeout (V110)

PWagner1:
* Add missing events that allows custom values to be used (no more hardcoded values).
* Documentation

### DIFF
--- a/Source/Krypton Components/Krypton.Utilities/Components/KryptonCheckSum/Controls Toolkit/KryptonComputeFileCheckSum.cs
+++ b/Source/Krypton Components/Krypton.Utilities/Components/KryptonCheckSum/Controls Toolkit/KryptonComputeFileCheckSum.cs
@@ -1,50 +1,16 @@
-#region BSD License
-/*
- *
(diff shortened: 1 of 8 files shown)


Source/Krypton Components/Krypton.Utilities/Components/KryptonCheckSum/Controls Toolkit/KryptonComputeFileCheckSum.cs
```diff
@@ -1,16 +1,50 @@
+#region BSD License
```
Smurf-IV: Move these into a separate PR please ?

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Utilities/Components/KryptonCheckSum/Controls Toolkit/KryptonComputeFileCheckSum.cs
PWagner1: Yes, sure.

PR #3108, opened 2026-02-22
* Colors in theme exporting XMLs is missing (V105 LTS)

### LINKED ISSUE #3101: [Bug]: Colors in theme exporting XMLs is missing
### Summary

Color values are not being serialized into the generated Krypton Palette XML file

### Description

When exporting a custom KryptonPalette using the Krypton Toolkit serializer, the resulting XML file does not include the serialized color values. Although the palette is configured with specific colors (e.g., for buttons, backgrounds, and borders), the generated XML frequently contains empty or default entries such as:

### Steps to Reproduce

[ThemeColoresNotBeingSerialized.zip](https://github.com/user-attachments/files/25462163/ThemeColoresNotBeingSerialized.zip)

### Expected Behavior

The colors values must be serialized

### Actual Behavior

Missing the values

### Operating System

Windows 10

### OS Version

_No response_

### Framework/.NET Version

4.8

### Toolkit Version

Krypton.Toolkit.Nightly version= 110.26.2.52-alpha

### Additional Information

Image

### Areas Affected

Toolkit

PWagner1:
# Fix: Colors missing from exported KryptonPalette XML (#3101)

Closes #3101

## Summary

When exporting a `KryptonCustomPaletteBase` (custom palette) to XML, configured color values were either silently dropped or written as `Value=""` (empty), making the exported file unusable. This was caused by a combination of bugs in the serialization logic and missing/incorrect `DefaultValueAttribute` decorations across several palette classes.

## Problem

Exporting a `KryptonPalette` via the designer or `Export()` API produced XML entries such as:

```xml
<Color1 Type="Color" Value="" />
<TextColor Type="Color" Value="" />
```

…or the properties were omitted entirely, even when explicit colors had been set.

## Root Causes

The `ExportObjectToElement` method in `KryptonCustomPaletteBase` uses reflection to walk the palette object graph. For leaf properties decorated with `[KryptonPersist(false)]`, it checks `DefaultValueAttribute` to determine whether a value equals its default (and can therefore be skipped when `ignoreDefaults=true`). Several bugs prevented this from working correctly:

1. **Overly strict attribute count check** – The condition `if (defaultAttribs.Length == 1)` silently skipped the default-value comparison whenever a property had zero *or* more than one `DefaultValueAttribute`-derived attributes (e.g., both `[KryptonDefaultColor]` and a redundant `[DefaultValue]`). Changed to `>= 1`.

2. **`[DefaultValue(typeof(Color), "")]` stores `null`** – `PaletteRibbonText.TextColor` used an empty string as the type-converter input. `ColorConverter` cannot parse `""`, so the attribute stored `null` instead of `Color.Empty`. The comparison `null.Equals(Color.Empty)` is skipped, meaning `Color.Empty` was always written as `Value=""`.

3. **Missing `[KryptonPersist(false)]` on `PaletteRibbonBack`** – `BackColor1` and `BackColor5` had no `[KryptonPersist]` attribute at all, so they were silently skipped by the serializer and never exported.

4. **Missing `[KryptonDefaultColor]` on `PaletteRibbonGeneral`** – All 18 color properties (`ContextTextColor`, `DisabledDark`, `DisabledLight`, `GroupDialogDark`, `GroupDialogLight`, `DropArrowDark`, `DropArrowLight`, `GroupSeparatorDark`, `GroupSeparatorLight`, `MinimizeBarDarkColor`, `MinimizeBarLightColor`, `TabRowBackgroundSolidColor`, `TabRowBackgroundGradientRaftingDarkColor`, `TabRowBackgroundGradientRaftingLightColor`, `TabSeparatorColor`, `TabSeparatorContextColor`, `QATButtonDarkColor`, `QATButtonLightColor`) were missing a `DefaultValueAttribute`, causing `Color.Empty` to always be written as `Value=""`.

5. **Missing `[DefaultValue]` on `PaletteRibbonFileAppTab`** – The three app-button color properties (`RibbonFileAppTabBottomColor`, `RibbonFileAppTabTopColor`, `RibbonFileAppTabTextColor`) lacked `DefaultValueAttribute`, so their non-empty defaults could never be detected and they were always exported even when unchanged.

6. **`KryptonPaletteFont.IsDefault` reference comparison bug** – `IsDefault` compared `Font` objects with `==` (reference equality), meaning it always returned `false`. The result was that the font palette node would always be exported, even when fonts were at their defaults. Fixed with a proper value-based `FontsAreDefault` helper.

### `KryptonCustomPaletteBase.cs`
- `ExportObjectToElement`: changed `defaultAttribs.Length == 1` → `>= 1` so that `KryptonDefaultColor` (a `DefaultValueAttribute` subclass) is correctly recognized even when it is the only attribute present.
- `ResetObjectToDefault`: same `== 1` → `>= 1` fix for consistency.
- `PaletteName` on export/import: now uses `Path.GetFileNameWithoutExtension` rather than the raw file path.

### `PaletteRibbonText.cs`
- `TextColor`: replaced `[DefaultValue(typeof(Color), "")]` with `[KryptonDefaultColor]`. The empty string failed to parse via `ColorConverter`, resulting in a `null` stored default.

### `PaletteRibbonBack.cs`
- `BackColor1`: added missing `[KryptonPersist(false)]` and `[DefaultValue(typeof(Color), "Empty")]`.
- `BackColor5`: added missing `[KryptonPersist(false)]`.

### `PaletteRibbonGeneral.cs`
- Added `[KryptonDefaultColor]` to all 18 color properties that default to `Color.Empty`.
- Added `[DefaultValue(typeof(Color), "Transparent")]` to `TabRowBackgroundGradientFirstColor` (default is `Color.Transparent`).

### `PaletteRibbonFileAppTab.cs`
- Added `[DefaultValue(typeof(Color), "31, 72, 161")]` to `RibbonFileAppTabBottomColor`.
- Added `[DefaultValue(typeof(Color), "84, 158, 243")]` to `RibbonFileAppTabTopColor`.
- Added `[DefaultValue(typeof(Color), "White")]` to `RibbonFileAppTabTextColor`.

### `KryptonPaletteFont.cs`
- `IsDefault`: replaced broken `new Font(...) ==` reference comparison with a `FontsAreDefault` helper that compares font name, size, style, and unit by value.

## Testing

1. Create a new `KryptonCustomPaletteBase` in the designer or at runtime.
2. Set one or more color properties (e.g., `Palette.ButtonStyles.ButtonCommon.StateCommon.Back.Color1 = Color.Red`).
3. Export the palette to XML via **right-click → Export…** or the `Export()` API.
4. Verify the exported XML contains the set color (`Value="Red"`) and does **not** contain `Value=""` entries for unset properties.
5. Re-import the XML and confirm the colors are restored correctly.
6. Repeat with Ribbon-specific color properties (`RibbonGeneral`, `RibbonAppMenuDocsTitle`, ribbon back colors, etc.).

## Linked Issues

- Fixes #3101

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -4,6 +4,7 @@
 
 ## 2026-04-20 - Build 2604 (Version 105-LTS - Patch 2) - April 2026
 
+* Resolved [#3101](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3101), Colors in theme exporting XMLs is missing
 * Resolved [#3103](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3103), Name of the theme is not being serialized in XML files
 * Resolved [#3031](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3031), Maximized form's size exceeds the screen's working area
 * Implemented [#3075](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3075), Tooltips with extended/infinite timeout
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
@@ -3527,8 +3527,8 @@ public class KryptonCustomPaletteBase : PaletteBase
                         {
                             var defaultAttribs = prop.GetCustomAttributes(typeof(DefaultValueAttribute), false);
 
-                            // Does this property have a default value attribute?
-                            if (defaultAttribs.Length == 1)
+                            // Use the first one found (KryptonDefaultColor is a DefaultValueAttribute subclass)
+                            if (defaultAttribs.Length >= 1)
                             {
                                 // Cast to correct type
                                 var defaultAttrib = defaultAttribs[0] as DefaultValueAttribute;
@@ -3691,8 +3691,8 @@ public class KryptonCustomPaletteBase : PaletteBase
                             {
                                 var defaultAttribs = prop.GetCustomAttributes(typeof(DefaultValueAttribute), false);
 
-                                // Does this property have a default value attribute?
-                                if (defaultAttribs.Length == 1)
+                                // Does this property have at least one default value attribute?
+                                if (defaultAttribs.Length >= 1)
                                 {
                                     // Cast to correct type
                                     var defaultAttrib = (DefaultValueAttribute)defaultAttribs[0];
--- a/Source/Krypton Components/Krypton.Toolkit/Palette Base/PaletteRibbon/PaletteRibbonBack.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Palette Base/PaletteRibbon/PaletteRibbonBack.cs
@@ -105,8 +105,10 @@ public class PaletteRibbonBack : Storage,
     /// <summary>
     /// Gets and sets the first background color for the ribbon item.
     /// </summary>
+    [KryptonPersist(false)]
     [Category(@"Visuals")]
     [Description(@"First background color for the ribbon item.")]
+    [DefaultValue(typeof(Color), "Empty")]
     [RefreshProperties(RefreshProperties.All)]
     public virtual Color BackColor1
     {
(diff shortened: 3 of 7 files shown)


PWagner1:
> This should go down to 85 as well

Will do.

PR #3110, opened 2026-02-22
* Controls should have "min design heights" be the same (V105 LTS)

### LINKED ISSUE #615: [Feature Request]: Controls should have "min design heights" be the same.

PWagner1:
# Implement MinimumControlHeight for Input Controls

Closes #615

## Summary

Input controls (`KryptonComboBox`, `KryptonTextBox`, `KryptonNumericUpDown`, `KryptonDateTimePicker`, `KryptonMaskedTextBox`, `KryptonDomainUpDown`) derive their height entirely from the active palette font size, with small per-control rendering differences causing pixel-level misalignment when controls are placed side-by-side on a form. This PR introduces a `MinimumControlHeight` property that allows developers to set a consistent design height across all input controls, independent of the active palette font.

### `VisualControlBase.cs`

- Added `private int _minimumControlHeight` backing field (default `0`).
- Added `public virtual int MinimumControlHeight` property to the `#region Public` section:
  - Category: **Layout**
  - Default value: `0` (disabled — no change to existing behaviour)
  - Calling `PerformLayout()` on change ensures immediate resize in the designer and at runtime.
  - Marked `virtual` so derived controls can override if necessary.

### `KryptonComboBox.cs`

- Updated `GetPreferredSize` to clamp the computed height against `MinimumControlHeight` when greater than zero, after all existing `MinimumSize`/`MaximumSize` clamping.

### `KryptonTextBox.cs`

- Updated `GetPreferredSize` with the same `MinimumControlHeight` clamp.

### `KryptonNumericUpDown.cs`

- Added `private int _cachedHeight` field (initialised to `-1` in the constructor), bringing this control into line with the pattern used by all other input controls.
- Replaced the ad-hoc `SetBoundsCore` implementation (which called `GetPreferredSize` inline) with the standard cache-on-first-set / always-override-to-`PreferredHeight` pattern used by `KryptonComboBox`, `KryptonDateTimePicker`, and `KryptonDomainUpDown`.
- Updated `GetPreferredSize` with the same `MinimumControlHeight` clamp.

### `KryptonDateTimePicker.cs`

- Updated `GetPreferredSize` with the same `MinimumControlHeight` clamp.

### `KryptonMaskedTextBox.cs`

- Updated `GetPreferredSize` with the same `MinimumControlHeight` clamp.

### `KryptonDomainUpDown.cs`

- Updated `GetPreferredSize` with the same `MinimumControlHeight` clamp.

## Behaviour

- **Default (`MinimumControlHeight = 0`):** No change whatsoever. All existing layouts and palette-driven sizing continue to work identically.
- **Non-zero value:** The control height will never fall below the specified pixel value, regardless of the active palette font size. If the font-driven preferred height is already larger than `MinimumControlHeight`, the font-driven height wins.

## Usage Example

```csharp
// Align all input controls to a consistent 26px design height
kryptonComboBox1.MinimumControlHeight = 26;
kryptonTextBox1.MinimumControlHeight = 26;
kryptonNumericUpDown1.MinimumControlHeight = 26;
kryptonDateTimePicker1.MinimumControlHeight = 26;
kryptonMaskedTextBox1.MinimumControlHeight = 26;
kryptonDomainUpDown1.MinimumControlHeight = 26;
```

The property is fully visible in the Visual Studio designer under the **Layout** category for each control.

## Notes

- The `+4` offset used in `KryptonComboBox.InternalPanel.GetPreferredSize` (`Font.Height + 4`) was audited and left unchanged. It is a deliberate OS-level fudge factor for the inner Win32 ComboBox rendering, and altering it would break existing layouts. The new `MinimumControlHeight` property is the correct mechanism for reconciling per-control height differences.
- No palette schema changes, no `.ktheme` file changes, and no breaking changes to any public API.

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -5,6 +5,7 @@
 ## 2026-04-20 - Build 2604 (Version 105-LTS - Patch 2) - April 2026
 
 * Implemented [#615](https://github.com/Krypton-Suite/Standard-Toolkit/issues/615), Controls should have "min design heights" be the same
+* Resolved [#3101](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3101), Colors in theme exporting XMLs is missing
 * Resolved [#3103](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3103), Name of the theme is not being serialized in XML files
 * Resolved [#3031](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3031), Maximized form's size exceeds the screen's working area
 * Implemented [#3075](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3075), Tooltips with extended/infinite timeout
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
@@ -3527,8 +3527,8 @@ public class KryptonCustomPaletteBase : PaletteBase
                         {
                             var defaultAttribs = prop.GetCustomAttributes(typeof(DefaultValueAttribute), false);
 
-                            // Does this property have a default value attribute?
-                            if (defaultAttribs.Length == 1)
+                            // Use the first one found (KryptonDefaultColor is a DefaultValueAttribute subclass)
+                            if (defaultAttribs.Length >= 1)
                             {
                                 // Cast to correct type
                                 var defaultAttrib = defaultAttribs[0] as DefaultValueAttribute;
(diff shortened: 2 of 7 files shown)


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonNumericUpDown.cs
```diff
@@ -1822,22 +1829,20 @@ protected override void SetBoundsCore(int x, int y,
         int width, int height,
         BoundsSpecified specified)
     {
-        // Get the preferred size of the entire control
-        Size preferredSize = GetPreferredSize(new Size(int.MaxValue, int.MaxValue));
-
         // If setting the actual height
-        if (specified.HasFlag(BoundsSpecified.Height))
+        if ((specified & BoundsSpecified.Height) == BoundsSpecified.Height)
```
Smurf-IV, line 1837: Why was this changed,

And it can be combined with the next if as well?

PWagner1: Documented changes

PR #3136, opened 2026-02-28
* Docking: Loading configuration while form is hidden remover splitte…

### LINKED ISSUE #3124: [Bug]: Docking: Loading configuration while form is hidden remover splitter
### Summary

 Loading configuration from a file while form is hidden disables resizing of some docka spaces

### Description

Hide the form hosting a DockingManager, load docking configuration from file and then show the form.
You then can not resize some (but not all) dock spaces. If you float the problematic dock space and re-dock it, you are then able to resize that dock space.

### Steps to Reproduce

  ` //this as Form or KryptonForm
this.Hide();
kryptonDockingManager1.LoadConfigFromFile(dockSavePath);
this.Show();`

### Expected Behavior

You should be able to resize all dock spaces.

### Actual Behavior

Some dock spaces cant be resized.

### Operating System

Windows 111

### OS Version

_No response_

### Framework/.NET Version

8

### Toolkit Version

100.26.1.19 & 95.25.8.235

### Additional Information

https://github.com/user-attachments/assets/edf66f0a-ab01-456e-8d18-ce2def3f0319

### Areas Affected

Docking

PWagner1:
…r (V105 LTS)

### Summary

Loading a docking configuration (via `KryptonDockingManager.LoadConfigFromFile` / `LoadConfigFromStream`) while the host form is hidden causes some dock space separators to become unresizable after the form is shown. Floating the affected dock space and re-docking it restores resizing, which confirmed the issue was a stale initialisation state rather than a structural defect in the saved configuration.

### Root Cause

`KryptonSpace.UpdateVisible(bool focus)` is the method responsible for setting the correct `Visible` state on each `KryptonWorkspaceCell` inside a dockspace after pages are loaded, and for triggering `PerformLayout()` to reflect those changes. It works by posting an async `BeginInvoke` call so the update runs in sync with the WinForms message queue.

`BeginInvoke` requires a valid window handle (`HWND`). The guard condition for this was:

```csharp
if (!_awaitingVisibleUpdate && IsHandleCreated)
{
    BeginInvoke(_visibleUpdate);
    _awaitingVisibleUpdate = true;
}
```

When the form is hidden (`this.Hide()`), the dockspace controls have no HWND yet — `IsHandleCreated` is `false`. The entire `UpdateVisible` call was silently skipped. The cells therefore remained in an incorrect visibility state (invisible by default) after config load.

This broke separator resizing because `FindMovementRect` — which computes how far a separator can be dragged — calls `DockingHelper.InnerRectangle`, which only counts `child.Visible == true` controls. With cells invisible, the inner rectangle was computed as empty, producing a zero-dimension movement rectangle for the separator, making the dock space appear completely unresizable.

When you floated and re-docked, a fresh dockspace was constructed after `Show()` so `IsHandleCreated` was already `true`, which is why that workaround fixed it.

### Fix

**File:** `Krypton Components\Krypton.Docking\Control Docking\KryptonSpace.cs`

Two small changes:

1. **`UpdateVisible(bool focus)`** — when `IsHandleCreated` is `false`, instead of silently doing nothing, set a new `_pendingVisibleUpdateOnHandle` flag so the deferred request is remembered.

2. **`OnHandleCreated(EventArgs e)` (new override)** — when the HWND is eventually created (i.e. when the form becomes visible), check `_pendingVisibleUpdateOnHandle` and call `UpdateVisible` at that point. This queues the `BeginInvoke` correctly and the cells get their proper visibility and layout applied before the user can interact with the separators.

### Behaviour

- **No change to existing behaviour** when the form is already visible during config load — the `IsHandleCreated` path is taken exactly as before.
- **Fix applies** only when `IsHandleCreated` is `false` at load time; the pending flag defers the update until `OnHandleCreated` fires, which is the earliest safe point to call `BeginInvoke`.

### Manual Repro Steps

1. Create a form with a `KryptonDockingManager` and two or more docked panels.
2. Save the docking configuration to a file.
3. On next launch, call `this.Hide()`, then `kryptonDockingManager1.LoadConfigFromFile(path)`, then `this.Show()`.
4. **Before fix:** some dock spaces cannot be resized by dragging their splitter.
5. **After fix:** all dock spaces resize correctly.

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -4,6 +4,7 @@
 
 ## 2026-04-20 - Build 2604 (Version 105-LTS - Patch 2) - April 2026
 
+* Resolved [#3124](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3124), Docking: Loading configuration while form is hidden remover splitter
 * Resolved [#3123](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3123), `KryptonComboBox` DropDownWidth doesn't resize with the control
 * Implemented [#3117](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3117), Disabling focus when clicking on a `KryptonButton`
 * Implemented [#615](https://github.com/Krypton-Suite/Standard-Toolkit/issues/615), Controls should have "min design heights" be the same
--- a/Source/Krypton Components/Krypton.Docking/Control Docking/KryptonSpace.cs
+++ b/Source/Krypton Components/Krypton.Docking/Control Docking/KryptonSpace.cs
@@ -66,6 +66,7 @@ public abstract class KryptonSpace : KryptonWorkspace
     private readonly EventHandler _visibleUpdate;
     private bool _awaitingFocusUpdate;
     private bool _awaitingVisibleUpdate;
+    private bool _pendingVisibleUpdateOnHandle;
     private bool _setFocus;
     private string _closeTooltip;
     private string _pinTooltip;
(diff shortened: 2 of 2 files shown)


PWagner1:
> Can this also go into v85 please ?

Aye!

Issue #3013, opened 2026-02-08
[Bug]: Maximized form's size exceeds the screen's working area


ICloneableX:
### Summary

A maximized form will exceed the size of the screen's working area and be (somewhat) visible on the second screen

### Description

<img width="807" height="269" alt="Image" src="https://github.com/user-attachments/assets/327bb409-b196-40dd-96d7-f651937d25d1" />

<img width="375" height="110" alt="Image" src="https://github.com/user-attachments/assets/9ab466a8-4125-4bc8-b09a-32ca1e1a2f2a" />

### Steps to Reproduce

1. Maximize the form.
2. Check `this.Size` and compare it to `Screen.PrimaryScreen.WorkingArea.Size` and/or notice the black rectangle appearing on the second screen.

### Expected Behavior

The size of a maximized form should fit exactly in the screen's working area.

### Actual Behavior

The form exceeds the screen's working area when maximized.

### Operating System

Windows 10

### Framework/.NET Version

.NET 10

### Toolkit Version

100.26.1.19 and 110.26.2.38-alpha

### Areas Affected

None

mcpbcs:
Sample project 
[WinFormAllIssuesNF48.zip](https://github.com/user-attachments/files/25201452/WinFormAllIssuesNF48.zip)

PWagner1:
Hi @ICloneableX & @mcpbcs 

Does this still happen in build >= .50?

ICloneableX:
@PWagner1 

Yep, still the same.

mcpbcs:
Hi @ICloneableX,

Is this bug fixed for you?

mcpbcs:
HI @PWagner1 

Maybe I'm doing something wrong, but I still see a border when the Krypton form is maximized.
If there's anything I can do to help resolve this, just let me know — I'm happy to contribute

PWagner1:
> HI [@PWagner1](https://github.com/PWagner1)
> 
> Maybe I'm doing something wrong, but I still see a border when the Krypton form is maximized. If there's anything I can do to help resolve this, just let me know — I'm happy to contribute

Hi @mcpbcs 

[...]

PR #3148, opened 2026-03-02
Fix for #3013 : Maximized form's size exceeds the screen's working area

mcpbcs:
Fix for #3013 : This change restores the original implementation of UpdateRegionForMaximized.


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
```diff
@@ -2776,27 +2776,35 @@ private void PerformViewPaint(Graphics g, Rectangle rect)
...
+	// Fix for #3013 : This change restores the original implementation of UpdateRegionForMaximized.
+	private void UpdateRegionForMaximized()
+	{
+		if (MdiParent == null)
+		{
+			// Get the size of each window border
+			var xBorder = PI.GetSystemMetrics(PI.SM_.CXSIZEFRAME) * 2;
+			var yBorder = PI.GetSystemMetrics(PI.SM_.CYSIZEFRAME) * 2;
+
+			Padding padding = RealWindowBorders;
```
Smurf-IV: `padding` is not used

PWagner1:
Also, please update the licence header with your name/username.

PWagner1:
Fixed conflict

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
mcpbcs: Using only SM_CXSIZEFRAME / SM_CYSIZEFRAME is intentional. These values represent the real system resize‑frame thickness applied by Windows when the window is maximized. Using them ensures the region matches the actual non‑client frame and prevents phantom margins or hit‑test issues.
RealWindowBorders is correct for NCCALCSIZE, but not for shaping the maximized region.
This keeps the behavior consistent with the original fix in #3013

KamaniAR:
Result from : Bug 3013 Test Form

Screen size: Height = 1020, Width = 1920,
Form Location: Left = -9, Top = -9, Right = 1929, Bottom = 1029,
Form size: Height = 1038, Width = 1938

image

if any settings need to change tell me.

PWagner1:
Hi @mcpbcs 

Is this good to go as well as the V105 counterpart?

mcpbcs:
> Hi @mcpbcs
> 
> Is this good to go as well as the V105 counterpart?

Hi Peter, yes, everything looks good on this one and on the V105 counterpart as well.

PR #3150, opened 2026-03-03
V105-LTS #3013 : Maximized form's size exceeds the screen's working area 

mcpbcs:
// Fix for #3013 : This change restores the original implementation of UpdateRegionForMaximized.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -20,7 +20,7 @@ on:
 
 jobs:
   build:
-    runs-on: windows-latest
+    runs-on: windows-2025-vs2026
     if: github.event_name != 'pull_request' || (github.event.pull_request.head.repo.full_name == github.repository || github.event_name == 'workflow_dispatch')
 
     steps:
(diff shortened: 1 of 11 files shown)


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
```diff
@@ -2634,27 +2634,32 @@ private void PerformViewPaint(Graphics g, Rectangle rect)
...
-    }
-
-    private void UpdateBorderRegion(Region? newRegion)
+	// Fix for #3013 : This change restores the original implementation of UpdateRegionForMaximized.
+	private void UpdateRegionForMaximized()
+	{
+		if (MdiParent == null)
+		{
+			// Get the size of each window border
+			Padding padding = RealWindowBorders;
```
Smurf-IV, line 2643: padding is used in this version but not the other, Please decide which is the correct way to go?

PWagner1:
Also, please update the licence header with your name/username.

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
mcpbcs: Using only SM_CXSIZEFRAME / SM_CYSIZEFRAME is intentional. These values represent the real system resize‑frame thickness applied by Windows when the window is maximized. Using them ensures the region matches the actual non‑client frame and prevents phantom margins or hit‑test issues.
RealWindowBorders is correct for NCCALCSIZE, but not for shaping the maximized region.
This keeps the behavior consistent with the original fix in #3013

PR #3160, opened 2026-03-09
* JumpList Fixes

PWagner1:
## Summary

This PR fixes and completes the Windows taskbar Jump List integration for Krypton, aligns the implementation with the current codebase, and updates the demo and documentation so the feature is usable and testable on modern Windows versions.

## Changes

- **Core jump list implementation**
  - Wire `JumpListValues` into `VisualForm` so changes reliably trigger a shell refresh via `OnJumpListChanged`.
  - Ensure `OnJumpListChanged` is called when the form handle is created, so jump lists are applied once the window exists.
  - Add `JumpListValues.Apply()` to raise `JumpListChanged` when `UserTasks` / `Categories` are modified programmatically.
  - Set the process AppUserModelID via `SetCurrentProcessExplicitAppUserModelID` when `AppId` is assigned, matching Windows’ jump list requirements.
  - Correct COM interop definitions in `PlatformInvoke.cs`:
    - Use the proper `CustomDestinationList` and `ObjectCollection` CLSIDs.
    - Add `IPropertyStore`, `PROPERTYKEY`, `PROPVARIANT`, and `PKEY_Title` to set the visible title for `IShellLinkW` items.
    - Provide `TrySetShellLinkTitle` helper so links get a proper display title (or gracefully fall back when unavailable).

- **Shell link creation and validation**
  - Build `IShellLinkW` instances from `JumpListItem` with all required fields for jump list tasks:
    - Always call `SetPath`.
    - Always provide `SetArguments` (using `" "` when no arguments are needed to satisfy the API contract).
    - Always set an icon location (falling back to the executable path when `IconPath` is not supplied).
    - Set `Description` using `Description` or `Title` as a fallback.
  - Create `IObjectCollection` / `IObjectArray` instances for both user tasks and custom categories and pass them through `ICustomDestinationList`.

- **Error handling and compatibility**
  - Catch and ignore `COMException` with `REGDB_E_CLASSNOTREG` in `OnJumpListChanged` so environments without the shell component (or with broken registration) don’t crash.
  - Keep other exceptions routed through `KryptonExceptionHandler` to aid diagnostics while remaining non-fatal.

- **TestForm jump list demo**
  - Update the Jump List Test demo form to exercise the new API:
    - Demonstrate setting `AppId`, adding/clearing user tasks, adding custom categories, and toggling Recent/Frequent flags.
    - Use `JumpList.Apply()` where appropriate so changes are actually pushed to the shell.
  - Add an optional WPF bridge for the demo:
    - Enable `<UseWPF>true</UseWPF>` in `TestForm.csproj` and create a `System.Windows.Application` instance at startup.
    - Implement `SyncToWpfJumpList()` in `JumpListTest` that mirrors `JumpListValues` (`UserTasks` and `Categories`) into `System.Windows.Shell.JumpList` and calls `Apply()` on the WPF jump list.
    - Resolve short executable names (e.g. `calc.exe`, `notepad.exe`) to full system paths so they are valid jump list targets.
    - Assign `CustomCategory` for custom-category items so they appear under named sections in the taskbar menu.
    - Marshal the sync call onto the WPF dispatcher when required.
  - Keep the WPF bridge **demo-only**; the core library still uses the native COM path.

- **Documentation**
  - Rewrite `Documents/KryptonJumpList.md` to match the actual implementation:
    - Remove references to non-existent `KryptonJumpList`, `KryptonJumpListItem`, and `KryptonJumpListItems` types.
    - Document `JumpListItem`, `JumpListValues`, and the COM interop (`ICustomDestinationList`, `IShellLinkW`, `IObjectCollection`, `IPropertyStore`, etc.).
    - Explain how `VisualForm`/`KryptonForm` integrates jump lists, including handle-creation timing and the `JumpListChanged` event.
    - Provide updated usage examples for designer and programmatic configuration, including `AppId`, `UserTasks`, `Categories`, and `Apply()`.
    - Clarify behavior and prerequisites for shell-managed **Recent** and **Frequent** categories (OS privacy setting, Windows-managed data).

## Rationale

- Previous documentation referenced types that do not exist and an earlier design that was never fully implemented, making the feature effectively unusable.
- The jump list COM code was close but incomplete: AppUserModelID timing, shell link requirements, and COM interop details prevented items from actually appearing in the taskbar menu.
- The updated TestForm demo now serves as a reliable end-to-end validation tool, and the documentation now reflects what ships in the repository.

## Testing

- Manually tested the Jump List Test demo on Windows 11:
  - Verified that right-clicking the app’s taskbar icon after adding user tasks shows those tasks under the **Tasks** section.
  - Verified that custom categories (e.g. “Recent Files”, “Templates”) appear as separate headings with their configured items.
  - Verified that enabling/disabling `ShowFrequentCategory` / `ShowRecentCategory` affects the presence of the corresponding shell-managed sections, subject to system privacy settings.
  - Confirmed that environments without functional jump list COM registration fail gracefully (no crash; feature is simply unavailable).

## Notes / Limitations

- Shell-managed **Recent** and **Frequent** sections are ultimately controlled by Windows (privacy settings and OS tracking). The code can request these categories, but cannot force them to show when the OS does not provide data.
- ~~The WPF `System.Windows.Shell.JumpList` bridge is limited to the TestForm demo and is not required for consumers of the Krypton libraries.~~

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/General/PlatformInvoke.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/General/PlatformInvoke.cs
@@ -3961,6 +3961,7 @@ No 	                    No 	                    Show text only
     #region nt.dll
 
     [DllImport(Libraries.NtDll, SetLastError = true)]
+    [DefaultDllImportSearchPaths(DllImportSearchPath.System32)]
     internal static extern int RtlGetVersion(ref OSVERSIONINFOEX lpVersionInformation);
 
     #endregion
@@ -5368,6 +5369,8 @@ No 	                    No 	                    Show text only
 
     /// <summary>
     /// CLSID for ObjectCollection COM object (EnumerableObjectCollection).
+    /// Must use the class CLSID here, not the IObjectCollection interface IID (5632b1a4-e38a-400a-928a-d4cd63230295).
+    /// Using the interface GUID causes REGDB_E_CLASSNOTREG when instantiating—COM expects a creatable class, not an interface.
     /// </summary>
     [ComImport]
     [Guid("2d3468c1-36a7-43b6-ac24-d3f02fd9607a")]
--- a/Source/Krypton Components/TestForm/Program.cs
+++ b/Source/Krypton Components/TestForm/Program.cs
@@ -17,10 +17,12 @@ internal static class Program
 
 #if NETFRAMEWORK
     [DllImport("user32.dll")]
+    [DefaultDllImportSearchPaths(DllImportSearchPath.System32)]
     private static extern bool SetProcessDPIAware();
 #endif
 
     [DllImport("shell32.dll", CharSet = CharSet.Unicode, SetLastError = true)]
+    [DefaultDllImportSearchPaths(DllImportSearchPath.System32)]
     private static extern int SetCurrentProcessExplicitAppUserModelID(string appId);
 
     /// <summary>


Source/Krypton Components/Krypton.Toolkit/General/PlatformInvoke.cs
```diff
@@ -3961,7 +3961,7 @@ internal static extern bool RoundRect(IntPtr hdc, int nLeftRect, int nTopRect, i
     #region nt.dll
 
     [DllImport(Libraries.NtDll, SetLastError = true)]
-    internal static extern int RtlGetVersion(ref PI.OSVERSIONINFOEX lpVersionInformation);
+    internal static extern int RtlGetVersion(ref OSVERSIONINFOEX lpVersionInformation);
```
Smurf-IV, line 3965: Needs `DefaultDllImportSearchPaths`

Source/Krypton Components/Krypton.Toolkit/General/PlatformInvoke.cs
```diff
@@ -5358,15 +5367,129 @@ internal class CustomDestinationList
     }
 
     /// <summary>
-    /// CLSID for ObjectCollection COM object.
+    /// CLSID for ObjectCollection COM object (EnumerableObjectCollection).
     /// </summary>
     [ComImport]
-    [Guid("5632b1a4-e38a-400a-928a-d4cd63230295")]
+    [Guid("2d3468c1-36a7-43b6-ac24-d3f02fd9607a")]
```
Smurf-IV, line 5376: Explain why this number has changed, is it a winver thing?

Source/Krypton Components/TestForm/Program.cs
```diff
@@ -13,16 +13,32 @@ namespace TestForm;
 
 internal static class Program
 {
+    private const string JumpListAppId = "KryptonToolkit.JumpListTest";
+
 #if NETFRAMEWORK
     [DllImport("user32.dll")]
     private static extern bool SetProcessDPIAware();
```
Smurf-IV, line 21: These need `DefaultDllImportSearchPaths`

Source/Krypton Components/Krypton.Toolkit.JumpList/GlobalUsings.cs
```diff
@@ -0,0 +1,6 @@
+// Global using directives
```
Smurf-IV, line 10: Needs the license header

Source/Krypton Components/Krypton.Toolkit/General/PlatformInvoke.cs
PWagner1: Now uses `DefaultDllImportSearchPaths`

Source/Krypton Components/Krypton.Toolkit/General/PlatformInvoke.cs
PWagner1: Added comment explaining why CLSID needs to be used here.

Issue #3201, opened 2026-03-11
[Bug]: Fix potential memory leaks in KryptonDataGridViewRatingColumn


giduac:
### Summary

Some declarations are not using the 'using' directive, fix dispose and clone routines.

### Toolkit Version

V100 & V110 (V85 does not have a rating column)

### Additional Information

PRs in the works...

PR #3216, opened 2026-03-13
3201 v110 fix potential memory leaks in krypton data grid view rating column

giduac:
- Issue: #3201
- Fix using directives
- Update dispose and clone methods
- Some refactoring
- Correct a few typos
- And the changelog

image


giduac:
@Smurf-IV 
cc: @Krypton-Suite/reviewers 

Saw your comments a bit late on the previous pr.
Have another look as the comments have been addressed and some more refactoring been done.

Issue #3227, opened 2026-03-18
[Bug]: `KryptonDockingManager.LoadConfigFromArray` throws exception


OJacot-Descombes:
### Summary

Attempt to load a docking config saved before throws an ArgumentException

### Description

My code:
```
_dockingManager.LoadConfigFromArray(Convert.FromBase64String(text));
```
This throws the following exception:

> System.ArgumentException: 'Cannot set the 'Space' property more than once.
Parameter name: SpaceControl'
>   at Krypton.Docking.KryptonDockingSpace.set_SpaceControl(KryptonSpace value)
   at Krypton.Docking.KryptonDockingDockspace.LoadElementFromXml(XmlReader xmlReader, KryptonPageCollection pages)
   at Krypton.Docking.DockingElement.LoadElementFromXml(XmlReader xmlReader, KryptonPageCollection pages)
   at Krypton.Docking.DockingElement.LoadElementFromXml(XmlReader xmlReader, KryptonPageCollection pages)
   at Krypton.Docking.DockingElement.LoadElementFromXml(XmlReader xmlReader, KryptonPageCollection pages)
   at Krypton.Docking.KryptonDockingControl.LoadElementFromXml(XmlReader xmlReader, KryptonPageCollection pages)
   at Krypton.Docking.DockingElement.LoadElementFromXml(XmlReader xmlReader, KryptonPageCollection pages)
   at Krypton.Docking.KryptonDockingManager.LoadConfigFromXml(XmlReader xmlReader)
   at Krypton.Docking.KryptonDockingManager.LoadConfigFromStream(Stream stream)
   at Krypton.Docking.KryptonDockingManager.LoadConfigFromArray(Byte[] buffer)
   at my application

### Steps to Reproduce

The base64 UTF-16 encoded config is:
```
[...]

PR #3250, opened 2026-03-23
* `KryptonDockingManager.LoadConfigFromArray` throws exception (V110)

PWagner1:
# Fix KryptonDockingManager.LoadConfigFromArray throws exception (#3227)

## Summary

Resolves an `ArgumentException` thrown when loading a docking configuration via `LoadConfigFromArray` in certain scenarios where a dockspace ends up with no pages after load.

## Problem

When calling `KryptonDockingManager.LoadConfigFromArray()`, the following exception could occur:

```
System.ArgumentException: 'Cannot set the 'Space' property more than once.'
Parameter name: SpaceControl'
  at Krypton.Docking.KryptonDockingSpace.set_SpaceControl(KryptonSpace value)
  at Krypton.Docking.KryptonDockingDockspace.LoadElementFromXml(XmlReader xmlReader, KryptonPageCollection pages)
  ...
```

This happened when:
- Loading a saved docking configuration where some pages referenced in the config were not present in the `pages` collection passed to the loader
- The workspace structure (cells, groups) was restored, but no pages could be matched
- Result: a dockspace with `PageCount == 0`
- The code disposed the empty dockspace and attempted `SpaceControl = null!` to clear the reference
- The `SpaceControl` setter in `KryptonDockingSpace` only allowed a single assignment and threw on any subsequent call (including when clearing with null)

## Solution

Modified the `SpaceControl` setter in `KryptonDockingSpace` to allow clearing the reference when passed `null`:

- **When `value == null`**: Unhook from the current control's events, set `_space = null`, and return (no throw)
- **When `value` is non-null**: Existing behavior unchanged — assignment allowed only once; subsequent non-null assignments still throw

This supports the "clear after dispose" case used when an empty dockspace is removed during config load.

## Changes

| File | Description |
|------|-------------|
| `Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingSpace.cs` | Updated `SpaceControl` setter to handle `null` for clearing the reference |

## Testing

1. Build the solution (e.g. via `run.cmd` or `Scripts/VS2022/build-stable.cmd`)
2. Load a saved docking config using `LoadConfigFromArray(Convert.FromBase64String(configText))` where:
   - The config was saved with a prior session, or
   - Some referenced pages are not in the current `pages` collection
3. Verify the config loads successfully without throwing
4. Run TestForm and validate docking save/load scenarios

## Related

- Fixes [#3227](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3227)
- Area: Docking

image

Warnings unrelated to this PR.


Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingSpace.cs
```diff
@@ -812,9 +812,9 @@ protected KryptonSpace? SpaceControl
             _space = value;
 
             // Hook into space events we need to monitor
-            SpaceControl.Disposed += OnSpaceDisposed;
-            SpaceControl.WorkspaceCellAdding += OnSpaceCellAdding;
-            SpaceControl.PageDrop += RaiseSpacePageDrop;
+            value.Disposed += OnSpaceDisposed;
```
Smurf-IV: please use the `_space ` just to make it clear that it is being done to the local (Yes I know they are the same thing!!)

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Docking/Elements Impl/KryptonDockingSpace.cs
PWagner1: Now fixed

Issue #3249, opened 2026-03-22
[Bug]: KryptonForm shows an extra border on the secondary monitor when maximized


mcpbcs:
### Summary

When a KryptonForm is maximized on a multi-monitor setup, an unexpected border appears partially on the secondary monitor. This does not happen with a standard WinForms Form and seems specific to KryptonForm.

### Description

In a multi-monitor environment, maximizing a KryptonForm causes a visible border or shadow area to extend into the secondary monitor. This creates a visual glitch where part of the window frame appears outside the intended screen.
The issue is especially noticeable when the primary monitor is on the left and the secondary monitor is positioned on the right (or vice versa). The form appears to calculate its maximized bounds incorrectly.
This behavior does not occur when using a standard WinForms Form, which suggests that the issue may be related to how KryptonForm handles window bounds, non-client areas, or screen working areas when maximized.

### Steps to Reproduce

Set up Windows with two monitors.
Create a WinForms application using Krypton Toolkit.
Use KryptonForm as the main form.
Run the application.
Maximize the KryptonForm on the primary monitor.
Observe the edge/border of the form near the secondary monitor.

### Expected Behavior

When maximized, the KryptonForm should:

Fully occupy the working area of the current monitor.
Not extend into or affect the secondary monitor.
Behave visually the same as a standard WinForms Form.

### Actual Behavior

A visible border of the KryptonForm appears on the secondary monitor.
The maximized window seems to exceed the bounds of the active monitor.
This results in an incorrect visual layout when working with multiple monitors.

### Operating System

10

### Framework/.NET Version

4.8

### Toolkit Version

Nightly

### Areas Affected

Toolkit

mcpbcs:
<img width="874" height="841" alt="Image" src="https://github.com/user-attachments/assets/b0227009-4356-46b6-aca2-5b0dadd22408" />

I believe the issue comes from how border sizes are evaluated in the BorderWidths(FormBorderStyle) method.
When UseThemeFormChromeBorderWidth is enabled, the method always returns (0, 0) and completely skips evaluating the actual FormBorderStyle and system metrics:


if (UseThemeFormChromeBorderWidth)
{
    xBorder = 0;
    yBorder = 0;
}

This means that even for styles like Sizable or FixedSingle, no border width is accounted for.
[...]

Issue #3164, opened 2026-03-09
[Bug]: Font property values are being serialized depending of the current culture in exported XML theme file


mcpbcs:
### Summary

Font property values are not being serialized in the exported XML File

### Description

The Font property value appers like "(None)" in english or "(Ninguno)" in spanish.

### Steps to Reproduce

Export a theme:

> this.saveFileDialog1.OverwritePrompt = true;
> this.saveFileDialog1.DefaultExt = "xml";
> this.saveFileDialog1.Filter = "Palette files (*.xml)|*.xml|All files (*.*)|(*.*)";
> this.saveFileDialog1.Title = "Save Palette As";
> this.saveFileDialog1.FileName = $"{klbThemeList.GetItemText(klbThemeList.SelectedItem)}.xml";
> if (this.saveFileDialog1.ShowDialog() == DialogResult.OK)
> {
>     this.kcpbCustom.PopulateFromBase(true); 
>     this.kcpbCustom.Export(this.saveFileDialog1.FileName, false);
> }	

### Expected Behavior

The font name value should be serialized

### Actual Behavior

"(none)" is serialized insteat of the default font value

### Operating System

Windows 10

### Framework/.NET Version

4.8

### Toolkit Version

Nightly

### Additional Information

<img width="540" height="48" alt="Image" src="https://github.com/user-attachments/assets/0b922fe3-6bd2-48a3-9b6b-d97b9c8ea0a9" />

### Areas Affected

Toolkit

mcpbcs:
Well… I discovered that the converter serializes using the current system language, which might not be English.

This is the reason why, when loading the theme, the fonts are not applied correctly. The solution would be to adjust the converter so it uses the en-US culture. That would fix the issue immediately.

PWagner1:
Hi @mcpbcs 

Is it possible to backport the fix to both V85 & V105?

mcpbcs:
Sure, I will add it 😉

PR #3252, opened 2026-03-23
* `KryptonForm` shows an extra border on the secondary monitor when m…

PWagner1:
…aximized

# Fix KryptonForm multi-monitor maximization border overflow (#3249)

## Summary

Fixes [#3249](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3249): KryptonForm shows an extra border on the secondary monitor when maximized in a multi-monitor setup. Also resolves a CS8600 nullable reference type warning in `KryptonCustomPaletteBase`.

## Problem

In a multi-monitor environment, maximizing a KryptonForm causes a visible border or shadow area to extend into the secondary monitor. This does not occur with a standard WinForms Form. The issue is especially noticeable when the primary monitor is on the left and the secondary monitor is on the right (or vice versa).

Per [issue comment #3249](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3249#issuecomment-4105787453), when `UseThemeFormChromeBorderWidth` is enabled, `BorderWidths(FormBorderStyle)` always returned (0, 0). Windows still applies non-client area adjustments internally when maximized. Since Krypton assumed no border, the calculated maximized bounds ended up slightly larger than the monitor's working area, causing overflow onto the secondary monitor.

### 1. `VisualForm.cs` – `OnWM_GETMINMAXINFO`

- **Fix:** Use primary-monitor coordinates for `ptMaxPosition` per MSDN documentation.
- For multi-monitor setups, `MINMAXINFO.ptMaxPosition` must be expressed relative to the primary monitor so Windows correctly places the maximized window on the target (possibly secondary) monitor.
- Added `MonitorFromWindow(IntPtr.Zero, MONITOR_DEFAULTTOPRIMARY)` to obtain primary monitor info and adjusted the position calculation.
- References: [StackOverflow](https://stackoverflow.com/questions/35984883/handling-wm-getminmaxinfo-for-multiple-monitors).

### 2. `PaletteFormBorder.cs` – `BorderWidths`

- **Fix:** Removed the `UseThemeFormChromeBorderWidth` branch that returned (0, 0).
- `BorderWidths` now always returns the actual system metrics based on `FormBorderStyle` (e.g. `SM_CXSIZEFRAME`, `SM_CYSIZEFRAME` for `Sizable`).
- This allows `UpdateRegionForMaximized` in KryptonForm to compute the maximized region correctly, accounting for the non-client area that Windows reserves internally.

### 3. `KryptonCustomPaletteBase.cs` – CS8600 warning

- **Fix:** Resolved CS8600 (converting possible null to non-nullable type) in the theme export path.
- `TypeConverter.ConvertTo` returns `object?`; the previous code cast directly to `string`, causing the warning.
- Now uses a local variable and `?.ToString() ?? string.Empty` to handle possible null safely.

## Testing

- Build solution for all target frameworks (net472, net48, net481, net8.0-windows, net9.0-windows, net10.0-windows, net11.0-windows).
- Run TestForm and maximize a KryptonForm on the primary monitor with a secondary monitor attached; the border should no longer extend onto the secondary monitor.
- Verify palette XML export still serializes property values correctly (regression check for #3164).

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonForm.cs
@@ -2741,45 +2741,26 @@ public class KryptonForm : VisualForm,
                 return;
             }
 
-            // Get the size of each window border
-            var xBorder = PI.GetSystemMetrics(PI.SM_.CXSIZEFRAME) * 2;
-            var yBorder = PI.GetSystemMetrics(PI.SM_.CYSIZEFRAME) * 2;
+            // Get per-side border widths
+            var formBorders = StateCommon?.Border as PaletteFormBorder;
 
-            // Fix for #2457, please do not remove!!!
-            // Get the actual border widths from the form's border palette
-            var formBorder = StateCommon?.Border as PaletteFormBorder;
-            var (leftBorder, topBorder) = formBorder?.BorderWidths(FormBorderStyle) ?? (xBorder / 2, yBorder / 2);
-            var rightBorder = leftBorder; // Use same width for right border
-            var bottomBorder = topBorder; // Use same width for bottom border
+            var (borderX, borderY) =
+                formBorders?.BorderWidths(FormBorderStyle)
+                ?? (
+                    PI.GetSystemMetrics(PI.SM_.CXSIZEFRAME),
+                    PI.GetSystemMetrics(PI.SM_.CYSIZEFRAME)
+                );
 
-            // Calculate the maximized region with proper border handling
+            // Convert to TOTAL border (this is REQUIRED for maximized Region)
+            int totalBorderX = borderX * 2;
+            int totalBorderY = borderY * 2;
+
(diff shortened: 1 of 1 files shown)


mcpbcs:
Hi @PWagner1 
The border is still there:

screenshort 
image

PWagner1:
> The border is still there:

Hi @mcpbcs 

Please retry, it's to do with how Windows handles multiple monitors.

mcpbcs:
> > The border is still there:
> 
> Hi @mcpbcs
> 
> Please retry, it's to do with how Windows handles multiple monitors.

Sorry, still the same..
image

image

mcpbcs:
Its not working...
image

PWagner1:
> Its not working... image

Would you mind taking a look?

mcpbcs:
> > Its not working... image
> 
> Would you mind taking a look?

Hi @PWagner1 
I’m pulling your latest changes in .NET 4.8, but the border is still showing. Even in the latest version, the form appears to be clipped

Issue #3256, opened 2026-03-24
[Bug]: Tree View Event is Crashing


PraveenSoftwareEngineer:
### Summary

crash when TreeView Clicked

### Description

when i click the TreeView application is crashing 

### Steps to Reproduce

run then testform 
scroll to TreeView
Clik TreeView 

### Expected Behavior

it should show the Tree Nodes , but its not showing , instead its crashing

### Actual Behavior

crashing with error message with " Method Signature: getpalletebuttonspec
,function : getpalletebuttonspec,file: e:\krytondocumentation\standard-toolkit\source\krypton\components\kryton.toolkit\controlstoolkit\krytoncustompalletebase.cs line number:4254"

### Operating System

windows 11

### OS Version

25H2

### Framework/.NET Version

4.8

### Additional Information

<img width="395" height="481" alt="Image" src="https://github.com/user-attachments/assets/23d3f4da-1886-454c-ae64-b993af738ad9" />

### Areas Affected

None

Issue #3072, opened 2026-02-18
[Bug]: Some samples do not work.


mcpbcs:
### Summary

Some samples do not work.

### Description

AboutBox.:click in "Show" button.  error: "The file 'KryptonTextBox1' does not exist"
Splash Screen Sample: Open the form: error "System.ArgumentException: "
TreeView sample: Error: "Krypton.Toolkit.KryptonCustomPaletteBase.GetPaletteButtonSpec(PaletteButtonSpecStyle style)"
PaletteViewerForm: Form is not loading

### Steps to Reproduce

Check test samples

### Expected Behavior

Some samples do not work.

### Actual Behavior

.

### Operating System

Windows 10

### Framework/.NET Version

4.8

### Areas Affected

None

PWagner1:
Hi @mcpbcs 

Is this affecting the demos repo? Also which demos are affected?

mcpbcs:
Hi @PWagner1 

The TreeView sample is still in error..

<img width="1019" height="258" alt="Image" src="https://github.com/user-attachments/assets/80d4bf01-f054-4305-9c17-6f8e12d5c883" />

PR #3258, opened 2026-03-24
* Tree View Event is Crashing (V110)

PWagner1:
# Fix crash when using custom palette with title-bar / toolbar button specs (#3256)

## Summary

Resolves a crash reported when interacting with controls (for example a `KryptonTreeView`) while using **`PaletteMode.Custom`** with **`KryptonCustomPaletteBase`**. The failure surfaced as the integrated “Not Implemented” dialog pointing at `GetPaletteButtonSpec` in `KryptonCustomPaletteBase.cs`.

## Root cause

`KryptonFormTitleBar` creates standard toolbar **`ButtonSpecAny`** entries with types such as **`New`**, **`Open`**, **`Save`**, and so on. Those types map to **`PaletteButtonSpecStyle`** values that must be resolved through the active palette.

If any **`PaletteButtonSpecStyle`** still fell through the **`switch`** in **`KryptonCustomPaletteBase.GetPaletteButtonSpec`**, the **`default`** branch called **`DebugTools.NotImplemented`**, which blocks the UI and throws. A repaint triggered by ordinary input (e.g. clicking the tree view) is enough to hit that path.

Related context: handling for the full set of toolbar-related **`PaletteButtonSpecStyle`** values in this method was expanded for **#3072**; this change completes the behaviour for **#3256** by making the **`default`** path safe when an unexpected or out-of-range value appears (forward compatibility, bad serialization, etc.).

## Changes

- **`KryptonCustomPaletteBase.GetPaletteButtonSpec`**: **`default`** now **`Debug.Assert`s** (debug builds) with the numeric style value, then **returns `ButtonSpecs.Generic`** instead of **`DebugTools.NotImplemented`**, so the application does not hard-crash.
- **`Documents/Changelog/Changelog.md`**: entry for **#3256**.

## Testing

- Run **Test Form** → **Tree View** example (or **Tree View Test**) with **`KryptoManager.GlobalPaletteMode = PaletteMode.Custom`** and a **`KryptonCustomPaletteBase`** instance as **`GlobalCustomPalette`**.
- Click the **`KryptonTreeView`** and interact with nodes; confirm no **“Not Implemented – GetPaletteButtonSpec”** dialog and no related exception.

## References

- Closes / fixes [Tree View Event is Crashing (#3256)](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3256)
- Related: [Some samples do not work (#3072)](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3072)

image

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
@@ -4309,7 +4309,10 @@ public class KryptonCustomPaletteBase : PaletteBase
             default:
                 // Unknown or out-of-range values (e.g. newer enum, bad serialization) — avoid crashing the UI.
                 Debug.Assert(false, $"Unhandled {nameof(PaletteButtonSpecStyle)}: {(int)style}");
-                DebugTools.NotImplemented(style.ToString());
+                if (Debugger.IsAttached)
+                {
+                    throw DebugTools.NotImplemented(style.ToString());
+                }
                 return ButtonSpecs.Generic;
         }
     }


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
```diff
@@ -4307,9 +4307,10 @@ private IPaletteButtonSpec GetPaletteButtonSpec(PaletteButtonSpecStyle style)
             case PaletteButtonSpecStyle.Search:
                 return ButtonSpecs.Generic;
             default:
-                // Should never happen!
-                Debug.Assert(false);
-                throw DebugTools.NotImplemented(style.ToString());
+                // Unknown or out-of-range values (e.g. newer enum, bad serialization) — avoid crashing the UI.
+                Debug.Assert(false, $"Unhandled {nameof(PaletteButtonSpecStyle)}: {(int)style}");
```
Smurf-IV, line 4311: Not sure I agree with this..
The whole idea of an exception is to improve the Style, not mask it out!

PWagner1: Has now been improved

PR #3263, opened 2026-03-25
* Small square rendered next to Close button on KryptonForm when usin…

### LINKED ISSUE #3183: [Bug]: Small square rendered next to Close button on KryptonForm when using a custom theme
### Summary

A small unexpected square appears next to the Close (X) button on the title bar of a KryptonForm when a custom palette/theme is applied.

### Description

Observed behavior
When running a KryptonForm with a custom theme/palette, a small square is drawn next to the Close button in the title bar. This element does not appear when using built-in palettes such as Office2010 or Office365.
Screenshot attached.

Image

### Steps to Reproduce

Load a custom theme.
Added: Microsoft365Pink

### Expected Behavior

The caption area should only display standard window buttons (Minimize / Maximize / Close), with no additional square or ghost button.

[Microsoft365 Super Pink.xml](https://github.com/user-attachments/files/25880156/Microsoft365.Super.Pink.xml)

### Actual Behavior

Only the standard window buttons (Minimize, Maximize, Close) should appear on the KryptonForm title bar. No additional shapes, icons, or squares should be rendered, even when a custom palette/theme is applied.

### Operating System

Windows 10

### OS Version

_No response_

### Framework/.NET Version

4.8

### Toolkit Version

Nightly

### Additional Information

Image

### Areas Affected

Toolkit

PWagner1:
…g a custom theme

# PR Description - #3183

## Summary
Fix a rendering artifact (a small square) appearing next to the `KryptonForm` Close (X) button when using a custom XML-based Krypton theme/palette.

## Root Cause
When importing a custom palette from XML, palette values that are omitted by the theme XML could remain at placeholder/default C# values instead of being initialized from the palette's `BasePalette`. This could leave caption/button chrome in an invalid/partial state, producing a small stray visual element.

## Changes
In `KryptonCustomPaletteBase.ImportFromXmlDocument`, the palette is now populated from `BasePalette` before applying the XML overrides:
- Call `PopulateFromBaseOperation(null)` before importing the `Properties` / `Images` elements from the XML.

## Why this works
The initial populate step ensures missing metrics/draw settings (including caption/button related values used by the form window buttons) are correctly inherited from the base palette. The subsequent XML import can then override only what the theme explicitly defines.

## Files Modified
- `Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs`

## Testing
- Repro: apply a custom theme (e.g. `Microsoft365Pink`) to a `KryptonForm` and verify the small square near the Close button is gone.
- Sanity: verify built-in palettes are unaffected.

## Notes
- A full `dotnet build` was not clean in this environment due to pre-existing MSBuild/resource configuration errors unrelated to this palette change.

image


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
```diff
@@ -3124,6 +3124,14 @@ private void ImportFromXmlDocument([DisallowNull] XmlDocument doc)
                 SetPaletteName(root.GetAttribute("Name"));
             }
 
+            // Populate missing palette values from the base palette first.
```
Smurf-IV: Should this be done "First" then allow the loaded pallette to override re-set ?

Smurf-IV (changes requested):

Source/Krypton Components/TestForm/Bug3183SmallSquareRenderedNextToClose.cs
```diff
@@ -0,0 +1,39 @@
...
+
+using TestForm.Properties;
+
+namespace TestForm
+{
+	public partial class Bug3183SmallSquareRenderedNextToClose : KryptonForm
+	{
+		public Bug3183SmallSquareRenderedNextToClose()
+		{
+			InitializeComponent();
```
Smurf-IV: Please state in a "Text element" What this form is attempting to check for, otherwise it's just a form that pop's up and may be deleted in the future !

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonCustomPaletteBase.cs
PWagner1: Added documentation explaining why this is the case

Source/Krypton Components/TestForm/Bug3183SmallSquareRenderedNextToClose.cs
PWagner1: Added documentation

mcpbcs:
Hi @PWagner1,

The issue is still present.
I noticed that the color value is coming directly from the XML. I captured the color using ColorPic and obtained the value RGB(185, 162, 171).
This makes me wonder if the problem could be related to the exporting process rather than the importing one. Do you think that could be the case?

image

PWagner1:
> Hi @PWagner1,
> 
> The issue is still present. I noticed that the color value is coming directly from the XML. I captured the color using ColorPic and obtained the value RGB(185, 162, 171). This makes me wonder if the problem could be related to the exporting process rather than the importing one. Do you think that could be the case?
> 
> image

Hi @mcpbcs 

Thanks for the info, there's only one way to find out and that is to test that theory.

PWagner1:
Hi @mcpbcs 

Would you mind testing the recent changes? It seemed to be a hard coded colour value, you'll need to re-export the theme to remove errant values.

cc. @Smurf-IV

mcpbcs:
Okay, I will test it

PR #3284, opened 2026-03-29
* Fix double bracket issue

PWagner1:
* Fix double bracket issue

image

### DIFF
--- a/Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
+++ b/Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
@@ -1,4 +1,4 @@
-﻿#region BSD License
+#region BSD License
 /*
  * 
  *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
@@ -124,13 +124,7 @@ public partial class RichTextBoxFormattingTest : KryptonForm
 
     private void KbtnVerifyFormatting_Click(object? sender, EventArgs e)
     {
-        // Check if RTF formatting exists
-        string? rtf = krtbRichTextBox.Rtf;
-        bool hasFormatting = !string.IsNullOrEmpty(rtf) &&
-                             rtf != null &&
-                             (rtf.Contains(@"\b") || rtf.Contains(@"\i") || rtf.Contains(@"\ul") ||
-                              rtf.Contains(@"\fs") || rtf.Contains(@"\cf") || rtf.Contains(@"\highlight") ||
-                              (rtf.Contains(@"\f") && !rtf.Contains(@"\f0")));
+        bool hasFormatting = RtfAppearsFormatted(krtbRichTextBox.Rtf);
 
         if (hasFormatting)
         {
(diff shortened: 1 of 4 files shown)


Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
```diff
@@ -118,18 +118,19 @@ private void KbtnLoadSample_Click(object? sender, EventArgs e)
...
     private void KbtnVerifyFormatting_Click(object? sender, EventArgs e)
     {
         // Check if RTF formatting exists
-        string rtf = krtbRichTextBox.Rtf;
+        string? rtf = krtbRichTextBox.Rtf;
         bool hasFormatting = !string.IsNullOrEmpty(rtf) &&
-            (rtf.Contains(@"\b") || rtf.Contains(@"\i") || rtf.Contains(@"\ul") ||
-             rtf.Contains(@"\fs") || rtf.Contains(@"\cf") || rtf.Contains(@"\highlight") ||
-             (rtf.Contains(@"\f") && !rtf.Contains(@"\f0")));
+                             rtf != null &&
```
Smurf-IV: Not needed as it is already checked in the `IsNull###` step

Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
```diff
@@ -118,18 +118,19 @@ private void KbtnLoadSample_Click(object? sender, EventArgs e)
...
+                             (rtf.Contains(@"\b") || rtf.Contains(@"\i") || rtf.Contains(@"\ul") ||
```
Smurf-IV: These seem inefficient.
cannot the rtf be made a readonly span and then check for each of these?

Smurf-IV (changes requested):

Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
PWagner1: Now removed

Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
PWagner1: Now uses span.

Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
```diff
@@ -144,6 +138,48 @@ private void KbtnVerifyFormatting_Click(object? sender, EventArgs e)
...
+    /// Single-pass heuristic over RTF: same substrings as prior Contains checks, without rescanning the string for each token.
+    /// </summary>
+    private static bool RtfAppearsFormatted(string? rtf)
+    {
+        if (string.IsNullOrEmpty(rtf))
+        {
+            return false;
+        }
+
+        ReadOnlySpan<char> s = rtf.AsSpan();
```
Smurf-IV: Thanks

Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
```diff
@@ -144,6 +138,48 @@ private void KbtnVerifyFormatting_Click(object? sender, EventArgs e)
...
+        bool hasF0 = false;
+
+        for (int i = 0; i < s.Length; i++)
+        {
+            if (s[i] != '\\')
+            {
+                continue;
+            }
+
+            ReadOnlySpan<char> tail = s.Slice(i);
```
Smurf-IV: No.
Why recreate a Full slice every time ?
Max size is the length of max string.
But that's not the point.
span has a contains equivalent ??
Or is it TFM'd ?

Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
```diff
@@ -144,6 +138,48 @@ private void KbtnVerifyFormatting_Click(object? sender, EventArgs e)
...
+                hasF0 = true;
+            }
+
+            if (tail.StartsWith("\\f"))
+            {
+                hasF = true;
+            }
+        }
+
+        return hasF && !hasF0;
```
Smurf-IV: If leaving the above loop in then, this should be inside it to break out earlier.

Source/Krypton Components/TestForm/TestForm.csproj
```diff
@@ -85,4 +85,8 @@
     <PackageReference Include="Cyotek.Windows.Forms.ColorPicker" Version="2.0.0-beta.7" />
   </ItemGroup>
 
+  <ItemGroup Condition="$(TargetFramework.StartsWith('net4'))">
+    <PackageReference Include="System.Memory" Version="4.5.5" />
```
Smurf-IV: does .net 4.8.x have a different version than 4.7.x

Smurf-IV (changes requested):

Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
PWagner1: Has now been reworked

Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
PWagner1: Used modified version from `alpha` branch

Source/Krypton Components/TestForm/RichTextBoxFormattingTest.cs
PWagner1: Used modified version from `alpha` branch

Source/Krypton Components/TestForm/TestForm.csproj
PWagner1: Has now been removed

Smurf-IV (commented):
So, What are the results from these 2, and why not also include the `AsSpan` version(s)

PWagner1:
> So, What are the results from these 2, and why not also include the `AsSpan` version(s)

Now has no warnings

image

image

`AsSpan` will be added in a future PR.

Smurf-IV (approved):

Issue #3343, opened 2026-04-21
[Bug]: Strange behavior of `KryptonRichTextBox` when editing text


antosupersayan-vb:
### Summary

you can't edit the text

### Description

You enter text, use the mouse to exit the control, and the entered text is lost.
I've only encountered this problem with this control; it doesn't happen on the KryptonTexbox.
I'm attaching a video.

### Steps to Reproduce

You enter text, use the mouse to exit the control, and the entered text is lost.

### Expected Behavior

that the entered text is not lost

### Actual Behavior

you cannot edit the text, losing the entered data

### Operating System

windows 11 PRO

### OS Version

26200.8246

### Framework/.NET Version

4.8.1

### Toolkit Version

110.26.4.110-alpha

### Areas Affected

Toolkit

antosupersayan-vb:
https://github.com/user-attachments/assets/a1d2d082-3a6f-4830-b313-be8b006f9413

PR #3345, opened 2026-04-21
* Strange behavior of `KryptonRichTextBox` when editing text

PWagner1:
# Fix: KryptonRichTextBox loses typed text when the mouse leaves the control (#3343)

## Summary

Fixes a regression where `KryptonRichTextBox` could clear or revert recently typed content when the pointer left the control (for example after `TrackMouseLeave` / hover state changes that trigger a full palette repaint). Adds a **TestForm** demo to reproduce and verify the scenario.

Closes https://github.com/Krypton-Suite/Standard-Toolkit/issues/3343

## Problem

- User types a small amount of text in `KryptonRichTextBox`.
- Without committing focus elsewhere, the user moves the mouse out of the control.
- The entered text disappeared or reverted. The same workflow with `KryptonTextBox` did not show the issue.

## Root cause

`KryptonRichTextBox.OnNeedPaint` saves the current `Rtf`, applies palette-driven `BackColor` / `ForeColor` / `Font` updates, then may reassign `Rtf` to undo formatting loss.

The restore path used a cached `_originalRtf` that was only refreshed when the RTF string length changed by more than an arbitrary threshold compared to the cache. **Small edits** did not refresh the cache, so after palette-related mutations the control could restore **stale RTF** and wipe the latest keystrokes.

Default RTF commonly includes font-table sequences such as `\f1`, so the “has RTF formatting” path often runs even for otherwise plain typing, which made the bug easy to hit.

## Solution

- Always treat the RTF snapshot taken **immediately before** palette mutations (`savedRtf`) as the authoritative document when restoring after those property changes.
- Sync `_originalRtf` with that snapshot for consistency.
- Apply existing dark-mode black-text RTF adjustments to `savedRtf` instead of a possibly stale `_originalRtf`.

## Files changed

| Area | File |
|------|------|
| Fix | `Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonRichTextBox.cs` (`OnNeedPaint` RTF restore block) | | Demo | `Source/Krypton Components/TestForm/Bug3343RichTextBoxEditLossDemo.cs` | | Demo | `Source/Krypton Components/TestForm/Bug3343RichTextBoxEditLossDemo.Designer.cs` | | Launcher | `Source/Krypton Components/TestForm/StartScreen.cs` (entry: **Bug 3343 RichTextBox mouse leave**) |

## How to test

1. Build and run TestForm, for example: `dotnet run --project "Source/Krypton Components/TestForm/TestForm.csproj" -c Debug -f net481`
2. Open **Bug 3343 RichTextBox mouse leave** from the start screen.
3. Type a few characters in `KryptonRichTextBox`, then move the mouse out of the rich text area (without relying on a large RTF delta).
4. Confirm **TextLength** and visible text stay correct; the **TrackMouseLeave** status line should show a non-zero length matching what you typed.

Optional: compare with the multiline `KryptonTextBox` on the same form.

## Risk / follow-up

- **Low risk:** Restoration now follows the pre-paint snapshot of the document, which matches the intended “preserve content across palette property writes” behavior.
- The **`Rtf` property setter** still uses a separate length-based heuristic when updating `_originalRtf`; that is unrelated to the mouse-leave paint path but could be reviewed later if programmatic small `Rtf` assignments need the same strictness.

## Checklist (for maintainers)

- [x] Targeted change in `KryptonRichTextBox` only where needed
- [x] Manual repro covered in TestForm
- [x] `Krypton.Toolkit` and `TestForm` build (`net481` verified)

image


Smurf-IV (changes requested):
The following needs sorting:
`Possible null reference argument for parameter 'rtf' in 'string KryptonRichTextBox.RemoveBlackColorCodes(string rtf)'.`

Also needs ot go into V105

PWagner1:
> The following needs sorting: `Possible null reference argument for parameter 'rtf' in 'string KryptonRichTextBox.RemoveBlackColorCodes(string rtf)'.`
> 
> Also needs ot go into V105

I thought this was fixed in https://github.com/Krypton-Suite/Standard-Toolkit/pull/3345/commits/5dc81739c3c16fcf7a8c069f2fda51d653eb466d?

Smurf-IV:
Merge conflicts and the 

The following needs sorting:
Possible null reference argument for parameter 'rtf' in 'string KryptonRichTextBox.RemoveBlackColorCodes(string rtf)'.

PWagner1:
> Possible null reference argument for parameter 'rtf' in 'string KryptonRichTextBox.RemoveBlackColorCodes(string rtf)'

@Smurf-IV 

Where is it reporting this, as I can't trigger it?

image

PWagner1:
@Smurf-IV 

Is the warning still here?

PWagner1:
Hi @Smurf-IV 

Is the warning still present?

PWagner1:
@Smurf-IV 

Ignore the failure, as the stable version is trying to build using .NET 11. Will resolve in a future PR, it does not affect normal build/NuGet operations.

Smurf-IV (approved):

Issue #3365, opened 2026-04-27
[Bug]: `KryptonLinkLabel` 'Autosize' Not Shrinking


cuppm:
### Summary

A `KryptonLinkLabel` with `Autosize=true` does not get smaller in v100 and v105

### Description

Something is preventing the `KryptonLinkLabel` (and also `KryptonLabel`) from appropriately resizing smaller.  It seems to have a minimum size of `(90, 25)` that it won't autosize down unless you delete (or set to `(0,0)`) the `Size` property in the designer.  It happens with both the v105 and 100 releases.

### Steps to Reproduce

1. Create project and add v105 or v100 from Nuget
2. Create form
3. Add a `KryptonLinkLabel` control
4. Change the text to "a" from the default

### Expected Behavior

The `KryptonLinkLabel` control should get smaller like it does with v95 library.

<img width="153" height="109" alt="Image" src="https://github.com/user-attachments/assets/25b6538d-63ee-49a6-b10e-cea3d1a610b9" />

### Actual Behavior

The KryptonLinkLabel control stays the same size as the initial size assigned at creation.
<img width="160" height="116" alt="Image" src="https://github.com/user-attachments/assets/c1ddcbe3-2412-46dd-8ee3-f9dd5f69bef4" />
  
If I make the text longer, it appropriately gets bigger:
<img width="341" height="101" alt="Image" src="https://github.com/user-attachments/assets/a9f78bea-855d-4db2-86be-05dcbedb7d04" />
  
Then if I make it smaller again, it goes back to the original size:
<img width="163" height="105" alt="Image" src="https://github.com/user-attachments/assets/76b7c275-c2ae-4c15-8cb9-acbef9bd86c3" />
  
Then if I delete the value for Size in the designer properties, it now starts to auto resize appropriately:
<img width="241" height="217" alt="Image" src="https://github.com/user-attachments/assets/fa2a9e8a-24aa-4862-8b03-d7cb5cbb24f4" />
<img width="170" height="98" alt="Image" src="https://github.com/user-attachments/assets/7b0a8086-f4bf-47c3-9ab7-53142b9c09b6" />

### Operating System

Windows 11

### OS Version

25H2 (26200.8246)

### Framework/.NET Version

4.8

### Toolkit Version

105.26.4.110

### Additional Information

If you create the project and form with v105 or v100 and then change the Krypton version to v95 in Nuget Package Manager, it will appropriately resize.

[...]

PR #3369, opened 2026-04-29
* `KryptonLinkLabel` 'Autosize' Not Shrinking (V110)

PWagner1:
## Summary

- Fixes `KryptonLabel` / `KryptonLinkLabel` autosize shrink behavior regression reported in [#3365](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3365).
- Updates `KryptonLabel.SetBoundsCore()` to always apply preferred size when `AutoSize` is enabled, instead of only when `BoundsSpecified.Size` is present.
- Prevents the initial designer-assigned `Size` from acting as an unintended minimum when label text becomes shorter.

## Root Cause

`SetBoundsCore()` only recalculated to preferred size when incoming bounds updates included `BoundsSpecified.Size`. Some autosize update paths (designer/runtime) do not set that flag when content changes, so shrink updates were skipped and the original control size remained as an effective floor.

## Implementation

- File changed: `Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonLabel.cs`
- Change made in `SetBoundsCore(int x, int y, int width, int height, BoundsSpecified specified)`:
  - Apply preferred size whenever `AutoSize` is `true`.
  - Force size update participation with `specified |= BoundsSpecified.Size` after a valid preferred size is computed.
- `KryptonLinkLabel` is fixed via inheritance (`KryptonLinkLabel : KryptonLabel`).

## Test Plan

- [ ] Create a WinForms form with a `KryptonLinkLabel` (default `AutoSize=true`).
- [ ] Set text to a long string, then change it to `"a"`.
- [ ] Verify the control shrinks to content size (does not snap back to the original creation size).
- [ ] Repeat the same validation with `KryptonLabel`.
- [ ] Confirm `AutoSize=false` controls still retain explicitly assigned sizes.

## Risk / Compatibility

- Low risk: change is scoped to autosize bounds handling for `KryptonLabel` and inherited `KryptonLinkLabel`.
- Existing guardrails for invalid preferred sizes remain in place to avoid unstable designer sizing.


KamaniAR (approved):
Work as expected.

PR #3370, opened 2026-04-29
* `KryptonLinkLabel` 'Autosize' Not Shrinking (V105 LTS)

### LINKED ISSUE #3365: [Bug]: `KryptonLinkLabel` 'Autosize' Not Shrinking

PWagner1:
## Summary

- Fixes `KryptonLabel` / `KryptonLinkLabel` autosize shrink regression reported in [#3365](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3365).
- Updates `KryptonLabel.SetBoundsCore()` to always apply preferred size when `AutoSize` is enabled.
- Prevents the initial designer-assigned `Size` from acting like a hidden minimum when label text becomes shorter.

## Root Cause

`SetBoundsCore()` did not consistently enforce preferred size in all autosize update paths. In affected runtime/designer paths, shrink recalculation could be skipped, so controls returned to their original creation size instead of shrinking to fit content.

## Implementation

- File changed: `Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonLabel.cs`
- Reintroduced `SetBoundsCore(int x, int y, int width, int height, BoundsSpecified specified)` to:
  - compute preferred size whenever `AutoSize` is `true`
  - apply the computed size when valid
  - force size bounds participation with `specified |= BoundsSpecified.Size`
- Existing guardrails for invalid preferred sizes remain to protect designer initialization.
- `KryptonLinkLabel` is fixed via inheritance (`KryptonLinkLabel : KryptonLabel`).

## Test Plan

- [ ] Add a `KryptonLinkLabel` with default `AutoSize=true`.
- [ ] Change text from long to `"a"` and verify it shrinks.
- [ ] Change text from short to long and verify it grows.
- [ ] Repeat with `KryptonLabel`.
- [ ] Confirm `AutoSize=false` labels retain explicit sizes.

## Risk / Compatibility

- Low risk: behavior scoped to autosize sizing path in `KryptonLabel` and inherited `KryptonLinkLabel`.
- No API surface changes. ## Summary

- Fixes `KryptonLabel` / `KryptonLinkLabel` autosize shrink behavior regression reported in [#3365](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3365).
- Updates `KryptonLabel.SetBoundsCore()` to always apply preferred size when `AutoSize` is enabled, instead of only when `BoundsSpecified.Size` is present.
- Prevents the initial designer-assigned `Size` from acting as an unintended minimum when label text becomes shorter.

## Root Cause

`SetBoundsCore()` only recalculated to preferred size when incoming bounds updates included `BoundsSpecified.Size`. Some autosize update paths (designer/runtime) do not set that flag when content changes, so shrink updates were skipped and the original control size remained as an effective floor.

## Implementation

- File changed: `Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonLabel.cs`
- Change made in `SetBoundsCore(int x, int y, int width, int height, BoundsSpecified specified)`:
  - Apply preferred size whenever `AutoSize` is `true`.
  - Force size update participation with `specified |= BoundsSpecified.Size` after a valid preferred size is computed.
- `KryptonLinkLabel` is fixed via inheritance (`KryptonLinkLabel : KryptonLabel`).

## Test Plan

- [ ] Create a WinForms form with a `KryptonLinkLabel` (default `AutoSize=true`).
- [ ] Set text to a long string, then change it to `"a"`.
- [ ] Verify the control shrinks to content size (does not snap back to the original creation size).
- [ ] Repeat the same validation with `KryptonLabel`.
- [ ] Confirm `AutoSize=false` controls still retain explicitly assigned sizes.

## Risk / Compatibility

- Low risk: change is scoped to autosize bounds handling for `KryptonLabel` and inherited `KryptonLinkLabel`.
- Existing guardrails for invalid preferred sizes remain in place to avoid unstable designer sizing.

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonNumericUpDown.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonNumericUpDown.cs
@@ -739,6 +739,7 @@ public class KryptonNumericUpDown : VisualControlBase,
     private bool _alwaysActive;
     private bool _trackingMouseEnter;
     private bool _autoSize;
+    private int _cachedWidth;
     private int _cachedHeight;
     private Graphics? _graphics;
 
@@ -831,6 +832,7 @@ public class KryptonNumericUpDown : VisualControlBase,
         _upDownButtonStyle = ButtonStyle.InputControl;
         _alwaysActive = true;
         _autoSize = false;
+        _cachedWidth = -1;
         _cachedHeight = -1;
         _graphics = null;
         AllowButtonSpecToolTips = false;
@@ -934,8 +936,18 @@ public class KryptonNumericUpDown : VisualControlBase,
         {
             if (_autoSize != value)
             {
+                CacheCurrentDimensionBeforeAutoSizeEnable(value, Width, ref _cachedWidth);
                 _autoSize = value;
-                UpdateAutoSizing();
+
+                if (_autoSize)
+                {
+                    UpdateAutoSizing();
+                }
+                else if (TryGetCachedDimensionForRestore(_cachedWidth, out int restoredWidth))
+                {
+                    Width = restoredWidth;
+                    PerformNeedPaint(true);
+                }
             }
         }
     }
@@ -1829,12 +1841,17 @@ public class KryptonNumericUpDown : VisualControlBase,
         int width, int height,
         BoundsSpecified specified)
     {
+        if (!_autoSize)
+        {
+            CacheDimensionIfSpecified(specified, BoundsSpecified.Width, width, ref _cachedWidth);
+        }
+
         // Changed from inline GetPreferredSize() to the same pattern as KryptonComboBox,
         // KryptonDateTimePicker, and KryptonDomainUpDown: cache incoming height on first set,
         // then always override to PreferredHeight (which honours MinimumControlHeight).
         // See https://github.com/Krypton-Suite/Standard-Toolkit/issues/615
         // If setting the actual height
-        if ((specified & BoundsSpecified.Height) == BoundsSpecified.Height)
+        if (IsDimensionSpecified(specified, BoundsSpecified.Height))
         {
             // First time the height is set, remember it
             if (_cachedHeight == -1)
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Visuals/VisualSimpleBase.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Visuals/VisualSimpleBase.cs
@@ -150,7 +150,7 @@ public abstract class VisualSimpleBase : VisualControlBase
     /// <param name="specified">A bitwise combination of the BoundsSpecified values.</param>
     protected override void SetBoundsCore(int x, int y, int width, int height, BoundsSpecified specified)
     {
-        if (AutoSize && GetAutoSizeMode() == AutoSizeMode.GrowAndShrink)
+        if (AutoSize)
         {
             Size preferredSize = GetPreferredSize(new Size(int.MaxValue, int.MaxValue));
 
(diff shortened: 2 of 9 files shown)


KamaniAR (approved):
Work as expected.

Source/Krypton Components/Krypton.Toolkit/Controls Visuals/VisualSimpleBase.cs
```diff
@@ -134,6 +139,34 @@ public override Size GetPreferredSize(Size proposedSize)
...
+    /// <param name="specified">A bitwise combination of the BoundsSpecified values.</param>
+    protected override void SetBoundsCore(int x, int y, int width, int height, BoundsSpecified specified)
+    {
+        if (AutoSize && GetAutoSizeMode() == AutoSizeMode.GrowAndShrink)
+        {
+            Size preferredSize = GetPreferredSize(new Size(int.MaxValue, int.MaxValue));
+
+            // Only apply sensible calculated sizes to avoid unstable initialization values.
+            if (preferredSize.Width > 0 && preferredSize.Height > 0
+                && preferredSize.Width < 10000 && preferredSize.Height < 10000)
```
Smurf-IV: Not sure I like this Max restriction without it having been read from the OS Max size. (i.e. might be on a multi high resolution screens spread across to display a Full width panel / window)

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonNumericUpDown.cs
```diff
@@ -739,6 +739,7 @@ public NumericUpDownButtonSpecCollection(KryptonNumericUpDown owner)
     private bool _alwaysActive;
     private bool _trackingMouseEnter;
     private bool _autoSize;
+    private int _cachedWidth;
```
Smurf-IV, line 742: Should this be in the base class ?

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Visuals/VisualSimpleBase.cs
PWagner1: Fixed

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonNumericUpDown.cs
PWagner1: No, as if it's too strict, it may have unintended consequences on other controls.

Smurf-IV (commented):
Is this also needed in the other place that these are checked

PWagner1:
> s this also needed in the other place that these are checked

Not really, as it's now all centralised.

Smurf-IV (approved):

PR #3387, opened 2026-05-02
* Memory Retention in Krypton Controls via SystemEvents (V105 LTS)

### LINKED ISSUE #3385: [Bug]: Memory Retention in Krypton Controls via SystemEvents
### Summary

Memory leak: All Krypton controls remaining in memory due to UserPreferenceChangedEvent after disposal

### Description

After resolving application-level leaks (WinForms user controls and event handlers), profiling shows that Krypton controls themselves are still retained after disposal.

Multiple control types (e.g., KryptonCheckBox, KryptonRadioButton, and others) exhibit the same retention pattern. All are rooted via the static Microsoft.Win32.SystemEvents.UserPreferenceChanged event.

Representative retention path:

SystemEvents.UserPreferenceChanged
→ Krypton palette/redirect/image objects
→ NeedPaintHandler
→ Krypton control instance

Example:

SystemEvents
→ PaletteRedirectCheckBox._images
→ CheckBoxImages._needPaint
→ NeedPaintHandler._target
→ KryptonCheckBox

This pattern is consistent across many Krypton control types, indicating a systemic issue within Krypton Toolkit where internal palette/painting infrastructure remains subscribed to SystemEvents, preventing disposed controls from being garbage collected.

Impact: On one affected example screen, each open/close cycle leaves approximately 200 KB+ of Krypton Toolkit objects retained. The retained size grows incrementally with repeated use, so this is not just a one-time static cache cost.

Open count    Retained Krypton memory
1             ~200 KB
5             ~1 MB
10            ~2 MB
50            ~10 MB

Environment:

Krypton.Toolkit 105.26.4.110
Conclusion

This appears to be a library-level memory retention issue affecting multiple Krypton controls, likely related to internal handling of SystemEvents.UserPreferenceChanged and NeedPaint infrastructure.

### Steps to Reproduce

1. Create a WinForms UserControl containing one or more Krypton controls (e.g., KryptonCheckBox, KryptonRadioButton).
2. Host the WinForms control inside a WPF application using WindowsFormsHost.
3. Display the control, then remove it from the visual tree and call Dispose() on the hosted WinForms control.
4. Force garbage collection
5. Capture a memory snapshot using a profiler (e.g., dotMemory).
6. Repeat steps 2–5 multiple times.

### Expected Behavior

- All Krypton controls should be eligible for garbage collection after their parent WinForms control is disposed.
- No references to disposed controls should remain rooted via static events such as SystemEvents.UserPreferenceChanged.
- Repeated creation and disposal of Krypton controls should not result in increasing retained object counts.

### Actual Behavior

- Krypton control instances are not released after disposal and accumulate over time.
- Memory usage increases with each create/dispose cycle of the WinForms host control.
- Retained objects are rooted via the static Microsoft.Win32.SystemEvents.UserPreferenceChanged event.
- The retention chain consistently involves Krypton palette/painting infrastructure (e.g., PaletteRedirect*, CheckBoxImages, NeedPaintHandler).
- This behaviour affects multiple Krypton control types, not just a single control.

### Operating System

Windows 11 Pro 25H2

### OS Version

26200.8037

### Framework/.NET Version

9

### Toolkit Version

105.26.4.110

### Additional Information

_No response_

### Areas Affected

Toolkit

PWagner1:
# Fix memory retention from `SystemEvents.UserPreferenceChanged` on undisposed `PaletteRedirect` (#3385)

## Summary

Krypton controls could remain reachable after disposal because `PaletteBase` always subscribed to the static `Microsoft.Win32.SystemEvents.UserPreferenceChanged` event. `PaletteRedirect` (and derived types) inherit that behaviour but are often **never disposed**—for example each `VisualControlBase` owns a `Redirector` instance that is not disposed on control teardown—so the static event kept redirect palettes, `NeedPaintHandler`, and the control graph alive.

This change **centralises** subscription policy in `PaletteBase`: redirect palettes opt out at construction time; builtin and other real palettes continue to subscribe as before.

Fixes: https://github.com/Krypton-Suite/Standard-Toolkit/issues/3385

## Technical approach

1. **`PaletteBase`**
   - Parameterless constructor chains to `PaletteBase(attachUserPreferenceChanged: true)` (unchanged behaviour for all existing callers).
   - New `protected PaletteBase(bool attachUserPreferenceChanged)` only registers `SystemEvents.UserPreferenceChanged` when `attachUserPreferenceChanged` is `true`.
   - Field `_systemEventsUserPreferenceSubscribed` records whether a subscription was made.
   - `Dispose(bool)` removes the handler **only** when that subscription exists, so redirect instances that never subscribed do not rely on a redundant `-=`.

2. **`PaletteRedirect`**
   - Constructor calls `: base(attachUserPreferenceChanged: false)` so redirect instances **do not** attach to `SystemEvents`.
   - Theme / user-preference updates still propagate through the **target** palette (global or builtin palettes remain subscribed and disposed with their normal lifecycle).

## Files changed

| Area | File |
|------|------|
| Core | `Source/Krypton Components/Krypton.Toolkit/Palette Builtin/Base/PaletteBase.cs` | | Core | `Source/Krypton Components/Krypton.Toolkit/Palette Base/PaletteRedirect/PaletteRedirect.cs` |

## Breaking changes

None intended. Public surface of `PaletteBase` gains an additional `protected` constructor overload; existing derived types that use the implicit parameterless `PaletteBase()` path are unchanged.

## Manual verification

1. Build `Krypton.Toolkit` (e.g. `dotnet build` on `Krypton.Toolkit 2022.csproj` for a `-windows` TFM).
2. Reproduce the scenario from the issue (e.g. WinForms user control with Krypton controls hosted in WPF via `WindowsFormsHost`, open/close/dispose, force GC).
3. Capture a memory snapshot (e.g. dotMemory) and confirm disposed controls are no longer rooted via `SystemEvents.UserPreferenceChanged` → `PaletteRedirect*` / `CheckBoxImages` / `NeedPaintHandler` → control.

## Follow-up (optional)

Other code paths still subscribe to `SystemEvents` directly (e.g. some `KryptonColorTable*` types, `KryptonManager`, `VisualForm`, certain theme classes). If profiling shows remaining roots, those can be addressed separately; they are outside the `PaletteRedirect` / `PaletteBase` chain described in #3385.

## Checklist for reviewers

- [ ] Behaviour after Windows theme / high-contrast / related user preference changes is acceptable for Krypton controls using redirect palettes.
- [ ] No regression for long-lived forms and global palette instances.

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -47,6 +47,7 @@
 
  ## 2026-07-20 - Build 2607 (Version 105-LTS - Patch 3) - July 2026
 
+ * Resolved [#3385](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3385), Memory Retention in Krypton Controls via SystemEvents
  * Resolved [#3342](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3342), `KryptonTextBox`, text flickers when resizing the control (multiline active)
  * Resolved [#3283](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3283), `KryptonThemeComboBox` does not change theme when `SelectedIndex` is changed
 
--- a/Source/Krypton Components/Krypton.Toolkit/Palette Base/PaletteRedirect/PaletteRedirect.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Palette Base/PaletteRedirect/PaletteRedirect.cs
@@ -29,7 +29,7 @@ public class PaletteRedirect : PaletteBase, IGlobalId
     /// </summary>
     /// <param name="target">Initial palette target for redirection.</param>
     /// <remarks>target can be null when performing a drag-drop operation (Amongst other usages !!)</remarks>
-    public PaletteRedirect(PaletteBase? target)
+    public PaletteRedirect(PaletteBase? target) : base(attachUserPreferenceChanged: false)
     {
         Id = CommonHelper.NextId;
         // Remember incoming target
--- a/Source/Krypton Components/Krypton.Toolkit/Palette Builtin/Base/PaletteBase.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Palette Builtin/Base/PaletteBase.cs
@@ -25,8 +25,12 @@ public abstract class PaletteBase : Component
     internal Color[] GetSchemeColors() => SchemeColors;
 
     private Padding? _inputControlPadding;
+
     private PaletteDragFeedback _dragFeedback;
+    
     private Image[] _toolBarImages;
+
+    private readonly bool _systemEventsUserPreferenceSubscribed;
 
     private readonly Font _defaultFontStyle = new Font("Segoe UI", 9f, FontStyle.Regular);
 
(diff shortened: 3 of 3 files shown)


Smurf-IV (approved):
Can this go into v85 as well ?

PR #3466, opened 2026-05-15
V110 3447 feature request add missing toolbox bitmap images

PWagner1:


Smurf-IV (approved):
Nice, Can these go into 85 as well ?

PWagner1:
> Nice, Can these go into 85 as well ?

Will do this week, V110 will have a further part to it, dealing with the utility libraries.

PR #3488, opened 2026-05-22
* Fix Update canary-lts-release.yml

PWagner1:
* Fix Update canary-lts-release.yml to skip the WebView2 step


tobitege (commented):
If the SDK setup in CI is not working, won't the job still fail when the TestForm is compiled (or wherever the WebView is used)?

PWagner1:
> If the SDK setup in CI is not working, won't the job still fail when the TestForm is compiled (or wherever the WebView is used)?

Commented the job out, for use in V120 since V105 does not have a 'Utilities' binary.

tobitege (approved):
LGTM

PR #3492, opened 2026-05-23
* KryptonWebView2 enhancements

PWagner1:
* Use `new` instead of `override` for certain properties.


tobitege (approved):
New it shall be :)

PR #3498, opened 2026-05-23
* KryptonTreeView items flicker when selected/clicked (V105)

### LINKED ISSUE #3282: [Bug]: KryptonTreeView items flicker when selected/clicked
### Summary

KryptonTreeView items flicker when selected/clicked

### Description

KryptonTreeView items flicker when selected/clicked.
See video

### Steps to Reproduce

Start the treeview example in testform.
See video

### Expected Behavior

it should not flicker

### Actual Behavior

N/a/

### Operating System

N/A

### OS Version

current alpha

### Framework/.NET Version

Ran from IDE

### Toolkit Version

_No response_

### Additional Information

https://github.com/user-attachments/assets/ad8ecb0c-b450-4f3a-b133-2a94e2c4d185

### Areas Affected

Toolkit

PWagner1:
# Fix `KryptonTreeView` selection flicker (#3282)

## Summary

- Fixes [#3282](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3282): `KryptonTreeView` items visibly flicker when selected or clicked, especially with `OwnerDrawAll`, `MultiSelect`, `CheckBoxes`, and large `ItemHeight` (e.g. TestForm TreeView example).
- Ensures owner-drawn nodes skip native item painting, composites the control in a single off-screen `WM_PAINT` buffer, and batches selection/checkbox updates so deselect/select does not surface as separate on-screen paints.
- Defaults empty MSBuild `Configuration` to `Debug` so solution-wide `dotnet build` resolves correct `Bin\Debug\<tfm>\` output paths.

## Problem

`KryptonTreeView` uses an internal `TreeView` with `DrawMode = OwnerDrawAll`. On selection change, users could see a brief flash because:

1. Native TreeView item painting could still run unless `DrawDefault` is cleared (`CDRF_SKIPDEFAULT`).
2. Per-node off-screen `BitBlt` in `DrawNode` caused rows to update in separate steps during selection changes.
3. Per-node `ViewDrawPanel.Render` repainted row chrome on every item draw.
4. `BackColor` was assigned during `WM_PAINT`, triggering an extra erase/paint cycle.
5. `MultiSelect` toggled `Checked` during `OnAfterSelect` outside a single redraw batch.

## Solution

| Change | Purpose |
|--------|---------|
| `e.DrawDefault = false` in `OnTreeViewDrawNode` | Skip native selection/focus chrome for `OwnerDrawAll`. | | `InternalTreeView.WmPaint` off-screen bitmap | Paint Krypton background once, `DefWndProc` (custom draw) into same DC, single `BitBlt`. | | `TVS_EX_DOUBLEBUFFER` via `TVM_SETEXTENDEDSTYLE` | comctl32 buffering for owner-draw notifications. | | Draw directly on `e.Graphics` in `DrawNode` (no per-node bitmap) | Avoid tearing between old/new selection rows. | | Remove per-row `ViewDrawPanel.Render` in `DrawNode` | Background from composite `WmPaint` only. | | `SyncTreeViewBackColor()` from `UpdateStateAndPalettes` | Avoid `BackColor` changes during `WM_PAINT`. | | `BeginUpdate` / `EndUpdate` + `OnNeedPaint` guard | Batch selection redraws. | | Defer `MultiSelect` checkbox toggle to `AfterSelect` (before `EndUpdate`) | One redraw for check + selection. |

## Files changed

- `Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonTreeView.cs`
- `Source/Krypton Components/Krypton.Toolkit/General/PlatformInvoke.cs`
- `Documents/Changelog/Changelog.md`
- `Directory.Build.props`
- `Source/Krypton Components/Directory.Build.props`

## Test plan

- [ ] Build toolkit and TestForm (Debug): ```cmd dotnet build "Source\Krypton Components\Krypton Toolkit Suite 2022 - VS2022.sln" -c Debug dotnet run --project "Source\Krypton Components\TestForm\TestForm.csproj" -c Debug -f net48 ```
- [ ] **TreeView Example**: `CheckBoxes = true`, `MultiSelect = true`; click different nodes rapidly — no flicker.
- [ ] Set **`ItemHeight`** to **64** and repeat.
- [ ] **TreeView Test** form: compare `KryptonTreeView` vs standard `TreeView` if useful.
- [ ] Expand/collapse, scroll, keyboard navigation, focus in/out — no painting regressions.
- [ ] With `MultiSelect = true`, confirm click toggles checked state without extra flicker.

## Breaking changes

None.

## Related

- Closes #3282

image

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -83,15 +83,6 @@
 		</Otherwise>
 	</Choose>
 
-	<!-- MSB3822/3823: non-string .resx entries need preserialized resources + this package for the SDK GenerateResource task. -->
-	<PropertyGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true' AND '$(UseWindowsForms)' == 'true'">
-		<GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
-	</PropertyGroup>
-
-	<ItemGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true' AND '$(UseWindowsForms)' == 'true'">
-		<PackageReference Include="System.Resources.Extensions" Version="9.0.10" />
-	</ItemGroup>
-
 	<PropertyGroup Condition="'$(Configuration)' == 'Nightly'">
 		<DefineConstants>$(DefineConstants);NIGHTLY</DefineConstants>
 	</PropertyGroup>
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -47,6 +47,7 @@
 
 ## 2026-07-20 - Build 2607 (Version 105-LTS - Patch 3) - July 2026
 
+* Resolved [#3282](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3282), `KryptonTreeView` items flicker when selected/clicked (`DrawDefault = false` for `OwnerDrawAll`; composite `WM_PAINT` off-screen buffer with `TVS_EX_DOUBLEBUFFER`; batch selection/check updates; no per-node background repaint)
 * Resolved [#3451](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3451), Unable to add child controls into `KryptonHeaderGroup` at design time (`ReadOnly controls collection`). `DisplayRectangle` now reports the view fill rect, group container designers call `PerformLayout` after `EnableDesignMode`, and `KryptonGroupPanelDesigner.CanBeParentedTo` checks the parent component type correctly. Same `DisplayRectangle` improvement applied to `KryptonGroup` and `KryptonGroupBox`.
 * Resolved [#3367](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3367), ButtonSpec hover flicker on `KryptonTextBox`, `KryptonMaskedTextBox`, and `KryptonForm` (including `ImageStates.ImageNormal` without `Image`)
 * Resolved [#3383](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3383), White inner border artifacts appear on `KryptonButton` when hovering (`StateTracking` rounding issue)
--- a/Source/Krypton Components/Directory.Build.props
+++ b/Source/Krypton Components/Directory.Build.props
@@ -20,6 +20,15 @@
 		<Version>$(LibraryVersion)</Version>
 	</PropertyGroup>
 
+	<!-- MSB3822/3823: non-string .resx entries need preserialized resources + this package for the SDK GenerateResource task. -->
+	<PropertyGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true'">
+		<GenerateResourceUsePreserializedResources>true</GenerateResourceUsePreserializedResources>
+	</PropertyGroup>
+
+	<ItemGroup Condition="'$(UsingMicrosoftNETSdk)' == 'true'">
+		<PackageReference Include="System.Resources.Extensions" Version="9.0.10" />
+	</ItemGroup>
+
 	<!-- Configure NuGet package output to separate directory for easier management -->
 	<PropertyGroup>
 		<PackageOutputPath>$(MSBuildThisFileDirectory)..\..\Bin\Packages\$(Configuration)\</PackageOutputPath>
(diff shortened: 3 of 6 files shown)


tobitege:
Tried to verify this PR locally, ran into many errors, did some digging.

Better explained by Gpt 5.5:
---
I found that the build failure was caused by all target frameworks sharing the same intermediate `obj\Release` folder. The final DLLs were already going into separate `Bin\Release\<target-framework>` folders, but generated reference assemblies, resources, and caches were still written to the same intermediate paths. That meant a later target framework such as `net9.0-windows` or `net10.0-windows` could overwrite `Krypton.Toolkit` reference files that `net8.0-windows` was about to consume, which produced the CS1705 version mismatch errors.

The fix is to make the intermediate output path target-framework-specific as well. Each framework now writes its generated files under its own `obj\<Configuration>\<TargetFramework>` folder, so the framework-specific reference assemblies and resource outputs no longer overwrite each other during the multi-targeted Stable build.

---

As result the build is running without errors and clean output paths:
{FBEA6F8F-6997-4D1D-88FF-552973DFAC6F}

tobitege:
Added commit `3dd5d92604e128a6855b8c26aadb036ea3fef7d0` to address two `KryptonPropertyGrid` crashes found while validating the TreeView forms.

Reason for the change:
- `KryptonPropertyGrid` was assigning palette-owned `Font` instances directly to the embedded WinForms `PropertyGrid` and its child controls. Palette fonts can be disposed/recreated during palette updates, which left WinForms internals with an invalid font and caused `Font.GetHeight()` to throw during handle/focus processing.
- The focused embedded child-control path used `PaletteState.FocusOverride` for input-control palette back/content lookups. `KryptonCustomPaletteBase.GetPaletteInputControl()` does not support that state for input controls, so holding Tab through the TreeView sub-form could throw while updating child control colors on focus changes.

The commit now clones/cache-disposes the fonts owned by `KryptonPropertyGrid` itself, and uses the supported active input-control state (`Tracking`) for focused embedded child controls.

---

I ran the TestForm again after each of the 2 fixes and the exceptions were gone.

tobitege (approved):
Tried the PR locally and could not reproduce any flicker.

PWagner1:
> Added commit `3dd5d92604e128a6855b8c26aadb036ea3fef7d0` to address two `KryptonPropertyGrid` crashes found while validating the TreeView forms.
> 
> Reason for the change:
> 
> * `KryptonPropertyGrid` was assigning palette-owned `Font` instances directly to the embedded WinForms `PropertyGrid` and its child controls. Palette fonts can be disposed/recreated during palette updates, which left WinForms internals with an invalid font and caused `Font.GetHeight()` to throw during handle/focus processing.
> * The focused embedded child-control path used `PaletteState.FocusOverride` for input-control palette back/content lookups. `KryptonCustomPaletteBase.GetPaletteInputControl()` does not support that state for input controls, so holding Tab through the TreeView sub-form could throw while updating child control colors on focus changes.
> 
> The commit now clones/cache-disposes the fonts owned by `KryptonPropertyGrid` itself, and uses the supported active input-control state (`Tracking`) for focused embedded child controls.
> 
> I ran the TestForm again after each of the 2 fixes and the exceptions were gone.

Hi @tobitege 

Does this need to go into V110 as well?

tobitege:
I'm not quite sure. In alpha there's a bit of a different font handling, a different kind of workaround.

giduac:
@tobitege 
Have look at this one:
https://github.com/Krypton-Suite/Standard-Toolkit/issues/3282

tobitege:
> @tobitege Have look at this one: #3282

Why? I reviewed and approved this PR as the flickering issue itself is resolved.
However, the TestForm ran into CTD's in secondary class, which got fixed, too.

giduac:
all in the clear then :)

tobitege:
Added commit `dc99b9a63cc7ab1a324b44ae66ae7a2db710709e` to fix the remaining TestForm CI resource errors.

Plain explanation:

The earlier fix put `GenerateResourceUsePreserializedResources=true` and the `System.Resources.Extensions` package reference in the repo-root `Directory.Build.props`, guarded by `UseWindowsForms=true`.

That did not reliably help the TestForm build, because `Directory.Build.props` is imported very early. At that point, `UseWindowsForms` had not been set yet by `Krypton.Toolkit 2022.csproj`, so the condition evaluated false and the resource workaround was skipped. The result was that CI still built `Krypton.Toolkit` for `net472` without `GenerateResourceUsePreserializedResources`, which produced the `MSB3823` errors for non-string `.resx` entries.

The new commit moves the workaround into `Source/Krypton Components/Directory.Build.props`, which is the shared props file imported by the toolkit projects and TestForm path. It also removes the too-early `UseWindowsForms` condition and only checks that the project is SDK-style. That makes the setting visible to `Krypton.Toolkit` when TestForm builds it as a referenced project.

Validation:

- Full toolkit build was reported passing locally.
- `TestForm.csproj -c Debug -f net472` passed locally.
- `TestForm.csproj -c Debug` passed locally across the configured target frameworks.

tobitege:
@PWagner1 good to go, build is fixed now :)

tobitege (approved):

PR #3501, opened 2026-05-24
3493-V110 Limit VS2022/26 target frameworks to supported ones

### LINKED ISSUE #3493: [Bug]: VS 2022 build scripts shall skip .NET 11 SDK
### Summary

VS 2022 build scripts fail with .NET 11 SDK preview

### Description

VS menu builds pass `TFMs=all`, and `TFMs=all` currently includes `net11.0-windows`.

When the VS 2022 or current VS MSBuild path resolves SDK-style projects to a .NET 10 SDK, the build fails with `NETSDK1045` because SDK 10 cannot target `.NET 11`.

The fix should split the TFM sets: keep `net11.0-windows` only for an explicit preview/.NET 11-capable build path, and have VS 2022 scripts use a stable TFM set ending at `net10.0-windows`.

### Steps to Reproduce

- run.cmd
- VS 2022
- Rebuild
Compile errors if .NET SDK 11 preview is installed

### Expected Behavior

Leave out 11

### Actual Behavior

NETSDK1045 errors

### Operating System

Windows 11

### OS Version

25H2, latest

### Framework/.NET Version

10

### Toolkit Version

10

### Additional Information

_No response_

### Areas Affected

None

tobitege:
Fixes #3493 

## Summary

This follow-up limits the VS2022 and VS2026 build path to target frameworks that Visual Studio 2022 officially supports.

- Adds a clearer `ExcludeVs2022UnsupportedTargetFrameworks` property.
- Uses that property for VS2022 builds instead of the narrower `ExcludeNet11` switch.
- Removes `net10.0-windows` and `net11.0-windows` from the VS2022-selected target framework lists.
- Keeps `net10.0-windows` and `net11.0-windows` available for the current / VS2026 build path.
- Split VS2022 and Current target selection logic so `ExcludeNet11=true` only removes `net11.0-windows`, while VS2022 still stops at `net9.0-windows`.
- Added `ExcludeNet11=true` to `Scripts\Current\*.proj` so VS2026 builds target up to `net10.0-windows` and avoid the local SDK `net11.0-windows` failure.
- Added restore-before-clean calls in `Scripts\Current\*.proj` so stale `project.assets.json` files from a different target set do not break `Clean` with `NETSDK1005`.
- Guarded `rd` cleanup calls in `run.cmd` and purge scripts with directory `if exist ...\` checks so missing folders do not print “file not found” messages.

## Why

The previous fix removed `net11.0-windows` for VS2022 builds, but VS2022 17.14 still warns on `net10.0-windows` with `NETSDK1233`.

Microsoft's .NET SDK / Visual Studio versioning documentation says:

- targeting `net9.0` is officially supported in Visual Studio 17.12+;
- targeting `net10.0` is officially supported in Visual Studio 18.0+;
- using a newer SDK inside an older Visual Studio version is allowed, but targeting newer runtimes is unsupported and produces a build warning.

Reference: https://learn.microsoft.com/en-us/dotnet/core/porting/versioning-sdk-msbuild-vs

{08DFB37A-9E60-4384-ABC7-54F1CC152EB1}

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -45,6 +45,7 @@
 
 ## 2026-11-xx - Build 2611 (V110 Nightly) - November 2026
 
+* Resolved [#3493](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3493), Fix Scripts build issues related to framework targeting
 * Enhanced `KryptonWebView2` palette integration. `BackColor`, `ForeColor`, and `DefaultBackgroundColor` are now driven from the active Krypton palette via `StateCommon`, `StateNormal`, `StateActive`, and `StateDisabled` (with `WebViewBackStyle` / `WebViewContentStyle`). Legacy appearance properties and their change events are hidden from the designer; use the `State###` entries under **Visuals** instead.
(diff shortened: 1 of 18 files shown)


tobitege:
Latest commit added similar fixes for the VS 2026 build scripts as well as suppressing "not found" output in Clear.
Added explicit "restore" step before "clear" to avoid stale "obj" content from a previous build.

PWagner1 (approved):

PR #3513, opened 2026-05-25
Fix RichTextBox cue hint build regressions across target frameworks

tobitege:
This fixes two build regressions caused by recent `KryptonRichTextBox` cue hint changes.

The first issue was that `KryptonRichTextBox.cs` used `ReadOnlySpan<char>` and related span APIs in helper methods. Those APIs are not available in the .NET Framework targets used by this branch, so builds for targets like `net481` failed. The helper code now uses string-based logic instead.

The second issue was that `Krypton.Ribbon` still expects `KryptonRichTextBox` to expose `RichTextBoxButtonSpecCollection` through `ButtonSpecs`. That type and property had been removed from the toolkit control, which broke `KryptonRibbonGroupRichTextBox`. The button spec collection and related manager wiring have been restored.

Verified builds:
- `Krypton.Toolkit 2022.csproj` for `net481`
- `Krypton.Toolkit 2022.csproj` for `net6.0-windows`
- `Krypton.Ribbon 2022.csproj` for `net6.0-windows`

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonRichTextBox.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonRichTextBox.cs
@@ -284,9 +284,29 @@ public class KryptonRichTextBox : VisualControlBase,
     }
     #endregion
 
+    #region Type Definitions
+    /// <summary>
+    /// Collection for managing ButtonSpecAny instances.
+    /// </summary>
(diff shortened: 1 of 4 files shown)


PWagner1 (approved):

tobitege:
Good morning @PWagner1 
please have a look at those 4 files, if you're ok with the implementation.
I didn't look at the mountain of pre-existing warnings in this PR, though.

PWagner1 (approved):

tobitege:
closed this since other merged PR resolved same errors

PR #2422, opened 2025-08-16
2418 V100 KryptonDataGridView: unify header painting

tobitege:
Fixes #2418 

### Summary of changes

- KryptonDataGridView
  - Header text now rendered using `TextRenderer.DrawText` with strict clipping to `GetColumnDisplayRectangle(...)`.
  - Preserves ClearType and correct clipping for all columns, including the last column.
  - Background, border, sort glyphs, and header icons remain rendered via the Krypton renderer.
  - Maintains `_columnCache` behavior for header tooltips.

- TestForm/DataGridViewDemo
  - Existing button did not have a demo behind it, now it does.

PR #3516, opened 2026-05-25
2418-V110-KDGridView title fix

### LINKED ISSUE #2681: [Bug]: KryptonDataGridView ( Problem in ColumnHeaders when scrolling  )

tobitege:
Fixes #2681.

The likely cause is in the #2422 header painting change. That change says header text is clipped to the column display rectangle, but the code called GetColumnDisplayRectangle(e.ColumnIndex, false). In DataGridView, false returns the entire column rectangle, while true returns only the visible rectangle.

That means partially visible column headers during horizontal scrolling could still use overflow geometry while the header background and border were painted through the separate offscreen path. This updates the call to request the visible rectangle so the direct TextRenderer.DrawText path clips to the same visible area that the scroll operation exposes.

The existing #2681 scroll invalidation workaround is intentionally left in place. This change addresses the clipping mismatch underneath it, and the workaround can be removed separately after the fast horizontal scrollbar-drag repro is checked visually.

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridView.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonDataGridView.cs
@@ -1709,7 +1709,7 @@ public class KryptonDataGridView : DataGridView
                                 if (!string.IsNullOrEmpty(text))
                                 {
                                     Rectangle constrainedContentBounds = headerContentBounds;
-                                    Rectangle colDisplayRect = GetColumnDisplayRectangle(e.ColumnIndex, false);
+                                    Rectangle colDisplayRect = GetColumnDisplayRectangle(e.ColumnIndex, true);
                                     if (colDisplayRect.Width > 0)
                                     {
                                         // Translate display rect into cell-local coords


tobitege:
@giduac @PWagner1 - looks like a single param needed switching.
Can't repro the column header painting issue on fast scrolling anymore.

giduac:
> @giduac @PWagner1 - looks like a single param needed switching. Can't repro the column header painting issue on fast scrolling anymore.

@tobitege 
With the work-around turned off ?

KamaniAR:
Look at column header


https://github.com/user-attachments/assets/7d5b4188-d4bc-41b1-be4a-47ff05d204e8

giduac:
@Krypton-Suite/reviewers 

With the scroll fix ON all is hunky dory.

With the scroll fix off the old behaviour still exists:

https://github.com/user-attachments/assets/70c18c2f-c0a9-4ce2-88d9-8d7a7e4271d4

tobitege:
@giduac at least right now it seems to work with both patches in place.
I'll do one more analysis run to maybe identify, what was wrong before the scroll fix came in.

tobitege:
Closing this. The existing OnScroll workaround seems like the best practical option for the issue.

PR #3539, opened 2026-05-27
FAIL TestForm layout sizing fixes-V110

tobitege:
## Summary
- Reworked the TestForm StartScreen layout to use a root TableLayoutPanel instead of independent sibling anchors.
- Put the filter, command-link list, and Exit button in a left fill column.
- Put the theme list in a fixed-width right column with a fixed gap between the columns.
- Docked the theme list in its column so it stays aligned with the right side as the form resizes.
- Anchored command-link buttons left and right so they resize with the launcher list.
- Set the StartScreen minimum width to 400 px.

## Why
The old layout relied on separate anchors on controls that were siblings on the form. That made the relationship between the left-side controls and the theme list implicit, so the left area could resize independently of the theme list. The new table layout gives the theme list its own reserved column and keeps the left side from growing into it.

## Validation
- dotnet build "Source/Krypton Components/TestForm/TestForm.csproj" -c Debug -f net472


tobitege:
Doh! Wrong branch... grrrr

Issue #1268, opened 2024-01-05
[Bug]: Many Krypton Controls have a `CornerRoundingRadius` that overrides the State#### Node Rounding values. Please remove!


Smurf-IV:
- `CornerRoundingRadius` overrides **ONLY** the `StateCommon.Border.Rounding` which is incorrect.
  - All `CornerRoundingRadius` have been removed e.g
  - Also reverse #877

![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/ad8a4f49-13cc-4663-aa2c-de63cd8deab5)

AS Rounding already exists in each of the `State####` for KTrees:
![image](https://github.com/Krypton-Suite/Standard-Toolkit/assets/2418812/48a80f35-b1ea-4817-b73b-4bb956719048)

Smurf-IV:
Similar to #1255
@Wagnerp thoughts ?

Wagnerp:
@Smurf-IV Yes, will go through every control.

Issue #519, opened 2021-12-07
[Feature Request]: `PaletteDataGridViewAll` does not expose the border value of the `private readonly PaletteDouble _background`


Smurf-IV:
Need to be able to set the border for focus / disabled etc.

And the following class does not expose the Border part of the PaletteDouble:

![image](https://user-images.githubusercontent.com/2418812/145039087-178e9293-d478-404b-8dd1-6db157ebbaae.png)

PWagner1:
Expose it in the constructor

Smurf-IV:
> @Smurf-IV & @giduac
> 
> I don't understand, `IPaletteDouble` is never assigned, should I remove?


No derived classes assign values as needed to it:
![image](https://github.com/user-attachments/assets/44cac8f1-9769-495d-b9c1-ccb81a1892b2)

PR #3567, opened 2026-05-28
* KryptonDataGridView external corner rounding

### LINKED ISSUE #804: [Feature Request]: Datagridview, add external rounding
Can we get the option to add rounding to the external corners of the overall datagridview
image

PWagner1:
# KryptonDataGridView external corner rounding (#804)

## Summary

Implements [issue #804](https://github.com/Krypton-Suite/Standard-Toolkit/issues/804) — optional rounded **external** corners on `KryptonDataGridView`, aligned with how other Krypton controls expose border rounding through the palette.

Rounding is controlled via **`StateNormal.Border.Rounding`** (and **`StateDisabled.Border`** for the disabled state). Use `-1` to inherit theme defaults (grid border styles resolve to `0`, so rounding is off unless explicitly set). Any value `> 0` enables corner rounding in logical pixels.

### `KryptonDataGridView`

- **Region clipping** — control clip region follows a rounded rectangle when rounding is active.
- **Outer border** — painted in `PaintBackground` using `StateNormal.Border` / `StateDisabled.Border` and `RenderStandardBorder`, sized to the grid cell area (scrollbar-aware).
- **Corner cells** — outer corner cells draw rounded backgrounds and borders via `PaletteBorderInheritForced.ForceBorderRounding` and restricted border edges (`Top|Left`, `Top|Right`, etc.).
- **Perimeter cells** — non-corner outer cells suppress edges that would double-draw against the outer frame (same pattern as `HideOuterBorders`).
- **Scrollbar awareness** — `GetGridCornerBounds()` excludes visible vertical/horizontal scrollbar gutters when detecting corners and drawing the outer border.
- **Constructor safety** — corner-rounding logic guards against `OnLayout` running before `StateNormal` / `StateDisabled` exist (fixes `NullReferenceException` during designer/ctor layout).

### `PaletteDataGridViewAll`

- Exposes **`Border`** in the designer under **Visuals** (consistent with `KryptonListView`, `KryptonButton`, etc.).
- **`PaletteBorder`** remains for code compatibility but is hidden from the designer (`Browsable(false)`).

### `PaletteBorderInheritForced`

- **`ForceBorderRounding(float)`** — per-cell rounding override for corner painting.
- **`ClearForcedState()`** — resets forced edges, rounding, and graphics hint before each cell paint.

### TestForm

- **DataGridView demo** — **Corner Rounding** checkbox toggles `StateNormal.Border.Rounding` / `StateDisabled.Border.Rounding` (`5` / `-1`).

### Changelog

- V110 nightly entry for #804.

## API / designer usage

```csharp
// Enable (5px radius)
kryptonDataGridView1.StateNormal.Border.Rounding = 5f;
kryptonDataGridView1.StateDisabled.Border.Rounding = 5f;

// Disable
kryptonDataGridView1.StateNormal.Border.Rounding = -1f;
kryptonDataGridView1.StateDisabled.Border.Rounding = -1f;
```

**Designer path:** `KryptonDataGridView` → **StateNormal** → **Border** → **Rounding**

Other border properties (`Draw`, `DrawBorders`, `Width`, `Color1`, etc.) on the same node apply to the outer frame.

## Files touched

| File | Change |
|------|--------|
| `KryptonDataGridView.cs` | Corner rounding paint, region, grid bounds, corner cell logic, ctor/layout guards | | `PaletteDataGridViewAll.cs` | `Border` designer property; hide `PaletteBorder` from designer | | `PaletteBorderInheritForced.cs` | Forced rounding + clear state | | `DataGridViewDemo.cs` / `.Designer.cs` | Demo checkbox | | `Documents/Changelog/Changelog.md` | Release note |

## Test plan

- [ ] Build `Krypton.Toolkit` (at least `net48` and one `-windows` TFM used locally).
- [ ] Run **TestForm** → **DataGridView demo**.
- [ ] Toggle **Corner Rounding** — grid shows rounded outer corners; no crash on open.
- [ ] With **ScrollBars = Both** and enough rows/columns to show scrollbars, confirm bottom-right rounding aligns with the last cell (not the scrollbar corner square).
- [ ] Toggle **Show Grid Lines** / **HideOuterBorders** behaviour — no double outer frame with rounding on.
- [ ] Change theme via theme combo — border colours follow palette; rounding still applies.
- [ ] Set **StateNormal.Border.Rounding** in designer, save form, reload — value persists and applies at runtime.
- [ ] Optional: RTL checkbox — corners still look correct on physical edges.
- [ ] Optional: disabled grid — `StateDisabled.Border.Rounding` used when `Enabled = false`.

## Breaking changes

None intended.

- No standalone `CornerRounding` property (uses palette `Border.Rounding` like other controls).
- `PaletteDataGridViewAll.PaletteBorder` is still available in code; designer now prefers **`Border`**.

## Related issues

- Closes #804
- Related: #519 (grid background border exposure), #1268 (suite direction away from per-control `CornerRoundingRadius`)


tobitege:
The top-right corner is being rounded at the grid/content edge, just before the vertical scrollbar. But the actual control region is rounded around the full `ClientRectangle`, including the scrollbar area. So there are two competing “right edges”:

- cell/border rounding uses bounds excluding the scrollbar
- clipping uses bounds including the scrollbar

That is why the top-right corner looks inconsistent once the vertical scrollbar appears. The fix should make the rounded border, corner-cell detection, and clipping all use the same rectangle, most likely the full control bounds if the scrollbar is meant to be inside the rounded KDGV frame.

{50306FC8-B08A-4862-A8AD-DB6A4F02D577}


In theory, the fix is to choose one rectangle as the single source of truth for the rounded outer shape.

For this control, the better choice is probably the full `ClientRectangle`, including scrollbars, because the scrollbar is part of the visible `DataGridView` surface. Then:

1. `UpdateCornerRoundingRegion()` should clip using that same full rectangle.

2. The outer rounded border drawn in `PaintBackground()` should use that same full rectangle.

3. `TryGetOuterCornerBorderEdges()` / `GetCellMaxBorderEdges()` should not treat the cell before the scrollbar as the outer top-right corner. With a vertical scrollbar visible, the last visible data/header cell is not actually at the control’s outer right edge.

4. Cell rounding should only apply to real outer cells touching the same outer rectangle. If the scrollbar occupies the right edge, the column header next to it should keep normal square/internal borders on its right side.

So practically, remove or change the scrollbar subtraction in `GetGridCornerBounds()`. If scrollbar overlap causes painting artifacts, handle scrollbars separately, but don’t let cell-corner logic and region clipping disagree about where the outer corner is.

tobitege (changes requested):
Rounded corners problematic with scrollbar present, haven't tested RTL.

PWagner1:
Hi @tobitege 

Please try the latest changes. Scrollbars are now 'detached' when rounding is in effect so they don't become rounded.

tobitege (approved):
Looks good - can't test right now (meetings) 😄

tobitege:
GPT-5.5 says:

One thing worth checking in the detached scrollbar logic: the viewport size currently subtracts the opposite scrollbar size based on whether that scrollbar is allowed, not whether it is actually visible.

For example, with ScrollBars.Both, GetDetachedScrollViewportHeight() subtracts the horizontal scrollbar height even when the horizontal scrollbar may not be shown. Near the threshold, that can make the vertical detached scrollbar appear even though the rows would fit without it. The same applies in the other direction for GetDetachedScrollViewportWidth().

A safer approach would be to base those viewport reductions on the actual visibility decision for the opposite detached scrollbar, so each scrollbar only reserves space when the other one is really present.

Issue #3181, opened 2026-03-10
[Feature Request]: Can we have retro theme(s)


PWagner1:
### Feature Title

Can we have retro theme(s)

### Feature Description

See the themes at the following:
https://github.com/AlexanderSharykin/RetroUI

This is a bit of fun and low priority, but might find use for Tablet style apps (i.e. low resolution)

https://github.com/Krypton-Suite-Legacy-Archive/Krypton-Toolkit-Suite-Extended-NET-5.470/issues/54

tobitege:
The feature is in basically. Minor PR followups may reference this later on. Closing for now.

PR #3592, opened 2026-05-29
* Retro themes

PWagner1:
Implements #3181

Probably another can of worms, but the 2 retro ones look like old days, almost. 😄

## Summary
- Adds RetroGreen, RetroBlue, and VisualStudio2022Dark theme support.
- Adds retro palette bases, schemes, renderer support, and shared retro assets.
- Updates palette mode conversion and string handling for the new modes.

Programs using this need to be designed for these retro themes as they take 8px as shadow border for the effect, thus certain controls like buttons may need to be taller, accordingly.

{8F1FF8B3-19C7-4109-AA11-ADD0102AD6C7}

{55C3C537-255F-4364-BDF2-F79D6E55B411}

Should I continue this? @PWagner1 @giduac et al


tobitege:
Closing this, #3596 is the correct one.

PR #3594, opened 2026-05-30
3493-V110-Build fix followup

tobitege:
Regression followup fixes to #3493 and #3494 

- Prevents VS 2026 local builds from evaluating unsupported net11 targets when only a net10-capable SDK is available.
- Keeps CI and explicit all-TFM builds able to include net11 unless the fallback is requested.
- Removes duplicate package helper item warnings for the System.Resources.Extensions props and targets files.
- Fixes Current script paths so local build/debug commands find the .NET SDK setup helper.

I tested the "run.cmd" nightly builds for both VS 2022+2026, incl. debug build of TestForm.
Then also did restore, clean and full rebuild of solution in both VStudio 2022 and 2026 and both built fine.
In VS 2022 only a single warning shows up due to .net 10, but that is expected for v17.


PWagner1 (approved):

tobitege:
Switching to draft for now.

PWagner1:
@tobitege 

The conflicts may be as a result of your previous fixes from a few days ago.

tobitege:
Yeh, makes sense, I'll get it working asap :)

PR #3623, opened 2026-05-31
1976-V105-LTS-Fix MenuItem theme settings

### LINKED ISSUE #1976: [Bug] Theme settings for "MenuItem###" do not appear to work
And also these do not appear to work either !
image

_Originally posted by @Smurf-IV in https://github.com/Krypton-Suite/Standard-Toolkit/issues/638#issuecomment-2551110762_

tobitege:
Fixes #1976

This updates ToolStrip menu item rendering so custom `ToolMenuStatus.Menu.MenuItem###` colors are used for selected, pressed, and border states.

The renderer now handles custom MenuItem color-table overrides before falling back to the existing renderer behavior, including the border-only case where the normal highlight background should remain visible.

A focused WinForms repro harness was added under `Source/TestHarnesses/MenuItemThemeSettingsRepro` for checking full MenuItem colors and border-only overrides.

### DIFF
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -47,6 +47,7 @@
 
 ## 2026-07-20 - Build 2607 (Version 105-LTS - Patch 3) - July 2026
 
+* Resolved [#1976](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1976), `MenuItem###` theme settings now apply to menu item selected, pressed, and border rendering.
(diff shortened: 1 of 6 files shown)


tobitege:
Visual confirmation with the help of the added harness demo (label text about magenta will be fixed separately, there is none):

{5F60F14A-ADF3-433B-899B-4D5E8EB9A6DC}


{80C4BC06-C1A4-4ABA-8308-67C951EC6F62}

PWagner1 (approved):
Tested using the demo, works as expected.

PR #3642, opened 2026-06-01
3641-V110-restore-catchup

### LINKED ISSUE #3641: V110 restore alpha catchup
## Summary

`alpha` is currently reported by GitHub as behind `master` because the restored `master` history contains five commits that are not reachable from `alpha`.

Those commits cancel out to no net workflow diff in the compare view, but GitHub still counts the missing commit ancestry.

## Proposed fix

Create a branch from `alpha` and merge `master` into it so `alpha` catches up to the restored `master` ancestry.

This is expected to be a history catch-up change and may have no effective file diff.

tobitege:
Fixes #3641

## Summary

Merges `master` into `alpha` so the restored `master` commits become reachable from the V110 development branch.

The five `master` commits cancel out to no net file diff, but GitHub still counts them as missing ancestry on `alpha`.

## Notes

This PR is expected to be a history catch-up PR with no effective file changes.

### DIFF
--- a/.github/workflows/templates-release.yml
+++ b/.github/workflows/templates-release.yml
@@ -206,7 +206,7 @@ jobs:
 
       - name: Upload workflow artifact
         if: steps.templates_release_kill_switch.outputs.enabled == 'true'
-        uses: actions/upload-artifact@v7
+        uses: actions/upload-artifact@v4
         with:
           name: templates-${{ steps.channel.outputs.channel }}
           path: |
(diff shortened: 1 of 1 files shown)


tobitege:
@PWagner1 what do you think about this approach?

Also, edge case due to empty files:
{7F0A1B69-5E83-464A-8CF6-31EDED59022E}

PWagner1:
> @PWagner1 what do you think about this approach?

Looks good to me, can it be replicated on other branches, also is it possible to add docs for future use?

tobitege:
In principle it should be the same kind of issue between master and the other branches, I guess.
I just told Gpt to create issue and PR (after discussing the actual issue, ofc) and here it was :)

PWagner1 (approved):

tobitege:
I'm an idiot! I shouldn't have squashed them, doh

Issue #2926, opened 2026-01-26
[Bug]: Scrollbar issues


PWagner1:
### Summary

Scrollbar issues

### Description

Some controls use both the `KryptonScrollBar` and the standard Win32 scrollbar.

<img width="314" height="510" alt="Image" src="https://github.com/user-attachments/assets/bd307aff-4933-48fb-bf86-9b6b874dc126" />

### Steps to Reproduce

Set `UseKryptonScrollbars` in `KryptonManager`
Use any control

### Expected Behavior

Use either `KryptonScrollBar` or the standard Win32 version.

### Actual Behavior

Some controls use both versions.

### Areas Affected

Toolkit

PR #3651, opened 2026-06-02
2926-V110-Scrollbar display fix followup

tobitege:
Follow-up to issue #2926.

This adjusts the themed vertical scrollbar tracking for ListBox controls so the calculated position follows the actual fully visible list items. It uses ListBox item rectangles from the current TopIndex, which keeps the top and bottom scrollbar states aligned with the top and bottom list items.

This also fixes KryptonThemeListBox/KryptonListBox mouse-wheel scrolling when Krypton scrollbars are enabled. Wheel input over the list surface now updates TopIndex directly when native hidden-scrollbar handling does not move the list, then refreshes the Krypton scrollbar state.

The existing ItemHeight-based calculation remains as a fallback when item rectangles are unavailable.


KamaniAR:
If related to this look at Mouse scroll.

Mouse scroll is not working in StartScreen -> Theme List.

tobitege:
> If related to this look at Mouse scroll.
> 
> Mouse scroll is not working in StartScreen -> Theme List.

I forgot about that, only works when over the scrollbars, not over the list.
Will have a look at that, now.

tobitege:
There are two scrolling fixes.

`7f7550e4e` fixes the themed scrollbar position.  
The old code estimated visible rows with `ClientSize.Height / ItemHeight`. That can be wrong when the last row is only partly visible. The manager now checks actual item rectangles, so top/bottom scrollbar state matches the real visible list items.

`60f6af389` fixes mouse-wheel scrolling over list items.  
With Krypton scrollbars enabled, native scrollbars are hidden. The wheel event can reach the list but native scrolling may not move `TopIndex`. Now the list reports the wheel event to `KryptonScrollbarManager`.

The manager then:
- does nothing extra if native scrolling already worked
- scrolls `TopIndex` itself if native scrolling did not move
- accumulates small wheel deltas correctly
- refreshes the Krypton scrollbar after the change

So `KryptonListBox` detects the wheel event; `KryptonScrollbarManager` owns the scrolling fix.

PWagner1 (approved):

Issue #3672, opened 2026-06-04
[Feature Request]: Drop support for `SLN` and move to `SLNX` solution files


giduac:
### Feature Title
Drop the support for sln style files.
V105 & V110

VS2019 cannot deal with the newer frameworks.
Therefore V85 will retain the SLN file.

tobitege:
This is not a good idea! As long as we support .NET framework 4.7+ we have to assume that Visual Studio 2019 is also still in use, which does NOT support .slnx format!
@giduac @PWagner1

giduac:
> This is not a good idea! As long as we support .NET framework 4.7+ we have to assume that Visual Studio 2019 is also still in use, which does NOT support .slnx format! [@giduac](https://github.com/giduac) [@PWagner1](https://github.com/PWagner1)

@Krypton-Suite/reviewers 

Valid point!
[VS2019 is serviced until 2029](https://learn.microsoft.com/en-us/visualstudio/releases/2019/servicing-vs2019)

Smurf-IV:
sln is Only applicable to V85
The others because of the TFM's that are not loadable into VS19 can go to slnx.

giduac:
Findings from all added to the ticket description.

tobitege:
No, that is not the issue! 
One should still be able to compile V110 with VS 2019 for .net 4.7.x, no? 
Thus a ".sln" would still be required and should not just be removed.

Locally, with small tweaks, I run a full rebuild within VS 2019 for `alpha` (sorry for German words):
```
1>------ Neues Erstellen gestartet: Projekt: Krypton.Toolkit 2022, Konfiguration: Debug Any CPU ------
1>Der Buildvorgang wurde am 05.06.2026 11:43:12 gestartet.
1>Ziel "Clean":
1>Ziel "DispatchToInnerBuilds":
1>    Ziel "CopyFilesToOutputDirectory":
1>        Krypton.Toolkit 2022 -> D:\github\Standard-Toolkit\Bin\Debug\net472\Krypton.Toolkit.dll
1>    Ziel "CopyFilesToOutputDirectory":
1>        Krypton.Toolkit 2022 -> D:\github\Standard-Toolkit\Bin\Debug\net48\Krypton.Toolkit.dll
1>    Ziel "CopyFilesToOutputDirectory":
1>        Krypton.Toolkit 2022 -> D:\github\Standard-Toolkit\Bin\Debug\net481\Krypton.Toolkit.dll
1>
[...]

PR #3674, opened 2026-06-04
* Drop support for `SLN` and move to `SLNX` solution files (V110)

PWagner1:
* Implemented #3672 - V110


tobitege:
I haven't tried, but would VS 2019 work with .slnx? Since we're still supporting .NET 4.7.2+

tobitege:
Closed in favor of #3704. The old PR targets `zzz-alpha-20260607`, which belongs to the archived zzz branch line; #3704 is the clean replacement targeting `alpha`.

PR #3675, opened 2026-06-04
* Drop support for `SLN` and move to `SLNX` solution files (V105)

### LINKED ISSUE #3672: [Feature Request]: Drop support for `SLN` and move to `SLNX` solution files

PWagner1:
* Implemented #3672 - V105


tobitege:
Can we close this so VS 2019 will still be usable?

PWagner1:
> Can we close this so VS 2019 will still be usable?

Yes, but keep the issue open for a future version

PR #3673, opened 2026-06-04
* Harden workflows

PWagner1:
# Harden workflows: branch policy and `.github` sync from master

Closes #3610

## Summary

Introduces a **warn-then-fail** pull request branch policy and an automated **Sync .github from master** workflow so CI/workflow changes can propagate to release lines **by file content** (the `.github/` tree), without requiring full merges of `master` or Git ancestry checks on downstream branches.

## Motivation

Release branches (`alpha`, `canary`, LTS lines, etc.) intentionally diverge in product code. Maintainers still need:

- Controlled propagation of **GitHub Actions and `.github` config** from `master`
- CI feedback on **invalid PR base/head pairs** (e.g. `alpha` → `master`, or `master` → `alpha` with `Source/` changes)
- A rollout path that **warns first**, then enforces via `BRANCH_POLICY_ENFORCE`

**Not in scope:** requiring every release line to contain all commits from `master` in Git history. That was considered and **rejected** as brittle and unrelated to file-based workflow sync.

### Automation

| Component | Purpose |
|-----------|---------|
| `.github/workflows/pr-branch-policy.yml` | Validates same-repo PRs on open/sync/reopen | | `.github/scripts/Invoke-BranchPolicyCheck.ps1` | Rule engine (PowerShell) | | `.github/workflows/sync-github-from-master.yml` | Opens PRs: `git checkout origin/master -- .github` per target branch | | `.github/branch-policy.json` | Branch lists and rule metadata | | `.github/BRANCH_POLICY.md` | Short maintainer cheat sheet |

### Policy rules (enforced)

| Rule ID | Behaviour |
|---------|-----------|
| `master-github-downstream` | PR `master` → downstream base may only change paths under `.github/` | | `alpha-to-alpha-backup` | PR into `alpha-backup` must have head `alpha` | | `no-long-lived-to-master` | PR from a long-lived release line into `master` is blocked |

### Sync targets (`syncGithubFromMasterTargets`)

`alpha`, `canary`, `gold`, `prerelease`, `V105-LTS`, `V85-LTS`, `V110`

Sync uses a **path checkout** on the target branch tip, not `git merge master`.

### Documentation

| Document |
|----------|
[...]

PR #3705, opened 2026-06-07
* Harden workflows (V110)

### LINKED ISSUE #3610: [Feature Request]: Harden workflows
### Feature Title

Harden workflows

### Feature Description

Harden workflows, so if we need to make pull requests downstream from `master`, it'll only be the `.github` directory which is pulled downstream, not the entire tee. All branches **should not** be behind `master`.

Image

Put into the workflows extra checks that source/target is an actually valid combination.

### Implementation Ideas (Optional)

_No response_

tobitege:
Replaces #3673.

The old PR targets `zzz-alpha-20260607`, which belongs to the archived zzz branch line. This replacement targets `alpha` and carries only the intended workflow hardening changes from the old PR.

Verified against `origin/alpha`: two replacement commits, 8 changed files, +25/-208.


PWagner1:
@tobitege 

Going to merge this one tomorrow, so that we get a nice round build number `160` on Tuesdays run :)

PWagner1 (approved):

Issue #3714, opened 2026-06-08
Improve accessibility for controls (V105-LTS, V110)


tobitege:
## Scope

Improve accessibility/UIA behavior for controls in the V110 cycle. This issue tracks concrete gaps observed while validating existing `AccessibilityTest` with CUA and while adding Ribbon accessibility coverage.

## Areas to improve

- Ribbon-related controls:
  - Expose stable UIA names/roles/actions for `KryptonRibbon` tabs, groups, group buttons, check boxes, QAT items, and `KryptonRibbonGroupCustomControl` hosts.
  - Ensure invoke/toggle/select behavior is available through standard accessibility actions.
  - Document the WinForms limitation around `AutomationId` for virtual `AccessibleObject` children; hosted real controls can still expose native IDs through `Control.Name`.

- `KryptonRichTextBox`:
  - CUA sees the outer element as `Edit "Rich Text Editor"` with `set_value`, but setting that outer element reports success without changing the actual value.
  - Setting the inner `Document "RichEdit Control"` does update the text correctly.
  - Investigate whether the outer provider should delegate `ValuePattern.SetValue` to the inner RichEdit provider or avoid advertising a misleading set action.

- Text input controls:
  - `KryptonTextBox` is visible as an editable element and can be changed through its inner edit element, but CUA showed the outer element without the configured `AccessibleName` (`User Name Input`).
  - `KryptonMaskedTextBox` similarly showed an empty name in the built-in `AccessibilityTest` result despite having `AccessibleName = "Date Input"`.
  - Check delegation from wrapper accessible objects to inner controls so `Name` is stable on the element automation tools actually target.

- `KryptonCheckedListBox`:
  - CUA exposes the wrapper as `List "Checklist"`, but items appear as generic `ListItem`s with `invoke/select`.
  - Checked items do not expose check box role/state clearly in the observed UIA tree.

- Selection/state reporting:
  - `KryptonListBox` items expose `invoke/select`, but after invoking `Item 2`, CUA did not show a selected state in the filtered tree.
  - Verify selection state propagation for list-style controls.

- `AccessibilityTest` expectations:
  - The built-in test reported `10/10 passed`, even when `KryptonTextBox` and `KryptonMaskedTextBox` had empty `Name` values.
  - Tighten the test so empty names fail for controls where `AccessibleName` was explicitly set.
  - Consider adding small checks for whether CUA-visible outer vs inner provider behavior matches what automation clients can actually use.

[...]

PR #3715, opened 2026-06-08
3714-v110-Improve accessibility of controls

tobitege:
Implements #3714 starting by implementing the Ribbon-related accessibility work first.

It gives `KryptonRibbon` a toolkit-owned accessibility tree so automation tools can discover and invoke ribbon tabs, groups, group buttons, check boxes, QAT items, and hosted custom controls by meaningful names instead of seeing only a generic ribbon container.

It also improves the wrapped-control accessibility providers that automation clients target in `AccessibilityTest`: text-like wrappers now forward names and values to the inner WinForms controls more consistently, list-like wrappers expose item children when the native inner provider does not surface them, and `KryptonListView` avoids duplicating item children already exposed by the native provider.

## Task List

- [x] Implement Ribbon-related accessibility for `KryptonRibbon`, tabs, groups, buttons, check boxes, QAT items, and hosted custom controls.
- [x] Review `KryptonRichTextBox` outer-to-inner provider behavior so automation clients do not get a misleading set-value surface.
- [x] Review text input controls such as `KryptonTextBox` and `KryptonMaskedTextBox` where configured accessible names are not consistently exposed on the targetable element.
- [x] Review `KryptonCheckedListBox` item roles and checked-state reporting.
  Added fallback item accessible objects when the wrapped WinForms provider does not expose real item children. These item objects report stable names, selection state, default toggle behavior, and checked/indeterminate states through the existing `AccessibleObject` surface. A deeper UIA TogglePattern provider is intentionally left out because it would require a larger provider rewrite.
- [x] Review list selection state reporting where selected items are not clearly reflected in the UIA tree.
  Added fallback item children and `GetSelected()` support for `KryptonListBox`, while leaving `KryptonListView` on the native item provider to avoid duplicate UIA rows. This keeps automation-visible list items available without replacing the inner WinForms provider wholesale.
- [x] Tighten `AccessibilityTest` so configured accessible names and states are verified instead of passing with empty exposed names.

### DIFF
--- a/Source/Krypton Components/Krypton.Ribbon/General/Accessibility/KryptonRibbonAccessibleObject.cs
+++ b/Source/Krypton Components/Krypton.Ribbon/General/Accessibility/KryptonRibbonAccessibleObject.cs
@@ -1,9 +1,6 @@
 #region BSD License
 /*
  *
- * Original BSD 3-Clause License (https://github.com/ComponentFactory/Krypton/blob/master/LICENSE)
- *  © Component Factory Pty Ltd, 2006 - 2016, All rights reserved.
- *
  *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
  *  Modifications by Peter Wagner (aka Wagnerp), Simon Coghlan (aka Smurf-IV), Giduac, Ahmed Abdelhameed, tobitege, KamaniAR, Lesandro Gotardo (aka lesandrog), Jorge A. Avilés (aka mcpbcs) et al. 2026 - 2026. All rights reserved.
  *
@@ -14,10 +11,6 @@ namespace Krypton.Ribbon;
 
 internal class KryptonRibbonAccessibleObject : Control.ControlAccessibleObject
 {
-    #region Static Fields
-    private static readonly char[] _whiteSpace = new[] { ' ', '\t', '\r', '\n' };
-    #endregion
-
     #region Instance Fields
     private readonly KryptonRibbon _ribbon;
     #endregion
@@ -142,14 +135,12 @@ internal class KryptonRibbonAccessibleObject : Control.ControlAccessibleObject
 
     private static string? Normalize(string? value)
     {
-        if (string.IsNullOrEmpty(value))
+        if (string.IsNullOrWhiteSpace(value))
         {
             return null;
         }
 
-        string normalized = value.Trim(_whiteSpace);
-
-        return normalized.Length == 0 ? null : normalized;
+        return value!.Trim();
     }
 
     private static string? CombineText(string? line1, string? line2)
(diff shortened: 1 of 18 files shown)


PWagner1 (approved):

Source/Krypton Components/Krypton.Ribbon/General/Accessibility/KryptonRibbonAccessibleObject.cs
```diff
@@ -0,0 +1,682 @@
+#region BSD License
+/*
+ *
+ * Original BSD 3-Clause License (https://github.com/ComponentFactory/Krypton/blob/master/LICENSE)
+ *  © Component Factory Pty Ltd, 2006 - 2016, All rights reserved.
```
Smurf-IV: This is a new class, therefore the original bsd statement is not needed

Source/Krypton Components/Krypton.Ribbon/General/Accessibility/KryptonRibbonAccessibleObject.cs
```diff
@@ -0,0 +1,682 @@
...
+    }
+
+    private static string? Normalize(string? value)
+    {
+        if (string.IsNullOrEmpty(value))
+        {
+            return null;
+        }
+
+        string normalized = value.Trim(_whiteSpace);
```
Smurf-IV: Fix warnings please

Source/Krypton Components/Krypton.Ribbon/General/Accessibility/KryptonRibbonAccessibleObject.cs
```diff
@@ -0,0 +1,682 @@
...
+        }
+
+        return null;
+    }
+
+    private static ViewBase? ItemView(KryptonRibbonGroupItem item)
+    {
+        switch (item)
+        {
+            case KryptonRibbonGroupButton button:
```
Smurf-IV: Please ask PWagner for a resharper License? then these will be converted into the expression switch form easily

Source/TestHarnesses/RibbonAccessibilityProbe/Program.cs
```diff
@@ -0,0 +1,219 @@
+#region BSD License
+/*
+ *
+ * Original BSD 3-Clause License (https://github.com/ComponentFactory/Krypton/blob/master/LICENSE)
```
Smurf-IV: New file, therefore no need for th eoriginal

Smurf-IV (changes requested):

tobitege:
@Smurf-IV Thanks for the feedback, fixed with latest commit.

{94651E98-CB79-47A0-832C-69A2322717BD}

PWagner1 (approved):

PR #3729, opened 2026-06-08
3714-V105-LTS-Improve accessibility of controls

### LINKED ISSUE #3714: Improve accessibility for controls (V105-LTS, V110)

tobitege:
Implements #3714 for V105-LTS.

It gives `KryptonRibbon` a toolkit-owned accessibility tree so automation tools can discover and invoke ribbon tabs, groups, group buttons, check boxes, QAT items, and hosted custom controls by meaningful names instead of seeing only a generic ribbon container.

It also improves the wrapped-control accessibility providers that automation clients target in `AccessibilityTest`: text-like wrappers now forward names and values to the inner WinForms controls more consistently, list-like wrappers expose item children when the native inner provider does not surface them, and `KryptonListView` avoids duplicating item children already exposed by the native provider.

## Task List

- [x] Implement Ribbon-related accessibility for `KryptonRibbon`, tabs, groups, buttons, check boxes, QAT items, and hosted custom controls.
- [x] Review `KryptonRichTextBox` outer-to-inner provider behavior so automation clients do not get a misleading set-value surface.
- [x] Review text input controls such as `KryptonTextBox` and `KryptonMaskedTextBox` where configured accessible names are not consistently exposed on the targetable element.
- [x] Review `KryptonCheckedListBox` item roles and checked-state reporting.
  Added fallback item accessible objects when the wrapped WinForms provider does not expose real item children. These item objects report stable names, selection state, default toggle behavior, and checked/indeterminate states through the existing `AccessibleObject` surface. A deeper UIA TogglePattern provider is intentionally left out because it would require a larger provider rewrite.
- [x] Review list selection state reporting where selected items are not clearly reflected in the UIA tree.
  Added fallback item children and `GetSelected()` support for `KryptonListBox`, while leaving `KryptonListView` on the native item provider to avoid duplicate UIA rows. This keeps automation-visible list items available without replacing the inner WinForms provider wholesale.
- [x] Tighten `AccessibilityTest` so configured accessible names and states are verified instead of passing with empty exposed names.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -31,11 +31,14 @@
 - Projects use `global using` like in GlobalDeclarations.cs, do not add new usings in other files
 - Before adding new variables check for existing ones
 - No variable aliasing
+- New files must use only the current Standard Toolkit BSD header. Do not add the original ComponentFactory BSD header unless the file is derived from original ComponentFactory source.
 
 ## C# Rules
 - Surgical edits: preserve structure, identifiers, and existing comments; avoid adding defensive checks unless asked
 - No unneeded `try/catch` blocks if there's no catch handling
 - Idioms: use null-propagation and object/collection initializers where consistent
+- Fix compiler, analyzer, and IDE warnings in new or modified code before handing work back
+- Prefer switch expressions for simple value/type dispatch that only returns a value; keep switch statements for complex control flow or side effects
 - Compatibility: ensure changes build for `net472` and C# 7.3
 - WinForms: `UseWindowsForms=true`; prefer designer-friendly patterns and keep partial classes tidy
 - WinForms designer: keep object declarations at file bottom; initialize in `*.Designer.cs` `InitializeComponent()`
--- a/Documents/Changelog/Changelog.md
+++ b/Documents/Changelog/Changelog.md
@@ -47,6 +47,7 @@
 
 ## 2026-07-20 - Build 2607 (Version 105-LTS - Patch 3) - July 2026
 
+* Implemented [#3714](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3714), Add accessibility support for custom drawn Krypton controls and editable control ButtonSpecs.
 * Resolved [#3690](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3690), Fix VS 2019 build targeting so VS 2019 uses only .NET Framework 4.x projects while VS2022/Current scripts refresh restore assets for modern target frameworks.
 * Resolved [#2862](https://github.com/Krypton-Suite/Standard-Toolkit/issues/2862), Fixed VisualForm resize flicker in .NET 10+
 * Implemented [#3550](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3550), Auto complete issues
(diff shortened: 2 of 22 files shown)


PWagner1 (approved):

AGENTS.md
```diff
@@ -31,11 +31,14 @@
 - Projects use `global using` like in GlobalDeclarations.cs, do not add new usings in other files
 - Before adding new variables check for existing ones
 - No variable aliasing
+- New files must use only the current Standard Toolkit BSD header. Do not add the original ComponentFactory BSD header unless the file is derived from original ComponentFactory source.
```
Smurf-IV, line 34: Thanks for these, 
Remember to add to the V110 on the next PR

Smurf-IV (approved):

PR #3797, opened 2026-06-27
* Expand `AGENTS.md` with comprehensive guidelines

PWagner1:
Replace the short "Recent Tooling Mistakes" notes with a full, structured repository guideline. Adds sections covering build/test/dev commands, shell and environment rules, project structure, architecture, editing philosophy, public API and C# rules, documentation/commenting guidance, changelog/TestForm demo processes, testing guidance, and commit/PR conventions. Also tightens guidance around PowerShell vs cmd.exe and when to run build scripts.

No changelog entry needed.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,31 +1,74 @@
 # Repository Guidelines
 
-## Recent Tooling Mistakes To Avoid
+## Always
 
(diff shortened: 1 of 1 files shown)


tobitege:
The operational directives at the start, that got removed, might still be worth it to avoid repeated errors that LLM's do still.
I had them added back then as PowerShell was the best supported compared to even git bash etc., so this may of course differ for anyone else.
The file grew a lot, which doesn't have to be a big issue, but maybe less is more - it is meant for the AI, not humans, to explain the crucial parts. :)
Other than that, go for it and we can compare notes and iterate on it.

tobitege (approved):

PR #3841, opened 2026-07-04
* An optional 'copy' button on messageboxes

PWagner1:
# Add an optional Copy button to message boxes (#3836)

Closes [#3836](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3836)

## Summary

Adds an optional **Copy** button to both `KryptonMessageBox` and `KryptonMessageBoxExtended`. When enabled, the button copies the message box contents to the clipboard (the same content produced by the existing `Ctrl+C` shortcut) **without closing the dialog**. The button is opt-in, defaults to hidden, and preserves the current appearance and behaviour when not requested.

While wiring the feature through the Extended message box, several pre-existing defects in that component were uncovered and fixed (dead `Ctrl+C` handler, a spurious caption countdown, collapsed line breaks, and button/text sizing problems).

### `KryptonMessageBox` (Krypton.Toolkit)
- New optional `bool? showCopyButton` parameter on all public `Show(...)` overloads, threaded through `ShowCore(...)` into the LTR (`VisualMessageBoxForm`) and RTL (`VisualMessageBoxRtlAwareForm`) forms.
- A dedicated, left-anchored `Copy` button (separate from the four action buttons) that returns `DialogResult.None` so it never dismisses the dialog.
- The `Ctrl+C` copy logic and the Copy button now share a single `CopyMessageBoxContentToClipboard()` method.

### `KryptonMessageBoxExtended` (Krypton.Toolkit.Utilities)
- Full parity: new optional `bool? showCopyButton` parameter on all public `Show(...)` overloads and the internal `ShowCore` / `ShowCoreWithBoolResult` / `ShowCoreWithCheckStateResult` methods.
- New `ShowCopyButton` property on `KryptonMessageBoxExtendedData` for the struct-based construction path.
- Copy works for all content types: **Normal** (`KryptonTextBox`), **RichTextBox** (`KryptonRichTextBox`) and **Hyperlink** (`KryptonLinkWrapLabel`).
- Mirrored across the LTR (`VisualMessageBoxExtendedForm`) and RTL (`VisualRTLMessageBoxExtendedForm`) forms.

### Localization
- New `MessageBoxStrings.CopyToClipboard` string (default `"C&opy to Clipboard"`), wired into `IsDefault` and `Reset()`, so the button caption is fully localizable.

## Bug fixes (found while implementing the above)

- **Dead `Ctrl+C` on the Extended message box:** `KeyPreview`/`KeyDown` were never wired up, so `Ctrl+C` did nothing. It is now wired in both LTR and RTL Extended forms, and the copy routine handles every content type (previously it only ever read the plain-text control).
- **Spurious `[60]` in the caption:** the Extended message box appended a timeout countdown to the title even when the timeout facility was not in use. It is now only appended when `useTimeOut` is `true`.
- **Collapsed line breaks:** a lone `\n` in the Normal (text) container was not rendered as a line break. Line endings are now normalised so `\n`, `\r`, and `\r\n` all render correctly.
- **Extended message box sizing:** action buttons could overlap and overflow the right edge, and the message text could be clipped.
  - Root cause: the Extended buttons were `AutoSize=true`, so under `AutoScaleMode=Font` they re-sized independently of the pre-computed layout, and the positioning math used a smaller slot width than the rendered button width.
  - Additionally, `UpdateMessageSizing` did not account for the icon margins or the content-container chrome, so the text control was handed less width than was measured, forcing extra wrapping and vertical clipping.
  - Fix mirrors the proven standard `KryptonMessageBox`: fixed-width buttons (`AutoSize=false`) with a minimum-width floor, and the content area is now sized using the full measured width/height (including icon margins and content chrome).

## Demos (TestForm)

- Updated the existing `MessageBoxTest` screen with a `showCopyButton` toggle.
- New `MessageBoxCopyButtonDemo` for `KryptonMessageBox`.
- New `MessageBoxExtendedCopyButtonDemo` for `KryptonMessageBoxExtended`, covering all content types plus preset scenarios (Error, RichText, Hyperlink, and the struct-based path). Both demos are registered on the start screen.

## Files changed

**Krypton.Toolkit**
- `Controls Toolkit/KryptonMessageBox.cs`
- `Controls Visuals/VisualMessageBoxForm.cs` + `.Designer.cs`
- `Controls Visuals/RTL Aware/VisualMessageBoxRtlAwareForm.cs` + `.Designer.cs`
- `Translations/General/MessageBoxStrings.cs`

**Krypton.Toolkit.Utilities**
- `Krypton MessageBox Extended/Controls Toolkit/KryptonMessageBoxExtended.cs`
- `Krypton MessageBox Extended/Controls Visuals/VisualMessageBoxExtendedForm.cs` + `.Designer.cs`
- `Krypton MessageBox Extended/Controls Visuals/VisualRTLMessageBoxExtendedForm.cs` + `.Designer.cs`
- `Krypton MessageBox Extended/General/KryptonMessageBoxExtendedData.cs`

**TestForm**
- `MessageBoxTest.cs` + `.Designer.cs`, `StartScreen.cs`
- `MessageBoxCopyButtonDemo.cs` + `.Designer.cs` (new)
- `MessageBoxExtendedCopyButtonDemo.cs` + `.Designer.cs` (new)

**Docs**
- `Documents/Changelog/Changelog.md`

## Compatibility / impact

- **Backwards compatible:** `showCopyButton` is an optional nullable parameter that defaults to hidden. Existing call sites are unaffected and the message boxes look/behave identically unless the button is explicitly requested.
- **Behavioural note:** Extended message box action buttons are now fixed-width (matching the standard `KryptonMessageBox`) rather than auto-sized. This corrects the previous overlap/overflow and makes the two message boxes visually consistent.
- **Target frameworks:** builds cleanly for `net472` and `net8.0-windows` (0 warnings / 0 errors). New code is `net472` / C# compatible with the rest of the toolkit.

## Test plan

- [x] `Krypton.Toolkit.Utilities` and `TestForm` build for `net8.0-windows` and `net472` with 0 warnings / 0 errors.
- [ ] `KryptonMessageBox`: Copy button appears only when `showCopyButton: true`; clicking copies contents and does not close the dialog; `Ctrl+C` still works; verify LTR and RTL.
- [ ] `KryptonMessageBoxExtended`: Copy button appears only when requested (via parameter and via `KryptonMessageBoxExtendedData.ShowCopyButton`); copy works for Normal, RichTextBox and Hyperlink content; `Ctrl+C` now works.
- [ ] Extended sizing: no button overlap/overflow, no horizontal scrollbar, and multi-line/wrapped text is not clipped.
- [ ] Caption no longer shows a `[60]` countdown when the timeout facility is not used; lone `\n` renders as a line break.
- [ ] Copy button caption is localizable via `MessageBoxStrings.CopyToClipboard`.

## Out of scope / follow-up

Some `KryptonMessageBoxExtended.ShowCore` variants (the `bool`- and `CheckState`-returning paths) construct the RTL form but never call `Show()`/`ShowDialog()`, so the RTL Extended box does not display through those paths. This is a pre-existing routing bug unrelated to this feature and is left for a separate fix; the `DialogResult` path displays the RTL box correctly.

image

image


PWagner1:
Merging, as it's preventing a bugfix

Issue #3861, opened 2026-07-05
[Feature Request]: Allow customizing glyph colors


lesandrog:
### Feature Title

Customizing glyph colors

### Feature Description

Allow customization of glyph colors using kryptonCustomPalette.

Issue #3663, opened 2026-06-03
[Feature Request]: Use glyphs for dropdown arrows


PWagner1:
### Feature Title

Use glyphs for dropdown arrows

### Feature Description

Following on from #2129, a sensible solution would be to use glyphs, which will be sharper.

PWagner1:
V105 to do still...

PR #3865, opened 2026-07-05
Customizing glyph colors (V110)

lesandrog:
# Allows you to customize glyph colors (#3663) using KryptonCustomPalette. (#3861 )

Glyph colors are now retrieved from theme palettes, rather than directly from the color scheme.

The colors remain the same, but you can now customize them using KryptonCustomPalette.

### TestForm 
GlyphColor

### Test Plan
- [ ] Add a KryptonCustomPalette.
- [ ] Associate the KryptonCustomPalette with a KryptonManager or with the components.
- [ ] Customize the colors:
- Enabled      
Outline:
`ButtonStyles.ButtonStandalone.StateNormal.Content.ShortText.Color1 `
Fill:
`ButtonStyles.ButtonInputControl.StateNormal.Content.ShortText.Color2`
- Disabled      
Outline:
`ButtonStyles.ButtonInputControl.StateDisabled.Content.ShortText.Color1`
Fill:
`ButtonStyles.ButtonInputControl.StateDisabled.Content.ShortText.Color2`


PWagner1 (approved):

Source/Krypton Components/Krypton.Toolkit/Rendering/DropDownArrowGlyphColors.cs
```diff
@@ -35,17 +35,17 @@ private static (Color Outline, Color Fill) FromPalette(PaletteBase palette, Pale
         Color fill;
         if (state == PaletteState.Disabled)
         {
-            outline = palette.GetSchemeColor(SchemeBaseColors.InputDropDownDisabled1);
-            fill = palette.GetSchemeColor(SchemeBaseColors.InputDropDownDisabled2);
+            outline = palette.GetContentShortTextColor1(PaletteContentStyle.ButtonInputControl, PaletteState.Disabled);
```
Smurf-IV, line 38: Why is this the text colour when it already was using the drop references, that are already customisable ?

Smurf-IV (commented):
?

Source/Krypton Components/Krypton.Toolkit/Rendering/DropDownArrowGlyphColors.cs
lesandrog: Are you referring to the change in how colors are obtained?
I only changed the method for obtaining the colors. The colors remain the same, but this way it’s possible to customize them using Kcustompalette.
As it was before, you could only customize the colors by editing the SchemeBase.cs files.

lesandrog: If there was another way to customize them, I’m not aware of it.

Smurf-IV: yes via whatever this enum points to: `SchemeBaseColors.InputDropDownDisabled1`

lesandrog: Forgive my ignorance, but as far as I know, the implementation of SchemeBaseColors only served to organize the color table. If there is another way for the user to customize colors besides kcustompalette, I don't know how to do it.

Smurf-IV: Use the repo that contains the Theme editor,
and have a "Play around" in there,
then when the you export the xml theme it should show what needs to be done, if being done by hand / programmatically.

lesandrog: But it is precisely for the sake of that XML import process that I suggested this change.
Here is what happens when importing the XML, before the change:

https://github.com/user-attachments/assets/5f6a0e6a-f645-499e-bafe-af392d221191

lesandrog: In the Palette Designer (I compiled it without the PR changes), the SchemeBase gets cached or something similar. When a color is changed, the cached SchemeBase is modified, and the colors are applied correctly to the arrows.

However, this doesn't happen in a normal project.

lesandrog: https://github.com/user-attachments/assets/3c97fc54-e1a7-489b-a003-6145d6a0f70f

Smurf-IV: > However, this doesn't happen in a normal project

I think that is because the pallete load does not perform a refresh when inside the designer (This also happens to some winform controls as well - like when changing font characteristics!)

What happens when you use the 1st xml at run time (And then the 2nd method)?

lesandrog: The result is the same, whether in design mode or at runtime. The colors in `SchemeBaseColors` are static. `GetSchemeColor` retrieves that static color, so any customization is ignored.

Even when defining a `KryptonCustomPalette`—using MS365 Blue as the base and customizing the colors—calling `palette.GetSchemeColor(SchemeBaseColors.TextButtonNormal);` goes straight to `PaletteBase` and always retrieves the static color: `public override Color TextLabelControl { get; set; } = Color.FromArgb(30, 57, 91);`

In contrast, the method `palette.GetContentShortTextColor1(PaletteContentStyle.ButtonStandalone, PaletteState.Normal);` first checks for customization within the `KryptonCustomPalette` to retrieve the color; if none exists, it proceeds to the base theme colors, and only falls back to `GetSchemeColor` as a last.

lesandrog: And my last argument is that, GetSchemeColor should not be used directly on the component, but only on the theme.
Note that, according to this search, it had never been used outside of themes before; this was the first time.

image

PR #3871, opened 2026-07-06
* Thread-safe logging

### LINKED ISSUE #3856: [Feature Request]: Replace `CommonHelper.LogOutput` with thread-safe, non-UAC-protected logging
`CommonHelper.LogOutput` limitations documented but unresolved

**Priority:** P2 (Medium)  
**Evidence:** `CommonHelper.cs` ~1530–1531:

```csharp
// TODO: Make this thread aware !
// TODO: DO NOT WRITE to the application path, as that might / will be UAC protected !!
```

Current implementation only writes to `Debug.WriteLine`.

**Recommendation:** Replace with optional `IKryptonLogger` or `DebugLogger` integration; never write to exe directory.

PWagner1:
# Replace `CommonHelper.LogOutput` with thread-safe, non-UAC-protected logging

Closes [#3856](https://github.com/Krypton-Suite/Standard-Toolkit/issues/3856)

## Summary

Introduces a unified diagnostic logging path for the toolkit via `IKryptonLogger` and `KryptonLogger`. `CommonHelper.LogOutput` and the internal `DebugLogger` now delegate to this shared infrastructure instead of ad-hoc `Debug.WriteLine` calls and the legacy (commented-out) file-to-exe-directory design.

- **Thread-safe** — file appends are serialized with a lock.
- **UAC-safe** — never writes beside the executable by default; optional file output uses `%LOCALAPPDATA%` or an explicit path from an environment variable.
- **Pluggable** — applications can register a custom `IKryptonLogger` (e.g. Serilog, ETW).
- **Binary compatible** — `CommonHelper.LogOutput(string)` signature is unchanged.

## Motivation

`CommonHelper.LogOutput` had unresolved TODOs to make logging thread-aware and to stop writing under the application directory (UAC-protected when installed under `Program Files`). A separate internal `DebugLogger` duplicated similar logic and defaulted to `AppContext.BaseDirectory\logs\` in DEBUG builds, carrying the same risk.

This change consolidates both paths behind one implementation with explicit opt-in file logging.

### New public API

| Type | Description |
|------|-------------|
| `IKryptonLogger` | Single `Write(string message)` contract for custom sinks | | `KryptonLogger` | Static gateway: `Current`, `SetLogger(IKryptonLogger?)`, `Write(string)` |

### Default logger behavior

1. Always writes to `System.Diagnostics.Debug`.
2. Optionally appends timestamped lines to a file when enabled via environment variable.
3. Swallows I/O failures so logging never throws from paint, dispose, or designer paths.

### Optional file logging

| Variable | Effect |
|----------|--------|
| `KRYPTON_LOG_PATH` | Full path to the log file | | `KRYPTON_LOG_WM` | Alias of `KRYPTON_LOG_PATH` (backward compatibility for WM/theme-swap tracing) | | `KRYPTON_LOG` | When `1`, `true`, `yes`, or `on` — writes to `%LOCALAPPDATA%\Krypton-Suite\Toolkit\Krypton.log` |

### Updated call paths

- `CommonHelper.LogOutput` → `KryptonLogger.Write`
- `DebugLogger.WriteLine` → `KryptonLogger.Write` with `[WM]` prefix (theme-swap tracing in `ThemeChangeCoordinator`, `KryptonComboBox`)

Existing internal call sites (`Clipping`, `KryptonCustomPaletteBase`, `KryptonSplitterPanelDesigner`) require no code changes.

## Behavior change (maintainers)

DEBUG builds **no longer** auto-create `logs\KryptonWM.log` under the application directory. To capture WM/theme-swap file traces, set `KRYPTON_LOG=1` or `KRYPTON_LOG_PATH` before running.

## Files changed

- `Source/Krypton Components/Krypton.Toolkit/General/Definitions.cs` — add `IKryptonLogger`
- `Source/Krypton Components/Krypton.Toolkit/General/KryptonLogger.cs` — new gateway and default implementation
- `Source/Krypton Components/Krypton.Toolkit/General/CommonHelper.cs` — delegate `LogOutput`, update XML docs
- `Source/Krypton Components/Krypton.Toolkit/General/DebugLogger.cs` — delegate to `KryptonLogger`
- `Documents/Changelog/Changelog.md` — release note for V110

## Test plan

- [ ] Build `Krypton.Toolkit` (Debug, `net472`) — no new warnings
- [ ] Run TestForm with no env vars — diagnostic messages appear in Visual Studio Output / DebugView only
- [ ] Set `KRYPTON_LOG=1` — confirm `%LOCALAPPDATA%\Krypton-Suite\Toolkit\Krypton.log` is created and appended
- [ ] Set `KRYPTON_LOG_PATH` to a temp file — confirm writes go to that path
- [ ] Run from a UAC-protected install location — confirm no write attempts next to the `.exe`
- [ ] Register a custom `IKryptonLogger` at startup — confirm `CommonHelper.LogOutput` traffic is received
- [ ] Trigger theme swap with `KRYPTON_LOG_WM` set — confirm `[WM]`-prefixed lines appear in the target file

## Custom logger example

```csharp
public sealed class MyLogger : IKryptonLogger
{
    public void Write(string message) => /* your sink */;
}

KryptonLogger.SetLogger(new MyLogger());
```


Source/Krypton Components/Krypton.Toolkit/General/DebugLogger.cs
```diff
@@ -2,64 +2,18 @@
...
-        {
-        }
-    }
-#else
-    public static void WriteLine(string message)
-    {
-    }
-#endif
+    public static void WriteLine(string message) =>
+        KryptonLogger.Write($"[WM] {message}");
```
Smurf-IV: This needs to be fast, so please use `string.Concat`

Smurf-IV: Also it's a writeline, so it should be assumed that it has an `environment.newline` as well ?

Source/Krypton Components/Krypton.Toolkit/General/KryptonLogger.cs
```diff
@@ -0,0 +1,164 @@
...
+
+    #region Public
+    /// <summary>
+    /// Gets the active logger instance.
+    /// </summary>
+    public static IKryptonLogger Current
+    {
+        get
+        {
+            var custom = _customLogger;
```
Smurf-IV, line 31: Why create a temp var ?

Source/Krypton Components/Krypton.Toolkit/General/KryptonLogger.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            return;
+        }
+
+        try
+        {
+            var line = $"[{DateTime.Now:HH:mm:ss.fff}] {message}";
+            lock (FileSync)
+            {
+                var directory = Path.GetDirectoryName(filePath);
```
Smurf-IV: Vert wasteful and slow, have this as a static on construction (Not startup!)

Source/Krypton Components/Krypton.Toolkit/General/KryptonLogger.cs
```diff
@@ -0,0 +1,164 @@
...
+    #endregion
+
+    #region IKryptonLogger
+    /// <inheritdoc />
+    public void Write(string message)
+    {
+        Debug.WriteLine(message);
+
+        var filePath = GetLogFilePath();
+        if (filePath == null)
```
Smurf-IV: Or empty / whitespace
Should be set as a field `bool` for speed

Source/Krypton Components/Krypton.Toolkit/General/KryptonLogger.cs
```diff
@@ -0,0 +1,164 @@
...
+            // Backward compatibility with theme-swap WM tracing.
+            explicitPath = Environment.GetEnvironmentVariable("KRYPTON_LOG_WM");
+        }
+
+        if (!string.IsNullOrWhiteSpace(explicitPath))
+        {
+            return explicitPath;
+        }
+
+        if (!IsFileLoggingEnabled())
```
Smurf-IV, line 141: Surely this should be higher up, so that all this other code can be jitted out!

Smurf-IV (changes requested):
There is a lot of slow code here, that will impact every call, regardless of logging being enabled !

Source/Krypton Components/Krypton.Toolkit/General/DebugLogger.cs
PWagner1: Uses `string.Concat` and improved documentation.

Source/Krypton Components/Krypton.Toolkit/General/KryptonLogger.cs
PWagner1: Improved documentation.

Source/Krypton Components/Krypton.Toolkit/General/KryptonLogger.cs
PWagner1: Now moved into construction

Source/Krypton Components/Krypton.Toolkit/General/KryptonLogger.cs
PWagner1: Now optimised.

Source/Krypton Components/Krypton.Toolkit/General/KryptonLogger.cs
PWagner1: Moved further up the chain.

Smurf-IV (approved):

PR #4065, opened 2026-07-24
* Fix docking indicator hit-target selection (V110)

PWagner1:
Clarified the logic for selecting docking indicator hit-targets during drag feedback. The feedback method must execute for every cluster (for its side effects), but only the first valid hit target should be retained. Removed confusing LINQ query and updated comments to reflect the intended behavior. Switched from unclear 'last match' semantics to explicit 'first match' using the null-coalescing operator.


PWagner1:
Merging, due to annoyance of changelog entry

PR #4072, opened 2026-07-25
* Use typed LINQ for Ribbon QAT enumeration (V110)

PWagner1:
Replace unsafe casts and Cast<T>() usages when enumerating QATButtons with typed LINQ query clauses and is-patterns. Adds a where filter for visible QAT buttons in accessibility and mini layout, and strengthens null-safety via Debug.Assert and pattern matching. Also updates Documents/Changelog/Changelog.md with a brief entry for #4060.


Source/Krypton Components/Krypton.Ribbon/Controls Ribbon/KryptonRibbon.cs
```diff
@@ -2768,12 +2768,12 @@ internal void DisplayQATCustomizeMenu(Rectangle screenRectangle,
         if (QATUserChange)
         {
             // Add an entry for each quick access toolbar button
-            foreach (var component in QATButtons)
+            foreach (IQuickAccessToolbarButton qatButton in from IQuickAccessToolbarButton qatButton in QATButtons
+                                                            select qatButton)
```
Smurf-IV: More like this:
`foreach (var qatButton in QATButtons.typeof<IQuickAccessToolbarButton>())`
It's not that, but looks the same !"

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Ribbon/Controls Ribbon/KryptonRibbon.cs
PWagner1: Updated to use `OfType<>()`

PR #4110, opened 2026-07-31
* Allow drill down in the custom file dialog navigation tree

### LINKED ISSUE #1231: [Bug / Feature Request]: The `KryptonOpenFileDialog` can have poor performance on some OS hardware
I have noticed that the KFOD has lot's of flickering issues on some installations of Windows 10.
So much so, that it has to be disabled.

Proposal:
1) Recreate the Open / save / Folder Dialogs with "Normal" krypton controls 
  - e.g. https://www.telerik.com/support/code-library/fileexplorer-demo-application
  - file-dialogs-radopenfiledialog003

2) Use `WindowsAPICodePack` and wrap the controls from that "https://github.com/dbarros/WindowsAPICodePack/blob/master/Samples/ExplorerBrowser/CS/WinForms/ExplorerBrowserTestForm.Designer.cs
3) Investigate the used of a themed WebBrowser
  - https://www.vbforums.com/showthread.php?597434-RESOLVED-Windows-File-Explorer-Control
4) Rework a WPF project:
  - https://www.codeproject.com/Articles/770837/WPF-x-FileExplorer-x-MVVM-2
  - FileExplorer3
5) Something else ?

PWagner1:
# Fix: Allow drill down in the custom file dialog navigation tree (#1231)

## Summary

The navigation tree in the custom (`ProviderMode = Custom`) Krypton file dialogs listed common places and drives as leaf nodes, so drives could not be expanded to reach their sub folders. Places and drives now expand, loading their sub folders on demand, and the tree keeps in step with navigation performed from the file list, the up button, the breadcrumbs and the address bar.

## Related issues

- Closes #1231

## Type of change

- [x] Bug fix (`Resolved`)
- [ ] **Breaking change** (API removal/rename, behavior change requiring migration, assembly/namespace move)
- [ ] Documentation only

## Changes

- `Krypton.Toolkit` — `VisualCustomFileDialogForm`:
  - Place and drive nodes get a placeholder child when the folder has sub folders, so the expand glyph is shown without walking the whole tree up front.
  - `BeforeExpand` replaces the placeholder with the real sub folders (`PopulateChildNodes`), and each new child gets its own placeholder, giving unlimited depth at a constant load cost.
  - Unreadable or disconnected locations are skipped instead of raising an error.
  - `SelectNavigationNode` resolves a path through `FindNodeForPath`, which walks down from the closest node already in the tree and lazily materialises the levels in between, so deep folders can be selected even if their branch has never been expanded.
  - Navigating from the file list (double-click / open of a folder) and the up button now selects the matching tree node.
  - First `Shown` navigate no longer early-returns when the constructor has already set `_currentPath`, which left the breadcrumb on the default `KryptonBreadCrumb` caption "Root" and never loaded the file list.
  - Removed the Common Places "Root Folder" node. `RootFolder` is still used only as a starting/fallback path in `ResolveStartingPath` (FolderBrowserDialog semantics), not as a duplicate place entry.
  - File name box starts empty when `FileName` is not set, with optional localizable cue text via `KryptonManager.Strings.CustomFileDialogStrings` (`FileNameCueText`, `FolderPathCueText`, `AddressPathCueText`, `SearchCueText`). Clear a cue string to disable it.
  - The editable address path and search boxes use a palette-aware `VisualPopup`/`KryptonListBox` autocomplete instead of the native WinForms dropdown. Address matching runs after a short delay off the UI thread and returns folders; search combines current-folder names with a bounded process-session history.
  - Autocomplete supports Up/Down selection, Enter acceptance, Escape dismissal and double-click acceptance.
  - Right-clicking the breadcrumb opens a Krypton context menu with localizable Copy Address, Copy Address as Text, Edit Address and Delete History commands. Copy Address uses a quoted path, while Copy Address as Text uses plain Unicode text; Delete History retains the current path and clears back/forward entries.
  - Added a Date modified combo beside the search box with Explorer-style presets (Any time through Last year). Text search and date filter combine; suggestion names also respect the date filter. Opt in with `ShowDateModifiedFilter = true` (default `false`). Strings live under `KryptonManager.Strings.CustomFileDialogStrings`.
  - Remaining custom-dialog UI chrome (buttons, labels, places, columns, view modes, status and validation prompts) is localizable through the same strings object; Cancel uses `KryptonManager.Strings.GeneralStrings.Cancel`.

- Packages: `Krypton.Toolkit`.
- TFMs verified: `net472`.

## Validation

- TestForm demo: `KryptonFileDialogProviderDemo` (registered in `StartScreen.AddButtons()`).
- Manual steps:
  1. Run TestForm and open the file dialog provider demo with `ProviderMode = Custom`.
  2. Expand `Drives` → `C:\` and confirm the drive expands into its folders, and that those folders expand further.
  3. Double-click a folder in the file list and confirm the tree expands and selects the matching node.
  4. Use the up button and confirm the tree selection moves to the parent folder.
  5. Press Ctrl+L, type a partial folder path and confirm a Krypton-themed folder suggestion popup appears and accepts keyboard or double-click selection.
  6. Type in Search and confirm matching current-folder names appear. Press Enter, change the search, then type part of the earlier term and confirm it is suggested from session history.
  7. Right-click the breadcrumb and exercise each command; confirm Delete History disables back/forward navigation and then becomes disabled.
  8. Change Date modified to Today / This week and confirm the file list and search suggestions only include matching items; combine with typed search text.
- Build: `dotnet build ".\Source\Krypton Components\Krypton.Toolkit\Krypton.Toolkit 2022.csproj" -c Debug -f net472`

## Changelog

```markdown
* Implemented [#1231](https://github.com/Krypton-Suite/Standard-Toolkit/issues/1231), The `KryptonOpenFileDialog` can have poor performance on some OS hardware
   * The custom dialog navigation tree can now be expanded into drives, places and their sub folders (folders are loaded on demand), and it follows navigation made from the file list, the up button and the address bar.
```

None.


PWagner1:
A few fixes, merging

PR #4134, opened 2026-08-01
* Stop the window region clipping the right and bottom form borders

### LINKED ISSUE #4132: [Bug]: ControlBox item borders do not paint correctly
### Summary

Border not paint at Right and Bottom

### Description

look at "Button Test" form border (Right and bottom)
Close button of ControlBox padding/margin removed 

Image

### Steps to Reproduce

Run test project,
open any form and check all borders

### Expected Behavior

Paint all borders

### Actual Behavior

Right and bottom border lost

### Operating System

_No response_

### OS Version

_No response_

### Framework/.NET Version

_No response_

### Toolkit Version

_No response_

### Additional Information

_No response_

### Areas Affected

None

PWagner1:
# Fix: Stop the window region clipping the right and bottom form borders (#4132)

## Summary

`KryptonForm` shapes itself by converting the renderer's "outside border path" into a window `Region`. That path was being built one pixel short on the right and bottom, so Windows clipped away exactly the column and row that the border is painted into. The result was a form with only its left and top borders visible, and a Close button that looked like its padding had been removed because it sits flush against that clipped column.

A related Close-button collision (right-hand control-box inset of `0` overlapping the form border column) is left at its historical default, but those palettes now read a single tunable (`GlobalStaticConstants.HEADER_BUTTON_EDGE_INSET_FORM_RIGHT`) so the gap can be adjusted later without hunting through every theme file.

## Related issues

- Closes #4132

## Type of change

- [x] Bug fix (`Resolved`)
- [ ] **Breaking change**
- [ ] Documentation only

## Changes

- **Krypton.Toolkit / `RenderStandard.CreateBorderBackPath`**: the `#1871` micro-radius workaround (forcing `PaletteCornerRounding.Uniform(0.1f)` when a palette reports square corners) is now applied only when building the pen-centred path (`middle`). Outside paths fall through to the square branch, which encloses the whole rectangle.
- **Krypton.Toolkit / built-in palettes**: 27 palettes that hardcoded `0` for `PaletteMetricInt.HeaderButtonEdgeInsetFormRight` now return `GlobalStaticConstants.HEADER_BUTTON_EDGE_INSET_FORM_RIGHT` (still defaults to `0`). Office 2007/2010 base and Professional themes keep `CommonHelper.GetFormHeaderButtonEdgeInsetRight`. macOS / OS X Aqua keep their deliberate larger insets; Material keeps its own `0`.

### Why the region path was clipping the border

`CreateBorderBackPath` serves three different consumers:

| Caller | `forBorder` / `middle` | Purpose |
| --- | --- | --- |
| `GetBorderPath` | `false` / `true` | Pen-centred path the border is stroked along | | `GetBackPath` | varies | Path the background is filled inside | | `GetOutsideBorderPath` | `true` / `false` | Path used as the window `Region` |

The `#1871` workaround forces a 0.1 radius so square borders take the rounded code path, whose arc boxes are inset by one pixel (`radius + 1`) to keep a one-pixel pen inside the rectangle. That inset is correct for stroking, but wrong for a region: a region derived from a `300 x 200` rectangle only covered `299 x 199`, so the border drawn at `Right - 1` / `Bottom - 1` fell outside it.

Verified with an offscreen probe against the built assembly:

```
before: outsideBorderPath bounds = {X=0,Y=0,Width=298.9999,Height=198.99997}
after:  outsideBorderPath bounds = {X=0,Y=0,Width=300,Height=200}
        borderPath        bounds = {X=0,Y=0,Width=298.9999,Height=198.99997}   (unchanged)
```

The stroked border geometry and the painted pixels are byte-identical before and after, so the `#1871` fix is preserved; only the region grew back to the full rectangle. Palettes that report a genuine rounding (>= 0.1) never entered the workaround and are unaffected.

### Close button right-border collision (tunable, default unchanged)

With a right-hand inset of `0`, the Close button extends into the column the form border paints, and `ViewDrawForm` paints that border in `RenderAfter`, after its children. Raising `GlobalStaticConstants.HEADER_BUTTON_EDGE_INSET_FORM_RIGHT` toward `CommonHelper.GetFormHeaderButtonEdgeInsetRight` (typically `Math.Max(2, RealWindowBorders.Right)`) clears the collision. Left at `0` for now so the visible caption layout stays historical while the region fix ships.

- Packages: `Krypton.Toolkit`.
- TFMs verified: `net8.0-windows` (Debug). The change remains `net472` / C# 7.3 compatible.

## Validation

- TestForm demo: `Bug4132FormBorderClippingDemo` (registered in `StartScreen.AddButtons()`).
- Manual steps:
  1. Open the demo. All four borders of the demo window must be visible; before the fix the right and bottom edges were missing.
  2. Hover the Close button in the caption. With the default inset of `0`, the right edge can still collide with the form border; raise `HEADER_BUTTON_EDGE_INSET_FORM_RIGHT` when tuning that gap.
  3. Open the Sizable / FixedSingle / FixedDialog / ToolWindow samples and check each one the same way.
  4. Switch themes with the combo. Square-cornered themes (Microsoft 365, Office 2013) showed the region bug; rounded themes were already correct and must stay correct.
  5. Resize and maximise/restore each window. The live readouts probe the window `Region` and must report `PASS` for left, top, right and bottom.
- Build: `dotnet build ".\Source\Krypton Components\TestForm\TestForm.csproj" -c Debug -f net8.0-windows` (0 warnings, 0 errors).

## Changelog

```markdown
* Resolved [#4132](https://github.com/Krypton-Suite/Standard-Toolkit/issues/4132), ControlBox item borders do not paint correctly
   * The right and bottom borders of a `KryptonForm` are no longer clipped away by the window region.
   * Themes that previously hardcoded a zero right-hand control-box inset now read `GlobalStaticConstants.HEADER_BUTTON_EDGE_INSET_FORM_RIGHT` (still defaults to `0`). Raise that value toward the near-edge inset when tuning the Close button gap against the form border.
```

None. `GetOutsideBorderPath` and the `KryptonForm` back path now cover one extra pixel column and row for square-cornered palettes, which is the rectangle the caller already asked for. Borders are still drawn after the background, so the extra background pixel is painted over by the border. The control-box inset default remains `0`.

## Checklist

- [x] Builds for `net472` and is C# 7.3 compatible where required
- [x] New compiler/analyzer warnings in touched code addressed
- [x] `Documents/Changelog/Changelog.md` updated
- [x] TestForm demo added or updated (features / observable bug fixes)
- [ ] `Documents/Development/` guide added (substantial features)
- [ ] Screenshots/GIFs included (UI changes)
- [x] Breaking-change impact and TFM notes documented above


KamaniAR (commented):
Border is now proper.
but Top and Right Padding/Margin is not same (Check if related to this else ignore)

Image

PWagner1:
Hi @KamaniAR 

Fixes have been made, would you mind giving it another go?

KamaniAR (approved):
main issue is resolved.

Currently Clipped from top,
Title bar icon and text look ok, but buttonspec are more clipped.

minor issue from base winforms (Top:-8, Left: -8) (May be resolve if anyone create issue on Microsoft repo)

when i drag mouse on top right corner close button is not focused.

look at this screenshot of maximized Form

Image

PR #4187, opened 2026-08-12
* Async Form and Dialog Methods (V105 LTS)

### LINKED ISSUE #4177: [Feature Request]: When will the Async Form Methods from .net10 (Started in .net9) actually be implemented
### Feature Title

Async Form Methods from .net10 (Started in .net9)

### Feature Description

.Net9 started to create Async form methods.
These have been ratified and expanded in .net 10
When will Krypton develop / implement these?

### Implementation Ideas (Optional)

_No response_

PWagner1:
This change adds .NET 9+ async dialog support for KryptonMessageBox.ShowAsync and KryptonTaskDialog.ShowDialogAsync, matching the existing KryptonForm async APIs. It also adds the net9+ warning suppression needed for the WinForms async API analyzers and wires a new Async Form Methods smoke scenario into TestForm and the start screen.

### DIFF
--- a/Source/Krypton Components/TestForm/AsyncFormMethodsTest.cs
+++ b/Source/Krypton Components/TestForm/AsyncFormMethodsTest.cs
@@ -1,0 +1,87 @@
+#region BSD License
+/*
+ *
+ *  New BSD 3-Clause License (https://github.com/Krypton-Suite/Standard-Toolkit/blob/master/LICENSE)
+ *  Modifications by Peter Wagner(aka Wagnerp) & Simon Coghlan(aka Smurf-IV), et al. 2026 - 2026. All rights reserved.
+ *
+ */
+#endregion
+
+namespace TestForm;
+
+public partial class AsyncFormMethodsTest : KryptonForm
+{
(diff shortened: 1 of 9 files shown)


Source/Krypton Components/TestForm/AsyncFormMethodsTest.cs
```diff
@@ -0,0 +1,87 @@
...
+#endregion
+
+namespace TestForm;
+
+public partial class AsyncFormMethodsTest : KryptonForm
+{
+    public AsyncFormMethodsTest()
+    {
+        InitializeComponent();
+#if !NET9_0_OR_GREATER
```
Smurf-IV: No, These need to be the same regardless of TFM, so that one code base works for all targets.

Smurf-IV (changes requested):

Source/Krypton Components/TestForm/AsyncFormMethodsTest.cs
PWagner1: Cleaned up.

Smurf-IV (approved):

PR #4259, opened 2026-08-22
* Eliminate black-frame artifacts on KryptonForm window-state transit…

### LINKED ISSUE #4255: [Bug]: VisualForm: eliminate black-frame artifacts and cut non-client paint/layout cost on window-state transitions Summary
### Summary

Black boxes when max / minimizing from taskbar 

### Description

This proposal targets Krypton.Toolkit.VisualForm (the base class that applies custom chrome to a Form, and the parent of KryptonForm). On custom-chrome forms it fixes three user-visible / performance problems that appear during maximize, minimize and restore, without changing behaviour during interactive border-drag resizing:

Black rectangles flashing in the frame/client area when the window is maximized, minimized or restored.
A re-entrant WM_NCPAINT storm in which the chrome is repainted several times for a single logical change (measured at ~32% of frame time in a real workload).
An uncoalesced layout cascade on atomic (non-interactive) state transitions, which for deeply nested content ran the full layout pass 2–3× per transition and needlessly resized child controls to a 0-px client on minimize.

All changes are confined to VisualForm. They were validated in a production Krypton-based application on .NET 10 (Windows Forms, heavily nested docking with a Direct2D-hosted child): the black artifacts are gone, the re-entrancy collapses to a single paint, and interactive drag is visually unchanged.

How it was diagnosed

Captured with JetBrains dotTrace (Timeline) on the real application, filtered by the UI Freeze and Waiting for CPU interval filters.

The freeze/paint time concentrated in OnWM_NCPAINT → OnNonClientPaint → KryptonForm.WindowChromePaint, and in a recurring WndProc → OnPaintNonClient → DefWndProc → WndProc cycle.
The GC filter showed negligible time (sub-second over a ~1-minute session), ruling GC out as a factor.
The black artifacts reproduced with the original synchronous OnNonClientPaint, which pointed away from the invalidation strategy and toward two concrete causes: an uninitialized back-buffer bitmap and the OS background erase covering exposed areas before the chrome repaints.
Root causes and fixes
1. Black frame from an uninitialized back-buffer — OnNonClientPaint

CreateCompatibleBitmap returns memory that is not zero-initialized; in practice it reads as black. When the cached NC back-buffer is (re)created after a size change, any pixel that WindowChromePaint does not cover (corners, transparent regions, transient sizes during a transition) is blitted to the window as pure black.

Fix: clear the back-buffer to the form background before painting the chrome. The destination DC already has ExcludeClipRect applied over the client area, so this never touches a hosted client control (e.g. a Direct2D surface).

```cs
using (Graphics g = Graphics.FromHdc(_screenDC))
{
    // Defensive clear: a freshly (re)created compatible bitmap is uninitialized
    // (reads as black). Clearing to BackColor guarantees no black pixels survive
    // if WindowChromePaint leaves any gap or the size is mid-transition. The
    // client area is untouched because the destination DC excludes it.
    g.Clear(BackColor);
    WindowChromePaint(g, windowBounds);
}
```

The same one-line clear is applied to the non-back-buffer fallback branch.

2. Black frame from the OS erase — WndProc (WM_ERASEBKGND)

During a transition Windows erases newly-exposed regions with the default background brush a fraction of a second before WM_NCPAINT/WM_PAINT repaints them. On a themed custom-chrome form that transient fill is visible as a black flash and, when remoted, is captured and transmitted as a separate frame.

Fix: claim the erase as handled while custom chrome is active, so Windows does not paint the transient fill.

```cs
case WM_ERASEBKGND:
    if (_themedApp && !DesignMode)
    {
        m.Result = (IntPtr)1;   // "background already erased"
        processed = true;
    }
    break;
```

Maintainer note: consumers whose client area is not fully covered by a child control benefit from DoubleBuffered/OptimizedDoubleBuffer. If preferred, this suppression can be placed behind an opt-in property rather than gated only on _themedApp.

3. Re-entrant WM_NCPAINT storm — OnPaintNonClient

OnPaintNonClient calls DefWndProc (to let default processing update the caption/text) and then InvalidateNonClient(). For messages such as WM_SETTEXT and non-interactive WM_SYSCOMMAND, DefWndProc synchronously re-enters WndProc → OnWM_NCPAINT → a full chrome repaint, and the subsequent InvalidateNonClient() schedules yet another. The result is several chrome repaints per logical change.

Fix: bracket the DefWndProc call with the existing SuspendPaint()/ResumePaint() so nested WM_NCPAINT is swallowed by the existing _ignoreCount > 0 guard, then request exactly one repaint. Commands that enter a modal move/size loop (SC_MOVE, SC_SIZE) are excluded, because DefWndProc does not return until the loop ends and the chrome must keep repainting during it.

```cs
protected virtual bool OnPaintNonClient(ref Message m)
{
    var suppressNestedPaint = true;
    if (m.Msg == PI.WM_.SYSCOMMAND)
    {
        long sc = m.WParam.ToInt64() & 0xFFF0;
        if (sc == SC_SIZE || sc == SC_MOVE)
        {
            suppressNestedPaint = false;   // modal loop: keep live repaint
        }
    }

    if (suppressNestedPaint)
    {
        SuspendPaint();
        try { DefWndProc(ref m); }
        finally { ResumePaint(); }
    }
    else
    {
        DefWndProc(ref m);
    }

    InvalidateNonClient();   // single coalesced repaint
    return true;
}
```

4. Synchronous chrome repaint at discrete transitions — RedrawNonClientNow (new)

InvalidateNonClient() deliberately defers (it omits RDW_UPDATENOW) so continuous drag-resizing stays cheap. But for a discrete transition (maximize / restore / un-minimize) that deferral can let DWM present the unpainted frame for one pump cycle. A new method forces a synchronous full-frame NC repaint for those cases only.

Two details:

It mirrors the exact region geometry used by InvalidateNonClient so both paths invalidate the same pixels.
It temporarily resets _ignoreCount around its own RedrawWindow so the synchronous WM_NCPAINT it triggers is not swallowed if the method happens to be invoked from inside a SuspendPaint bracket. Ambient WM_NCPAINTs emitted by DefWndProc outside this call remain gated, so this does not reopen the re-entrancy storm.

```cs
protected void RedrawNonClientNow()
{
    if (IsDisposed || Disposing || !IsHandleCreated) return;
    if (CommonHelper.IsFormMinimized(this)) return;   // nothing visible

    lock (lockObject)
    {
        Padding realWindowBorders = RealWindowBorders;
        Rectangle realWindowRectangle = RealWindowRectangle;
        Rectangle invalidRect = realWindowRectangle with
        {
            X = -realWindowBorders.Left,
            Y = -realWindowBorders.Top
        };

        IntPtr hRgn = PI.CreateRectRgn(invalidRect.Left, invalidRect.Top,
                                       invalidRect.Right, invalidRect.Bottom);
        if (hRgn == IntPtr.Zero) return;

        try
        {
            Rectangle clientRect = ClientRectangle;
            IntPtr hClientRgn = PI.CreateRectRgn(clientRect.Left, clientRect.Top,
                                                 clientRect.Right, clientRect.Bottom);
            if (hClientRgn != IntPtr.Zero)
            {
                try { PI.CombineRgn(hRgn, hRgn, hClientRgn, 4); /* RGN_DIFF */ }
                finally { PI.DeleteObject(hClientRgn); }
            }

            // Ungate ONLY our own deliberate synchronous paint.
            int savedIgnoreCount = _ignoreCount;
            _ignoreCount = 0;
            try
            {
                PI.RedrawWindow(Handle, IntPtr.Zero, hRgn,
                                PI.RDW_FRAME | PI.RDW_INVALIDATE | RDW_UPDATENOW);
            }
            finally { _ignoreCount = savedIgnoreCount; }
        }
        catch (InvalidOperationException ioEx) { Debug.WriteLine(ioEx.Message); }
        finally { PI.DeleteObject(hRgn); }
    }
}
```

It is invoked from the WM_SIZE tail, only on a genuine state transition (not on every drag step, which also reports SIZE_RESTORED):

```cs
if (m.Msg == PI.WM_.SIZE)
{
    _blurManager.TakeSnapshot();

    int sizeState = (int)(m.WParam.ToInt64() & 0xFFFF);
    if (sizeState != _lastWmSizeState)
    {
        int previousState = _lastWmSizeState;
        _lastWmSizeState = sizeState;

        if (sizeState == SIZE_MAXIMIZED
            || (sizeState == SIZE_RESTORED
                && (previousState == SIZE_MAXIMIZED || previousState == SIZE_MINIMIZED)))
        {
            RedrawNonClientNow();
        }
    }
}
```

5. Coalesce the layout cascade on atomic transitions — WndProc (WM_SYSCOMMAND)

Minimize / maximize / restore from the taskbar or the caption buttons arrive as WM_SYSCOMMAND and drive the whole transition synchronously inside DefWndProc, emitting several WM_WINDOWPOSCHANGED / WM_SIZE messages. Uncoalesced, the layout pass runs 2–3× per transition; on minimize the client collapses to ~0 px, so every docked child (including Direct2D swap chains) is resized to nothing and then back.

Because these commands are atomic and non-interactive (the user never sees intermediate frames), it is safe to suspend layout across the transition and lay out once at the end — the opposite of border-drag, where suspension would freeze content the user is actively watching.

```cs
var scCommand = m.WParam.ToInt64() & 0xFFF0;

if (scCommand == SC_MINIMIZE || scCommand == SC_MAXIMIZE || scCommand == SC_RESTORE)
{
    SuspendLayout();
    try { base.WndProc(ref m); }   // performs the state change; layout deferred
    finally
    {
        if (CommonHelper.IsFormMinimized(this))
            ResumeLayout(false);   // minimized: skip the wasted 0-px layout
        else
            ResumeLayout(true);    // one cascade at the final size
    }
    processed = true;
    break;
}
```

SC_CLOSE handling (setting CloseReason) is preserved, and all other system commands still fall through to the existing OnPaintNonClient path.

Maintainer note: this is the change most worth reviewing for edge cases (MDI children, programmatic WindowState assignment, which does not raise WM_SYSCOMMAND). It could reasonably ship behind an opt-in property. The chrome still repaints correctly during the transition via the WM_SIZE tail above; SuspendLayout defers layout only, not painting.

Supporting fields / constants

```cs
private int  _lastWmSizeState = -1;

private const int  WM_ERASEBKGND  = 0x0014;
private const int  SIZE_RESTORED  = 0;
private const int  SIZE_MINIMIZED = 1;
private const int  SIZE_MAXIMIZED = 2;
private const long SC_SIZE        = 0xF000;
private const long SC_MOVE        = 0xF010;
private const long SC_MINIMIZE    = 0xF020;
private const long SC_MAXIMIZE    = 0xF030;
private const long SC_RESTORE     = 0xF120;
private const int  RDW_UPDATENOW  = 0x0100;
```

(Where equivalents already exist under PI.* those can be reused instead; locals are shown for clarity.)

Why it is safe
Interactive drag is untouched. No WM_ENTERSIZEMOVE/WM_EXITSIZEMOVE bracketing is introduced, and the modal SC_SIZE/SC_MOVE commands are explicitly excluded from paint suppression. The chrome repaints in real time throughout a drag.
Re-entrancy is collapsed, not disabled. Suppression is scoped to a single DefWndProc call that returns in milliseconds; the RedrawNonClientNow ungate is scoped to a single RedrawWindow. Ambient WM_NCPAINT gating via _ignoreCount is otherwise unchanged.
Transition detection uses information already in the message (WM_SIZE wParam), so ordinary drag (SIZE_RESTORED repeated) triggers nothing new.
Client content is never overpainted — the defensive Clear writes into a DC that excludes the client area.
Behaviour is gated on _themedApp where relevant, matching existing conventions in the class.
How to validate

Re-capture a session of repeated minimize / restore / maximize with the UI Freeze filter and compare against baseline:

OnWM_NCPAINT / OnNonClientPaint / WindowChromePaint total drops sharply (redundant passes removed).
The WndProc → OnPaintNonClient → DefWndProc → WndProc re-entrant subtree collapses to a single paint.
No black frame is presented on maximize / minimize / restore.
On minimize, child controls are no longer resized to a 0-px client (visible as the absence of a 0→full swap-chain resize for hosted Direct2D/Direct3D children).
Scope / out of scope

This proposal is intentionally limited to VisualForm. In the originating application, further gains came from application-side changes that are not part of this proposal and should not be assumed by Krypton:

Coalescing a nested docking/workspace subtree and guarding a hosted Direct2D control against 0-px (OnSizeChanged) transitions — application code.
A deployment factor unrelated to Krypton: on Remote Desktop Session Host, DirectX 3D content renders on the Microsoft Basic Render Driver (WARP, software) unless the "Use hardware graphics adapters for all RDS sessions" policy is enabled — a configuration item, not a code change.

### Steps to Reproduce

Using a Profiler minimize and maximize a KryptonForm with several Docking Manage Objects inside a another Docking Manage Object

### Expected Behavior

Black box desappear

### Actual Behavior

Blackboxes are present

### Operating System

_No response_

### OS Version

_No response_

### Framework/.NET Version

_No response_

### Toolkit Version

_No response_

### Additional Information

_No response_

### Areas Affected

None

PWagner1:
…ions (V110)

# Fix: Eliminate black-frame artifacts on KryptonForm window-state transitions (#4255)

## Summary

`KryptonForm` (via `VisualForm` custom chrome) no longer flashes black rectangles when maximized, minimized, or restored from the caption or taskbar. The same transitions also drop redundant non-client paints and skip a wasted 0-pixel layout pass on minimize, without changing live chrome painting during border-drag resize.

## Related issues

- Closes #4255

## Type of change

- [x] Bug fix (`Resolved`)
- [ ] **Breaking change** (API removal/rename, behavior change requiring migration, assembly/namespace move)
- [ ] Documentation only

## Changes

- `VisualForm.OnNonClientPaint`: clear the compatible bitmap (and fallback DC) to `BackColor` before chrome paint so uninitialized GDI bits are not blit as black.
- `VisualForm.WndProc`: claim `WM_ERASEBKGND` while themed custom chrome is active, so Windows does not flash a default erase before `WM_NCPAINT`.
- `VisualForm.OnPaintNonClient`: bracket `DefWndProc` with `SuspendPaint`/`ResumePaint` except for modal `SC_SIZE`/`SC_MOVE`, collapsing re-entrant `WM_NCPAINT` to one coalesced invalidate.
- `VisualForm.RedrawNonClientNow`: synchronous non-client paint on genuine maximize/restore `WM_SIZE` transitions only (not repeated `SIZE_RESTORED` during drag).
- `VisualForm.WndProc` (`WM_SYSCOMMAND`): suspend layout across caption/taskbar min/max/restore and resume once at the final size (`ResumeLayout(false)` when minimized).

- Packages: `Krypton.Toolkit` (`VisualForm`, base of `KryptonForm`).
- TFMs verified: Debug build of `Krypton.Toolkit` plus `TestForm`.

## Validation

- TestForm demo: `Bug4255VisualFormStateTransitionDemo` (registered in `StartScreen.AddButtons()`).
- Manual steps:
  1. Open **Bug 4255 Form max/min black frame**.
  2. Maximize, restore, and minimize from the caption buttons and from the taskbar — no black frame.
  3. Drag a border to resize — chrome continues to paint during the drag.
  4. Confirm the inner size-probe log does not record 0×0 on caption/taskbar minimize.
- Build: `dotnet build ".\Source\Krypton Components\Krypton.Toolkit\Krypton.Toolkit 2022.csproj" -c Debug`

## Changelog

```markdown
* Resolved [#4255](https://github.com/Krypton-Suite/Standard-Toolkit/issues/4255), Black frame flash and extra layout/paint cost when maximizing, minimizing, or restoring a `KryptonForm` from the caption or taskbar.
```

None. Custom-chrome erase is claimed as handled for themed forms; client content that does not cover the client area should continue to use `DoubleBuffered` / `OptimizedDoubleBuffer` as before.

## Checklist

- [x] Builds for `net472` and is C# 7.3 compatible where required
- [x] New compiler/analyzer warnings in touched code addressed
- [x] `Documents/Changelog/Changelog.md` updated
- [x] TestForm demo added or updated (features / observable bug fixes)
- [x] Breaking-change impact and TFM notes documented above


PWagner1:
Merging, same as V105 fix.

PR #4331, opened 2026-09-02
* Fresh KTree does not allow MultiSelect to be set to false (V105 LTS)

### LINKED ISSUE #4326: [Bug]: Fresh KTree does not allow `MultiSelect` to be set to false
.net 10 KForm with fresh dropped KTree
Properties do not allow `MultiSelect` to be set to false
V100 and v110 Canary

PWagner1:
* See PR #4330

### DIFF
--- a/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonTreeView.cs
+++ b/Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonTreeView.cs
@@ -1018,9 +1018,9 @@ public class KryptonTreeView : VisualControlBase,
             foreach (TreeNode node in value)
             {
                 node.Checked = true;
-                if (!MultiSelect && !CheckBoxes)
+                if (!MultiSelect)
                 {
-                    // Only keep the first node when neither click-to-toggle multi-select nor check boxes are enabled.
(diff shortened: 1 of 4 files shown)


Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonTreeView.cs
```diff
@@ -1011,9 +1018,9 @@ public List<TreeNode> CheckedNodes
             foreach (TreeNode node in value)
             {
                 node.Checked = true;
-                if (!MultiSelect)
+                if (!MultiSelect && !CheckBoxes)
```
Smurf-IV: This logic is not correct.
Why check for `CheckBoxes` it has nothing to do with MultiSelect.

Smurf-IV (changes requested):

Source/Krypton Components/Krypton.Toolkit/Controls Toolkit/KryptonTreeView.cs
PWagner1: Fixed in latest commit

Smurf-IV (approved):

